PR #1, opened 2015-08-24
Use existing connection in X.H.DynamicBars

0:
This is just the patch from https://code.google.com/p/xmonad/issues/detail?id=538, updated so it applies to the latest version of the file.

This fixes that issue for me using both the minimal `xmonad.hs` from the bug report and using my real configuration.


vrs:
I was going to open a PR for just this patch. I use it as well, no problems.

PR #4, opened 2015-09-19
X.H.DynamicBars: support per-monitor cleanup

mathstuf:
Add functions to allow cleaning up only screens which disappear. This
works better where killing the statusbar for a specific screen is
possible. The old way is still relevant for setups which do not have
such a method (e.g., safeSpawn xmobar/spawnPipe xmonadpropwrite).

### DIFF
--- a/XMonad/Hooks/DynamicBars.hs
+++ b/XMonad/Hooks/DynamicBars.hs
@@ -18,8 +18,11 @@ module XMonad.Hooks.DynamicBars (
   -- $usage
     DynamicStatusBar
   , DynamicStatusBarCleanup
+  , DynamicStatusBarPartialCleanup
   , dynStatusBarStartup
+  , dynStatusBarStartup'
   , dynStatusBarEventHook
+  , dynStatusBarEventHook'
   , multiPP
   , multiPPFormat
   ) where
@@ -30,6 +33,7 @@ import Control.Monad
 import Control.Monad.Trans (lift)
 import Control.Monad.Writer (WriterT, execWriterT, tell)
 
+import Data.List
 import Data.Maybe
 import Data.Monoid
 import Data.Foldable (traverse_)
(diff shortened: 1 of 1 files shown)


XMonad/Hooks/DynamicBars.hs
```diff
@@ -64,6 +68,9 @@ import qualified XMonad.Util.ExtensibleState as XS
 -- is called when the number of screens changes and on startup.
 --
 
+-- TODO: Make into [(ScreenId, Handle)] instead of keeping the two parallel
+-- lists.
```
byorgey: Are these `TODO` and following `XXX` comment things that should be dealt with now?  Or can they be put off until later?

mathstuf: Yeah, I should probably fix this.

mathstuf:
Updated.

byorgey:
,,00

byorgey:
Hmm, it doesn't seem to build for me now, I get errors about `dsbInfoScreens` not being in scope.

Also, sorry for the stray comment earlier, not sure what happened there.

mathstuf:
Gah, sorry. Ended up copy/paste from the wrong parts once I replaced:

```haskell
  info <- XS.get
  let (dsbInfoScreens, dsbInfoHandles) = unzip (dsbInfo info)
```

with:

```haskell
  (dsbInfoScreens, dsbInfoHandles) <- XS.get >>= return . unzip . dsbInfo
```

byorgey:
Sorry for letting this go so long.  I was looking back through some old to-do items and noticed this.  (The spring semester happened in between!)  Anyway, it seems to build now.  The only issue I can see is that it really needs documentation, in the form of Haddock comments --- there is no way for a user to know the difference between the primed and non-primed variants of the functions.

mathstuf:
Rebased for conflicts; I'll get to updating the haddock later this week.

mathstuf:
Documentation improved.

byorgey:
Thanks!

PR #5, opened 2015-09-23
Add layout hook for ScreenCorners.

kurnevsky:
It is needed because other windows can cover ours corner windows.


kurnevsky:
Maybe it also can be done with StackMode = TopIf. I will check it.

kurnevsky:
I checked - with StackMode we also need to specify sibling window, so i don't think that it is usable.

kurnevsky:
Maybe, would be better to use logHook?

byorgey:
Thanks!

PR #6, opened 2015-09-23
X.A.Submap: establish pointer grab to avoid freezing X.

sgf-dma:
Establish active asynchronous pointer grab before entering infinity cycle.
Because xmonad already has passive synchronous pointer grab, this overwrites
it temporary and avoids freezing X, when button press occurs after submap key
press.

Also, terminate submap at button press the same way, as we do for wrong key
press.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,12 @@
     EWMH taskbars and pagers. Useful for `NamedScratchpad` windows, since
     you will usually be taken to the `NSP` workspace by them.
 
+  * `XMonad.Actions.Submap`
+
+    Establish pointer grab to avoid freezing X, when button press occurs after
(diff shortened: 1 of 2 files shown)


pjones:
Please include an entry in `CHANGES.md`.  Thank you.

sgf-dma:
Updated.

pjones:
Thanks!

sgf-dma:
I've noticed only now, that note in CHANGES.md i've added in the wrong section. Sorry, how should i fix?

byorgey:
@sgf-dma At this point, probably best to just submit a new pull request with the fix.  Feel free to cc me.  If it's just a documentation fix it should be merged quickly.

PR #11, opened 2015-11-18
New module: XMonad.Actions.DynamicProjects

pjones:
Inspired by `TopicSpace`, `DynamicWorkspaces`, and `WorkspaceDir`,
`DynamicProjects` treats workspaces as projects while maintaining
compatibility with all existing workspace-related functionality in
XMonad.


byorgey:
Looks great, thanks!

PR #13, opened 2015-11-22
Adjust debugWindow to the addition of wa_colormap.

asjo:
After adding wa_colormap to getWindowAttributes in Graphics.X11.Xlib.Extras
in https://github.com/asjo/X11/commit/29c8035648d5b17c09730193ae6aa002c3fc8f5c
this change is needed in xmonad-contrib.

### DIFF
--- a/XMonad/Util/DebugWindow.hs
+++ b/XMonad/Util/DebugWindow.hs
@@ -41,7 +41,7 @@ debugWindow w =  do
   case w' of
     Nothing                                   ->
       return $ "(deleted window " ++ wx ++ ")"
-    Just (WindowAttributes x y wid ht bw cm m o) -> do
+    Just (WindowAttributes x y wid ht bw cm mi m o) -> do
       c' <- withDisplay $ \d ->
             io (getWindowProperty8 d wM_CLASS w)
       let c = case c' of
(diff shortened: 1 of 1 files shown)


dmwit:
Let's switch over to a record pattern so we don't have to do this kind of update every time we support more `WindowAttribute`s. That should mean we don't have to bump the X11 dependency as well, which is a nice bonus.

asjo:
Sounds like a great idea - I'll delete this pull request in favour of that solution. Thanks!

ptrxyz:
Since X11 already accepted the patch, wouldn't it be good to accept the patch in xmonad-contrib, too? For consistency reasons and to make it compile flawlessly.

asjo:
@ptrxyz I think @dmwit is suggesting a better solution than having to, for every field added to WindowAttributes, add to the code in xmonad-contrib, in lockstep. I'm still a the beginning of the learning curve, so "switch over to a record pattern" is out of my reach currently, that's why I closed this pull request.

dmwit:
I didn't realize you didn't plan on implementing that. Okay.

I've put together a patch for this, but my understanding is that xmonad and xmonad-contrib are currently in code freeze pending a release. I will push once we're unfrozen.

asjo:
I was planning on trying at some point, but didn't have time, sorry.

Note that if you don't catch exceptions from getWindowAttributes in XMonad, then you should remove the throwIfZero from X11, otherwise the release will be very crashy.

ptrxyz:
Are xmoand/xmonad-contrib unfrozen already? :)

geekosaur:
They weren't really frozen. Just I've been low-spoons of late and I'm not sure anyone else is active.

The debug* modules need a few fixes, some of which are waiting for me to clean up my local copies and submit.

PR #17, opened 2015-12-08
UpdatePointer bugfix.

kurnevsky:
Sometimes cursor was stuck in one of the corners.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -67,6 +67,10 @@
 
     - Fix `raiseNextMaybe` cycling between 2 workspaces only.
 
+  * `XMonad.Actions.UpdatePointer`
+
+    - Fix bug when cursor gets stuck in one of the corners.
+
 
 ## 0.12 (December 14, 2015)
 
--- a/XMonad/Actions/UpdatePointer.hs
+++ b/XMonad/Actions/UpdatePointer.hs
@@ -1,7 +1,7 @@
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonadContrib.UpdatePointer
--- Copyright   :  (c) Robert Marlow <robreim@bobturf.org>
+-- Copyright   :  (c) Robert Marlow <robreim@bobturf.org>, 2015 Evgeny Kurnevsky
 -- License     :  BSD3-style (see LICENSE)
 --
 -- Maintainer  :  Robert Marlow <robreim@bobturf.org>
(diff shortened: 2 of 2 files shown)


byorgey:
This looks good to me.  My only comment is that perhaps `getWindowAttributesMaybe` should go in `Graphics.X11.Xlib.Extras` in the `X11` package instead of here?  It seems like it might be more generally useful.

asjo:
I have tried to fix the getWindowAttributes failing problem in a different way: https://github.com/asjo/X11/commit/e5421307a02ea0cadcb381d282615d56a0e848c0 in https://github.com/xmonad/X11/pull/35 - (to solve the transparent border problem: https://github.com/xmonad/xmonad/pull/9) I don't know what method is preferred.

kurnevsky:
@byorgey Perhaps you are right. I just thought that it assumed to use xGetWindowAttributes in such cases.
@asjo I think it would be good if getWindowAttributes will throw exception on zero status. It is better than return invalid result. But it might be useful to have another exception-less function like tryGetWindowAttributes.

byorgey:
Yes, I agree it makes sense for `getWindowAttributes` to throw an exception rather than return an invalid result.  So let's also add `getWindowAttributesMaybe` (or `tryGetWindowAttributes`, I don't care too much what it's called) to the `X11` package.

kurnevsky:
@byorgey ok, I rewrote this bugfix with Control.Exception.try because @dmwit didn't merge PR with tryGetWindowAttributes.

kurnevsky:
Oops, sorry, it doesn't compile now.

kurnevsky:
Fixed.

kurnevsky:
(I didn't test it because xmonad-contrib doesn't compile with X11 head)

bb010g:
Anything holding this back?

geekosaur:
This looks okay to me. Has it actually been tested by now (noting the earliier "did not test because...")? The "brown paper bag" failure with struts was annoying enough to have let through.

kurnevsky:
@bb010g @geekosaur This pullrequest will fix the bug only with version 1.7 of X11. The latest stable version doesn't throw an exception https://github.com/xmonad/X11/blob/1.6.1.2/Graphics/X11/Xlib/Extras.hsc#L976
But originally I fixed it with xGetWindowAttributes (in the first commit, this works with latest stable X11) and changed it to try $ getWindowAttributes after discussion above.

pjones:
Please include an entry in `CHANGES.md` and after X11 1.7 is released we will merge this PR.  Thanks.

kurnevsky:
Done.

pjones:
Thanks!

liskin:
So, um, will this fix work with X11 <1.7? Cause xmonad-contrib.cabal allows that.

kurnevsky:
It won't. But it doesn't break building. However there was breking changes in X11 1.7 so I'm not sure if current xmonad works with older versions.

PR #19, opened 2015-12-09
Bugfix for moveHistory when alwaysHighlight is enabled.

kurnevsky:
alwaysHighlight option doesn't work with moveHistory currently. We need to reset complIndex and select first element inside moveHistory.


byorgey:
Thanks!

PR #21, opened 2015-12-11
fix #16

### LINKED ISSUE #16: gridSelectWorkspace loses focus
If I have gridSelectWorkspaces from XMonad.Actions.GridSelect active and I click a window on another screen, the gridselect loses focus and it can't be restored. If the gridselect has lost focus, you can't click it, can't type anything and can't escape out of it. The only way I've found to fix it is by restarting xmonad.

f1u77y:
i can't check if it works because i haven't mutiple screens. anyway, grabButton was useless and grabPointer seems to be not.


kurnevsky:
I checked it - it works.

f1u77y:
could this be merged?

byorgey:
Thanks!  Sorry for the delay!

PR #23, opened 2015-12-13
Add multiple key support for completion key

psibi:
This patch enables support for key binding like Ctrl + i which was not
previously possible. Technically, this changes the type of completionKey
from KeySym to (KeyMask, KeySym). Note that programs like dmenu support
Ctrl + i by default.


psibi:
CC: @byorgey @pjones 

Can this be merged ?

pjones:
Please add an entry to the `CHANGES.md` file under the 0.13 heading (you may have to create it, see the existing headings for an example).

Be sure to note that your change might break existing configurations. 

Also add your name to `.mailmap` if it isn't already there. 

Finally, push those changes so they show up in this PR and I'll merge it.

psibi:
@pjones Done.

pjones:
Thanks!

PR #29, opened 2015-12-31
Prompt position

f1u77y:
Add new position for X.Prompt: `CenteredAt y w`(centered prompt of width `w` and top border at `y`). Also fix border width between main window and completions window: now it is `promptBorderWidth conf`, not `2 * promptBorderWidth conf`


pjones:
I love this idea.

Please add some documentation about what the two `Rational`s are in `CenteredAt`.  Is the height a percentage of screen height?

XMonad/Prompt.hs
```diff
@@ -228,6 +228,15 @@ class XPrompt t where
 
 data XPPosition = Top
                 | Bottom
+                -- | Prompt will be placed in the center horizontally and
+                --   in the certain place of screen vertically. If it's in the upper
+                --   part of the screen, completion window will be placed below(like
+                --   in 'Top') and otherwise above(like in 'Bottom')
+                | CenteredAt { xpCenterY :: Rational
+                             -- ^ Relative Y coordinate of center of the prompt.
```
byorgey: What does "relative Y coordinate" mean?  Is it an offset from the center, measured in pixels?  Is it a value between 0 and 1? etc?

XMonad/Prompt.hs
```diff
@@ -228,6 +228,15 @@ class XPrompt t where
...
+                             , xpWidth  :: Rational
+                             -- ^ Relative width of the prompt
```
byorgey: Same question --- what does "relative width" mean?

XMonad/Prompt.hs
f1u77y: "relative" means relative to screen width for prompt width and relative to screen hight for y coordinate. i did't think that this could be understood in other way

byorgey: My concern was more that the units are not specified.  I suppose saying
"relative" plus the fact that it is a Rational value should be a hint that
it should be a value from 0 to 1 (as opposed to, say, a number of pixels).
But it would still be nice to say this explicitly.  "A rational between 0
and 1, giving the relative..."

f1u77y: ok. i think "relative to screen width/height" is enough, isn't it?

f1u77y: ok, i understood. fixed

f1u77y:
@pjones i've added documentation, can it be merged?

pjones:
Please add an entry to the `CHANGES.md` file under the 0.13 heading (you may have to create it, see the existing headings for an example).

Also add your name to `.mailmap` if it isn't already there.

Finally, push those changes so they show up in this PR and I'll merge it.

f1u77y:
@pjones done

pjones:
Thanks!

PR #31, opened 2016-01-16
add X.U.Ungrab

geekosaur:
This module provides a better workaround for xmonad holding the keyboard grab through actions, causing spawned commands that also want to do a grab to fail.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,14 @@
   * New constructor `CenteredAt Rational Rational` added for
     `XMonad.Prompt.XPPosition`.
 
+### New Modules
+
+  * `XMonad.Util.Ungrab`
+
+    Release xmonad's keyboard and pointer grabs immediately, so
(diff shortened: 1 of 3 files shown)


pjones:
Don't forget to add an entry to `CHANGES.md`.

geekosaur:
have some energy for a change. also almost over the "whoops" of doing this on master...
(naturally I forgot to tag the changelog update skip-ci...)

geekosaur:
so nobody else is reviewing these? Seems wrong to be approving my own patches.
(And I need to do something with this, one way or the other, to unhose my fork.)

byorgey:
I try to review xmonad patches when I have the time; but these days I don't have much time, and in any case I don't really understand all the X-specific stuff.  I wouldn't say approving one's own patches is *wrong*, I would just say it is *not ideal*.  If no one else has the time or expertise to review a patch then there's not much else to do.

PR #33, opened 2016-01-24
Add stack.yaml

sid-kap:
If this PR gets merged, I'd eventually like to add xmonad-contrib to Stackage.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,0 +1,6 @@
+flags: {}
+packages:
+- '.'
+extra-deps:
+  - 'X11-xft-0.3.1'
+resolver: lts-4.1


pjones:
What's the advantage of having this file added to the repository? 

From what I understand, many Haskell project maintainers are keeping these files out of their repos since `stack init` can generate them easily.

Also, why is this file necessary for getting contrib  on stackage?

sid-kap:
I was under the impression that the repository needed a `stack.yaml` file in order to add it to Stackage. After looking it up, it doesn't seem like that's the case.

One advantage of adding this file is to make it easier for people to build your project. If there project needs additional dependencies, `stack init` might not pick up those dependencies, which means the user has to manually add them. For example, I had to manually add X11-xft to the stack.yaml file in order to build it, since X11-xft is not on Stackage.

I'm not sure about what other project maintainers are doing. For instance, Gabriel Gonzalez's projects, and Pandoc, keep it in their repos. On the other hand, Bryan O'Sullivan and Edward Kmett don't.

I'm not too knowledgeable about this, so you probably know better than I do whether it would be a good idea for this project.

mgsloan:
Se [this thread](https://www.reddit.com/r/haskell/comments/415vne/to_ship_stackyaml_or_not_to/) for "to ship stack.yaml or not".

byorgey:
It is nice to let the user know that if they want to build xmonad-contrib with stack then they will need to manually add X11-xft.  However, personally I don't think I want this file in the repo because then we have to remember to update it in the future, when the best LTS version to use or the version of X11-xft to use changes.

mgsloan:
As long as you have something like Travis CI, you should be at least be notified when the x11-xft version must be updated. I also think it's fine to have one of these with an older snapshot of that's what the author is using.

That said, looking at this closer, there isn't a ton of value in this case, because xmonad-contrib is only really useful when built along with the xmonad package.  If xmonad was a submodule, or there was a repo with both these repos as submodules, then a stack.yaml would be quite useful, for specifying how to build the xmonad executable along with xmonad-contrib

sid-kap:
Okay, I'm closing this issue. Thanks for all of your comments.

PR #34, opened 2016-01-26
Added missing boundary check in Layout.Spacing.shrinkRect

bb-h8:


XMonad/Layout/Minimize.hs
```diff
@@ -112,7 +112,7 @@ instance LayoutModifier Minimize Window where
                 case M.lookup w (W.floating ws) of
                   Nothing -> return $ Just $ Minimize (w:minimized) unfloated
                   Just r -> do
-                    modify (\s -> s { windowset = W.sink w ws})
+                    windows (W.sink w)
```
byorgey: What is the purpose of this change?  If I remember correctly, `windows` does not just modify the stackset, but also redraws all the windows and does some other stuff.  Is that what we want here?

geekosaur: In particular, it's going to re-run the layout. This seems like a bad thing to do from the layout.

bb-h8: Oh, wtf is going on? This has obviously nothing to do with the Spacing thing. Why is this even in this pull request?
I found a minor bug with Minimize that this commit fixes, but I realized that Minimize prior used `windows` instead of `modify` and changed it because it caused problems with decorations for exactly the mentioned reasons. I didn't have the time to make this properly, so I just left it there.
How do I fix this? Help!

bb-h8: Ok, I think I got it.

byorgey:
Thanks!

PR #41, opened 2016-02-19
Fixed bottom-yielding definition of toggleGroupFull

jmickelin:
The current value of `toggleGroupFull` is ⊥, since it's defined as `toggleGroupFull = toggleGroupFull`.

Presumably, this is a typo for `toggleColumnFull` from `XMonad.Layout.Groups.Examples`.


byorgey:
Thanks!

PR #42, opened 2016-02-23
handle messages in X.L.IfMax

f1u77y:


byorgey:
Thanks! However, this does not build on GHC < 7.10 because <$> is not in scope (see the Travis build). The easiest fix would be to just use fmap.

f1u77y:
sorry, forgot about that. fixed

PR #45, opened 2016-02-29
watch only for tiled windows in X.L.IfMax

f1u77y:
watching for floating windows can cause problems when using with X.L.MouseResizableTile and watching only for tiled is more intuitive.


geekosaur:
Merged.

Issue #46, opened 2016-02-29
multiple windows are highlighted as active when using X.L.Minimize


f1u77y:
Steps to reproduce:
- Add `minimize` and `boringWindows` to layout
- Open 2 or more windows on workspace
- Minimize one of them
- Restore it with `sendMessage RestoreNextMinimizedWin`
  Two windows(one is that what active before last step and other is restored window) will be highlighted as active

pjones:
It seems this will be fixed with #102 so I'm going to close it.

PR #49, opened 2016-04-06
Redraw border when minimizing and restoring window

f1u77y:
Border color isn't set corretly with `modify`.

### DIFF
--- a/XMonad/Layout/Minimize.hs
+++ b/XMonad/Layout/Minimize.hs
@@ -91,6 +91,16 @@ setMinimizedState win st f = do
             fi_mini = fromIntegral mini
         io $ changeProperty32 dpy win wm_state ptype propModeReplace (f fi_mini wstate)
 
+-- We need that to redraw border without running hooks in `windows`
+modifyAndRedrawBorder :: (WindowSet -> WindowSet) -> X ()
+modifyAndRedrawBorder f = do
+    XState { windowset = old } <- get
+    let ws = f old
+    XConf { display = d , normalBorder = nbc, focusedBorder = fbc } <- ask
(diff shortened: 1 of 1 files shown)


geekosaur:
I'd imagine the original was intended to be used with `sendMessage`, which causes a refresh afterward; that *should* force the window border to be correct. since `windows` would then recolor all visible window borders. (And conversely, using `windows` here means a double update which could be slow with many windows.)

That said, this whole area is fairly complex. It is worth testing this fairly well, and possibly reviewing the sources of the messages to verify that they aren't doing things like sending without refresh (or if they are, finding out *why* they are because it may mean potential lurking issues).

f1u77y:
> `windows` would then recolor all visible window border
As I can understand, `windows f` recolors only windows whose focus have been changed by `f`. So, modifying `windowset` by `modify` and not by `windows` leads to such a bug(#46).

>  (And conversely, using windows here means a double update which could be slow with many windows.)
I know that's bad, but restoring window *needs* to call `windows`. Anyway, restoring windows, I think, should not be called only by a message, because any modifying `windowset` from the internal layout functions AFAIK is bad. So, any solution that uses layout hooks for minimizing should run `windows` twice.

geekosaur:
_"any modifying windowset from the internal layout functions AFAIK is bad"_

That's kinda the point; this *is* effectively an internal layout function, so I'm wary.

f1u77y:
Ok, then now I see 2 ways to fix the bug:

1. Make `restoreNextMinimizedWindow` a function and call `windows` and other `X` stuff from it after `sendMessage`. That is more safe but still runs `windows` twice.
2. Rewrite `X.L.Minimize` without any layout hooks.

f1u77y:
Closing that PR giving preference to #102

PR #52, opened 2016-04-22
XMonad.Util.Loggers for NamedScratchPad-s

geekosaur:
Adds machinery and several Loggers for NamedScratchPads. Also adds a utility function to hide windows from EWMH taskbars and pagers, since they won't do what you expect with windows in the NSP workspace.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,17 @@
     screen grabbers and lock utilities, etc. will work. Replaces
     the short sleep hackaround.
 
+  * `XMonad.Util.Loggers.NamedScratchpad`
+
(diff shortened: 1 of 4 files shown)


byorgey:
It appears that `deriving Typeable` requires a LANGUAGE pragma on versions of GHC before 7.10 --- can you add that?

geekosaur:
Sorry, made that PR and promptly forgot about it, including checking CI.

byorgey:
Thanks! This looks useful.

geekosaur:
Might have been a bit premature, I used it with an older program and it doesn't get recorded properly except during startup. I suspect the program predates (more precisely, is modified from a program that predates) ConfigureRequest and I will also have to monitor MapRequest in the handleEventHook. :/

PR #53, opened 2016-04-27
Make WindowBringer case insensitive

jablko:
I love WindowBringer but I sometimes forget the menu items (window titles) are all lowercase. I type a capital letter, e.g. "Makefile" and get no results.
I tested this change and it works, whether the user enters all lowercase (as before) or mixed case.


byorgey:
Looks reasonable to me.  Thanks!

jablko:
Thank you!

PR #57, opened 2016-06-07
Fix minor things

CaptainPatate:
* Fix a minor error in the documentation of the function `seconds`
* Fix type declaration for two EZConfig functions used to configure keybindings. I think they should take a `KeyMask` instead of a `ButtonMask` (both are synonym for `Modifier` which is a `CUInt`).


geekosaur:
Applied, thanks!

PR #59, opened 2016-06-28
New SortedLayout module.

chongli:
A new LayoutModifier that sorts a given layout by a list of properties. The order of properties in the list determines the order of windows in the final layout. Any unmatched windows go to the end of the order.

### DIFF
--- a/XMonad/Layout/SortedLayout.hs
+++ b/XMonad/Layout/SortedLayout.hs
@@ -1,0 +1,88 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeSynonymInstances  #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Layout.SortedLayout
+-- Copyright   :  (c) 2016 Kurt Dietrich
+-- License     :  BSD-style (see xmonad/LICENSE)
+--
+-- Maintainer  :  kurto@mac.com
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- A 'LayoutModifier' that sorts the windows in another layout, given a
+-- list of properties. The order of properties in the list determines
+-- the order of windows in the final layout. Any unmatched windows
+-- go to the end of the order.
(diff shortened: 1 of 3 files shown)


XMonad/Layout/SortedLayout.hs
```diff
@@ -0,0 +1,88 @@
...
+  matching = filterM (hasProperty prop) wids
+
+sortLayout :: (LayoutClass l Window)
+           => [Property]
+           -> W.Workspace WorkspaceId (l Window) Window
+           -> Rectangle
+           -> X ([(Window, Rectangle)], Maybe (l Window))
+sortLayout props (W.Workspace w l r) rect = do
+  let wids = W.integrate' r
+  sortedWids <- map wdId . sort . nub . concat <$> zipWithM (\s p -> findMatchingWindows s p wids) [0..] props
```
aavogt: `sort . nub` looks off at first glance, since `map head . group . sort` or even `nub . sort` do less work. If this module was larger, I'd say "don't write instance Eq/Ord WindowDescriptor` and use `sortOn wdId . nubOn wdSeqn` (or equivalent with nubBy/sortBy) instead. Maybe I'm not alone in assuming `a == b`  gives the same result as `compare a b == EQ`.

chongli: Good catch with `sort . nub`. I changed it to `nub . sort`.

I could add `sortOn wdSeqn` but `nubOn` is part of the `extra` package. Seems a bit excessive to pull in another dependency just for that, no?

I could also do `nubBy (\a b -> wdId a == wdId b)` but that is a real mouthful, isn't it?

aavogt:
Looks good.

If the layout has n windows and m properties to reorder, I think your layout will make n*m queries to X. I think you could do it with less, but I dunno if that's worth the effort.

geekosaur:
On Fri, Jul 1, 2016 at 11:52 AM, Adam Vogt <notifications@github.com> wrote:

> If the layout has n windows and m properties to reorder, I think your
> layout will make n*m queries to X. I think you could do it with less, but I
> dunno if that's worth the effort.


Ideally we'd cache property lookups; this would also speed up complex
ManageHooks considerably.

-- 

PR #65, opened 2016-07-15
Add `XMonad.Actions.KeyChord`

pjrt:
`KeyChord` is a module to building keychords using `submap` from`XMonad.Actions.Submap`. However, this module exports a simplerinterface for creating long keychords. Here, a "keychord" is a series of keystrokes, similar to what you would get in vi/vim.

Addd a new dependency: `PathTree` https://hackage.haskell.org/package/PathTree

`PathTree` is also created and maintained by me.

### DIFF
--- a/XMonad/Actions/KeyChord.hs
+++ b/XMonad/Actions/KeyChord.hs
@@ -1,0 +1,85 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Actions.KeyChord
+-- Copyright   :  (c) Pedro Rodriguez Tavarez <pedro@pjrt.co>
+-- License     :  BSD3-style (see LICENSE)
+--
+-- Maintainer  :  Pedro Rodriguez Tavarez <pedro@pjrt.co>
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- A module that allows the user to create keychords similar to vi using
+-- 'submap' from "XMonad.Actions.Submap".
+--
+-- What 'keychords' does can be manually (and painfully) replicated using plain
+-- 'submap'. This module aims to make the process simpler.
+--
+-------------------------------------------------------------------------------
+module XMonad.Actions.KeyChord
+( -- Usage
+  -- $usage
+  keychords
+) where
+
+import Data.PathTree (fromPaths, LCRSTree(Empty, Leaf, Node))
+import XMonad
+import XMonad.Actions.Submap (submap)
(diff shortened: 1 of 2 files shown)


pjrt:
Unn, now that I think about it, should this module be called "keychains" instead? Chord can mean multiple keys hit together. 

And more importantly, given that it is just a simplifying tool for Submap, should the function be placed there?

geekosaur:
On Sat, Jul 16, 2016 at 4:23 PM, Pedro Rodriguez <notifications@github.com>
wrote:

> Unn, now that I think about it, should this module be called "keychains"
> instead? Chord can mean multiple keys hit together.
>
> And more importantly, given that it is just a simplifying tool for Submap,
> should the function be placed there?
>
I'm not fond of either name, since "keychain" conventionally means a
collection of encryption key rings --- and we already have at least one
password manager interface under XMonad.Util, so the confusion *is*
possible. KeySequence (or KeySeq), maybe?

-- 

pjrt:
KeySequence sounds great, I'll go with that one.

aavogt:
The prompt module has an autoComplete option which lets you get almost the same result if you are restricted to printable characters, with the difference being:

A list of letters to type shows up which helps you see duplicate/overlapping key sequence. Helps but it could also be a source of delay or distraction.

Configuration here uses xK_... while there it's almost like EZConfig.

But that similarity doesn't hold if you use modifier keys in the submaps. So emacs like keys might be a compelling reason to include this.

byorgey:
Note that EZConfig already has a facility for building chains of keystrokes
with nested submaps. I'm travelling and can't look at this PR in detail
right now but I wonder what the relationship is.

pjrt:
@byorgey ha! Well that's embarrassing, you are correct. Not sure how I didn't see that when I was searching for a way to do this. 

Apologies to everyone for wasting their time. Closing.

@aavogt 
> But that similarity doesn't hold if you use modifier keys in the submaps. So emacs like keys might be a compelling reason to include this.

Not sure I follow, I tried adding a modifier in the submap (ie, "M-; i S-a") and it worked. Unless you mean something else by "use modifier keys in the submap".

byorgey:
@pjrt does EZConfig fulfill your use case?  Or would it be useful pulling out the key sequence/submap code into a separate module?

pjrt:
@byorgey It does fulfill my use case, I've already changed my configuration to use it and it works like a charm. Thanks!

PR #72, opened 2016-08-08
Add TreeSelect action for selecting from many workspaces and X actions

TomSmeets:
You can select the desired workspace/action with the cursor or hjkl keys

This module is fully configurable and very useful if you have a lot of workspaces.
I tryed to document the entire module. For more info look at that module (XMonad.Actions.TreeSelect)

Selecting Workspaces
Selecting Actions

### DIFF
--- a/XMonad/Actions/TreeSelect.hs
+++ b/XMonad/Actions/TreeSelect.hs
@@ -49,6 +49,9 @@ module XMonad.Actions.TreeSelect
     , moveChild
     , moveNext
     , movePrev
+    , moveHistBack
+    , moveHistForward
+    , moveTo
 
       -- * Advanced usage
       -- $advusage
@@ -71,6 +74,7 @@ import XMonad.StackSet as W
 import XMonad.Util.Font
 import XMonad.Util.NamedWindows
 import XMonad.Util.TreeZipper
+import XMonad.Hooks.WorkspaceHistory
 import qualified Data.Map as M
 
 #ifdef XFT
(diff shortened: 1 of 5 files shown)


byorgey:
This looks very cool, thanks!  I hope to get a chance to look at it soon.

byorgey:
So I just got a chance to try this, and unfortunately it does not work for me.  When I activate `treeselectWorkspace`, one of my screens just turns completely gray, and none of the default keybindings do anything.  I am still able to switch to the other screen and use other xmonad commands, but the only thing that gets the gray screen back to normal is to restart xmonad with mod-q.

I wish I had more time to come up with a minimal working example, but my config is huge and complicated.  I hope I will have a chance to do that later, or maybe someone else can try this and see if it works for them or not.

TomSmeets:
Thanks for the feedback!

I think i have found the problem. You might have forgotten to include this part:

> Then add it to your XMonad Config using the `toWorkspaces` function
> 
> > xmonad $ defaultConfig {
> >                                        , workspaces = toWorkspaces myWorkspaces
> >                                        }

This is needed because `treeselectWorkspace` expects your `XConfig.workspaces` to be a string representation of your workspace-tree (`myWorkspaces` in this example).

The `toWorkspaces` function does this automatically for you.
```
> toWorkspaces [ Node "Hello"  [ Node "World" [], Node "Planet" []]]
["Hello","Hello.World","Hello.Planet"]
```

byorgey:
Ah, thanks, that might have been it.  I'll try to take a look again soon!

byorgey:
I got around to trying this again and it works great!

One final comment: the documentation mentions setting `logHook = workspaceHistory`, but this does not type check: `logHook` expects an `X ()` but `workspaceHistory :: X [WorkspaceId]`.  I didn't think it was worth holding up merging this, but the documentation should be fixed.

No worries about the extra commit, since it's already been merged into master it doesn't make any difference.

TomSmeets:
Thanks for merging!

`workspaceHistory` was supposed to be `workspaceHistoryHook` from `XMonad.Hooks.WorkspaceHistory`.
I created a new pull request (https://github.com/xmonad/xmonad-contrib/pull/84)

PR #77, opened 2016-08-26
A very simple extension to Navigation2D so it plays better with gaps

LSLeary:
The documentation in the diffs pretty much covers it, but to summarise, I simply combined the two navigation methods that were already implemented into a 'best of both worlds' version. Line navigation fails to produce the expected motions when the line it's using falls on gaps, so in these cases you want to fall back on the more robust Center navigation, and that's exactly what Hybrid navigation does.

This is my first contribution to xmonad and also my first pull request, so please do provide constructive criticism if I've done something stupid or failed to follow some style guideline.

Edit: An example of a layout where gaps break Line navigation, in case it wasn't clear what the problem was. The layout below is achieved with SplitGrid tiling mode.
2016-08-27-041550_1920x1080_scrot

Here, if focus is on one of the six windows in the bottom right, then line navigation can take you left, but you won't be able to go right afterwards. Similarly, you can go up, but not back down. If Center navigation is used, then you can move down and left as needed, but you cannot go up from all four windows that are directly below the top window, you can only do so from the two most central windows.

With Hybrid navigation all these motions are possible, with priority to the more intuitive motions derived from Line navigation.

My xmonad config file is available here on github if it's of use or relevance.

Further edit: It may be worth noting that the Navigation2D type has a 'generality' parameter, which in this case I chose somewhat arbitrarily to be 2, the same as centerNavigation. On further thought, maybe it should be 3, since it will sometimes move when Center navigation won't. That said, I can't see anywhere the parameter is actually used, except for in making Navigation2D an instance of the Eq and Ord typeclasses.

Further further edit:
The only piece of meaningful code introduced is the function below.

```haskell
doHybridNavigation dir (cur, rect) winrects
  | (doLineNavigation dir (cur, rect) winrects == Nothing) = doCenterNavigation dir (cur, rect) winrects
  | otherwise = doLineNavigation dir (cur, rect) winrects
```

At the time it was just written to work, and the thought of making it a bit cleaner has only occurred to me just now. Rewriting it, I find the following vastly more aesthetically pleasing. It's untested since I'm currently on the wrong machine, but I'll test it and push the commit ~~within a day or two~~. **Done.**

```haskell
doHybridNavigation = applyToBoth pickSomething doLineNavigation doCenterNavigation
    where
        applyToBoth f g h a b c = f (g a b c) (h a b c)
        pickSomething Nothing b = b
        pickSomething a _ = a
```

### DIFF
--- a/XMonad/Actions/Navigation2D.hs
+++ b/XMonad/Actions/Navigation2D.hs
@@ -43,6 +43,7 @@ module XMonad.Actions.Navigation2D ( -- * Usage
                                    , Navigation2D
                                    , lineNavigation
                                    , centerNavigation
+                                   , hybridNavigation
                                    , fullScreenRect
                                    , singleWindowRect
                                    , switchLayer
@@ -74,10 +75,11 @@ import XMonad.Util.Types
 -- natural but may make it impossible to navigate to a given window from the
 -- current window, particularly in the floating layer.  /Center navigation/
 -- feels less natural in certain situations but ensures that all windows can be
--- reached without the need to involve the mouse.  Navigation2D allows different
--- navigation strategies to be used in the two layers and allows customization
--- of the navigation strategy for the tiled layer based on the layout currently
--- in effect.
+-- reached without the need to involve the mouse. A third option is to use
+-- /Hybrid navigation/, which automatically chooses between the two whenever
+-- navigation is attempted.  Navigation2D allows different navigation strategies
+-- to be used in the two layers and allows customization of the navigation strategy
+-- for the tiled layer based on the layout currently in effect.
 --
 -- You can use this module with (a subset of) the following in your @~\/.xmonad\/xmonad.hs@:
 --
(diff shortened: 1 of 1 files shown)


XMonad/Actions/Navigation2D.hs
```diff
@@ -771,9 +771,11 @@ doCenterNavigation dir (cur, rect) winrects
...
 -- then falls back on Center navigation if it finds no suitable target window.
 doHybridNavigation :: Eq a => Direction2D -> Rect a -> [Rect a] -> Maybe a
-doHybridNavigation dir (cur, rect) winrects
-  | (doLineNavigation dir (cur, rect) winrects == Nothing) = doCenterNavigation dir (cur, rect) winrects
-  | otherwise = doLineNavigation dir (cur, rect) winrects
+doHybridNavigation = applyToBoth pickSomething doLineNavigation doCenterNavigation
+    where
+        applyToBoth f g h a b c = f (g a b c) (h a b c)
+        pickSomething Nothing b = b
+        pickSomething a _ = a
```
byorgey: Note `pickSomething` already exists in the standard library, as `(<|>)`.

byorgey:
Looks nice!  If you want to change `pickSomething` to `(<|>)` I'll let you do that first and then merge it.  But it's not necessary, I'm happy to merge as-is.

LSLeary:
Done!

byorgey:
Thanks!

PR #79, opened 2016-08-30
set _NET_WORKAREA for each workspace

### LINKED ISSUE #73: X.H.ManageDocks.AvoidStruts breaks dropdown menus on multi-monitor screens
Starting a few months ago, dropdown menus in some applications sometimes stop working when the window is on one screen, then seem to work again with the window on the other screen - only for them to stop working again after changing to another workspace.

The reason for the bug is that [AvoidStruts](https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Hooks/ManageDocks.hs#L272) calculates _NET_WORKAREA for the root window using the position, dimensions and struts of only one screen, and updates it whenever a window is mapped on that screen. Hence, on multi-monitor setups, when swapping windows from one screen to another, the calculated workarea often ends up not being the correct one for the currently focused window.

This affects several applications, notably [chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=510079#c59), most Qt apps (due to [similar use](http://code.qt.io/cgit/qt/qtbase.git/tree/src/plugins/platforms/xcb/qxcbscreen.cpp#n132) of the workarea rectangle to make dropdowns avoid struts), and it probably also affects all applications that rely on _NET_WORKAREA for any purpose.

This issue was noted before in https://code.google.com/archive/p/xmonad/issues/158 (comment 12), and https://code.google.com/archive/p/xmonad/issues/526. It seems generally agreed that _NET_WORKAREA is unsuited for multi-monitor setups with docks in-between screens, but in this case, applications will break regardless of where the docks are.

I'm not sure what the best solution is for this. One way would be to make xmonad calculate and store the workarea for each screen in _NET_WORKAREA, but this also would require patches to client applications since most of them only ever look at the first 4 values of the property. The other way would be to drop support for _NET_WORKAREA altogether ; this is what [i3](https://github.com/i3/i3/blob/eb837cbd009dab61e713ca12c0d882d707c3f2a3/include/ewmh.h#L91) and [awesome](https://github.com/awesomeWM/awesome/issues/677) ended up doing long ago, basically for the same reasons.

f1u77y:
fixes xmonad/xmonad#42
fixes #73 
i'm not able to test because i don't have multiple monitors. if someone can, please test it

### DIFF
--- a/XMonad/Hooks/ManageDocks.hs
+++ b/XMonad/Hooks/ManageDocks.hs
@@ -34,10 +34,12 @@ module XMonad.Hooks.ManageDocks (
 
 -----------------------------------------------------------------------------
 import XMonad
+import qualified XMonad.StackSet as W
 import Foreign.C.Types (CLong)
 import XMonad.Layout.LayoutModifier
 import XMonad.Util.Types
 import XMonad.Util.WindowProperties (getProp32s)
+import XMonad.Util.WorkspaceCompare(getWsIndex)
 import XMonad.Util.XUtils (fi)
 import Data.Monoid (All(..), mempty)
 import Data.Functor((<$>))
@@ -271,6 +273,8 @@ instance Message SetStruts
 
 instance LayoutModifier AvoidStruts a where
     modifyLayoutWithUpdate as@(AvoidStruts ss cache smap) w r = do
+        wsIndex <- getWsIndex
+        let mwid = wsIndex (W.tag w)
         let dockWins = M.keys smap
         (nr, nsmap) <- case cache of
             Just (ss', r', nr) | ss' == ss, r' == r -> do
@@ -278,14 +282,16 @@ instance LayoutModifier AvoidStruts a where
                 if nsmap /= smap
                   then do
                     wnr <- fmap ($ r) (calcGap dockWins ss)
-                    setWorkarea wnr
+                    whenJust mwid $ \wid ->
+                      setWorkareaAt wid wnr
                     return (wnr, nsmap)
                   else do
                     return (nr, smap)
             _ -> do
                 nsset <- getRawStruts dockWins
                 nr <- fmap ($ r) (calcGap dockWins ss)
-                setWorkarea nr
+                whenJust mwid $ \wid ->
+                  setWorkareaAt wid nr
                 return (nr, nsset)
         arranged <- runLayout w nr
         let newCache = Just (ss, r, nr)
@@ -314,13 +320,25 @@ instance LayoutModifier AvoidStruts a where
             toggleOne x xs | x `S.member` xs = S.delete x xs
                            | otherwise   = x `S.insert` xs
 
-setWorkarea :: Rectangle -> X ()
-setWorkarea (Rectangle x y w h) = withDisplay $ \dpy -> do
+setWorkarea :: [Rectangle] -> X ()
+setWorkarea rects = withDisplay $ \dpy -> do
     a <- getAtom "_NET_WORKAREA"
     c <- getAtom "CARDINAL"
     r <- asks theRoot
-    io $ changeProperty32 dpy r a c propModeReplace [fi x, fi y, fi w, fi h]
+    let newWA = concatMap (\(Rectangle x y w h) -> [fi x, fi y, fi w, fi h]) rects
+    io $ changeProperty32 dpy r a c propModeReplace newWA
 
+getWorkarea :: X [Rectangle]
+getWorkarea = do
+    r <- asks theRoot
+    let toRects (x:y:w:h:other) = Rectangle (fi x) (fi y) (fi w) (fi h):toRects other
+        toRects _               = []
+    maybe [] toRects <$> getProp32s "_NET_WORKAREA" r
+
+setWorkareaAt :: Int -> Rectangle -> X ()
+setWorkareaAt i rect = setWorkarea =<< (changeElemAt i rect (Rectangle 0 0 0 0) <$> getWorkarea)
+  where changeElemAt j e d l | i >= length l = l ++ [d | _ <- [length l..i-1]] ++ [rect]
+                             | otherwise     = a ++ (e:b) where (a, _:b) = splitAt j l
 
 -- | (Direction, height\/width, initial pixel, final pixel).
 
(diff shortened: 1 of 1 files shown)


erebe:
I will give it a shot if I can tomorrow.
I keep you in touch.

joshbenner:
I use three monitors. Chrome dropdowns work on only one of them, so I believe I'm impacted by the linked bugs.This PR did not appear to fix the problem for me, but to be sure it was a reasonable test, here's what I did:

1. grabbed cabal-install
2. cloned both xmonad and xmonad-contrib
3. checked out this PR
4. did a `cabal install xmonad/ xmonad-contrib/`
5. removed the generated xmonad binary at `~/.xmonad/xmonad-x86_64-linux`
6. Changed my X session to use `~/.cabal/bin/xmonad` instead of `/usr/bin/xmonad`
7. Logged out, then logged back into my Xmonad X session
8. Opened chrome to a page with a select box
9. Cried when select box continued to work only on one of the three monitors

davama:
Attempted this myself, unfortunately i could not build xmonad-contrib with the patch.
Got this error:
```
cabal: Entering directory 'xmonad-contrib/'
Configuring xmonad-contrib-0.12...
Building xmonad-contrib-0.12...
Preprocessing library xmonad-contrib-0.12...

XMonad/Actions/GridSelect.hs:1:104: warning:
    -XOverlappingInstances is deprecated: instead use per-instance pragmas OVERLAPPING/OVERLAPPABLE/OVERLAPS
[187 of 256] Compiling XMonad.Hooks.ManageDocks ( XMonad/Hooks/ManageDocks.hs, dist/build/XMonad/Hooks/ManageDocks.o )

XMonad/Hooks/ManageDocks.hs:158:29: error:
    Not in scope: ‘W.workspaces’
    No module named ‘W’ is imported.
cabal: Leaving directory 'xmonad-contrib/'
```

geekosaur:
On Wed, Oct 26, 2016 at 11:50 AM, Dave <notifications@github.com> wrote:

> No module named ‘W’ is imported.


`import qualified XMonad.StackSet as W`


-- 

davama:
@geekosaur you're right, thanks!
Did a `vimdiff` of @f1u77y 's ManageDocks.hs file and my local file. There were way more changes than 1ed908a which is expected of course. Just copied the whole file and built contrib successfully.

Currently use two monitors.

Did similar steps to @joshbenner 
* rm ~/.xmonad/xmonad-x86-linux
* logged out/in
* Opened chrome to a page with a select box on WS 2 on monitor 2
* drop-down works 
* moved WS 2 to monitor 1
* drop-down is blank

Thanks

pjones:
From the comments it appears this PR does not resolve the issue.  Please refactor and let us know when it is ready.  Thanks.

geekosaur:
On Thu, Dec 8, 2016 at 7:44 PM, Peter J. Jones <notifications@github.com>
wrote:

> From the comments it appears this PR does not resolve the issue. Please
> refactor and let us know when it is ready. Thanks.


There was more discussion about this in the "revert avoidStruts changes"
ticket, since _NET_WORKAREA is managed by the same code.


-- 

pjones:
After digging into this issue I think we should just stop setting `_NET_WORKAREA` for now and revisit the issue after v0.13 if we really want to set this property.

As @geekosaur pointed out, other tiling window managers have stopped supporting this property:

* https://github.com/awesomeWM/awesome/commit/9c020877dd1357d7d42a29526d3567586f3db330

* https://github.com/qtile/qtile/issues/847

* https://github.com/i3/i3/commit/eec80838ab39377f18e3f68dc8a27923ae19e57e

I pushed a branch that references this PR that just sets `_NET_WORKAREA` to an empty list in order to remove it whenever the layout runs.

If you have an objection please speak up.

thomasf:
I removed the `setWorkarea` call some months ago in my own xm-contrib-fork and I can't remember if it has caused any problem at all.. I was already strongly frustrated about the blank chrome drop down issues for a long time but added to that I suddenly needed to do some QT development which forced me to identify what was going on. I'm very happy with how everything works now.

geekosaur:
Without any `_NET_WORKAREA` set, I find `calibre` only displays menus properly on my external monitor (with no docks/bars) and `google-chrome` displays empty comboboxes.

pjones:
@geekosaur Are you using my patch that actively sets `_NET_WORKAREA` to an empty list?  According i3, some display managers set `_NET_WORKAREA` as well.

With my patch, chrome and calibre work just fine on all monitors.

geekosaur:
No, I'm still running an old (with some local experiments that I need to at some point remove) version and setting `_NET_WORKAREA` explicitly in the `logHook`. In any case I am responding specifically to the idea of not setting `_NET_WORKAREA` at all, not setting it empty.

(If the later works, I'd strongly suspect it is accidental and could change at any time with Qt/KDE and/or Gtk updates.)

pjones:
I'm happy to remove it instead of setting it to an empty list I just don't see a function in X11 for it.  Do we already have a function for deleting a window property?

geekosaur:
I am confused now, you are certain that removing it (which causes Qt and Gtk malfunctions, which is why we started setting it in the first place) is somehow going to fix things?

pjones:
@geekosaur Yes.  Removing the property (or from my experiments setting it to an empty list) fixes the issues with Qt and Gtk.  That's what other tiling WMs are doing as well due to the same bugs.

Of course, desktop file managers won't avoid docks when placing their icons.  But all these whacky menu issues are fixed by ensuring `_NET_WORKAREA` doesn't exist.

pjones:
BTW: I'm setting the `_NET_WORKAREA` property to an empty list because we don't have a binding to `XDeleteProperty` in X11.  I suppose a proper fix would be to add that binding and then use it in ManageDocks.

geekosaur:
Er, is there some reason that removing the code to set it in the first
place is the incorrect answer? (Yes, this won't help with someone who just
mod-q-s into a new xmonad, but `xprop -root -remove _NET_WORKAREA` works
for what is probably not going to be a common case.)

pjones:
Yes.  We should actively delete the property for a couple of reasons.  First, the case you mention, an inplace upgrade of XMonad (or another WM).  Second, some display managers set `_NET_WORKAREA`.

pjones:
This is the solution I'm proposing: #143

pjones:
Superseded by xmonad/xmonad-contrib#143

Issue #76, opened 2016-08-24
X.L.PerWorkspace possible bug


davama:
Hello, 
Had this issue on xmonad mailing list [here](https://mail.haskell.org/pipermail/xmonad/2016-March/015078.html)
Not sure really if it's a `PerWorkspace` bug but that was the final conclusion.

So basically to summarize what happened on that chain. I was having issues with `avoidStruts` under [git-xmonad](https://github.com/xmonad/xmonad) but nothing of that sort using [haskell-xmonad](https://hackage.haskell.org/package/xmonad) (using `cabal install`).

Mr. Bogdan suggested to remove `PerWorkspace` all together, which i did but no change. (changes [here](https://gist.githubusercontent.com/davama/c380cb816a0b578f94ae/raw/da796fef72bd382bdea3b9f117ecb42728cf4052/xmonad.hs-without-perworkpace-module)) (currently using PerWorkSpace under haskell-xmonad; willing to test again)

When i use haskell-xmonad `avoidstruts` and `ToggleStruts` seems to work fine. Under git-xmonad they do not. My dzen bars are covered by the window clients on all WS (visible, non-visible).

My config is [here](https://github.com/davama/xmonad/blob/master/xmonad/xmonad.hs).

Decided to bring this issue to GIT since it's much easier to read and maybe more eyes will look at it.

Hope this is clear.

Thanks,
Dave

geekosaur:
What problems were you having, exactly?

Removing PerWorkspace pretty much rules it out as a cause, so I don't quite see how it could have been concluded that it was at fault. ManageDocks (which provides `avoidStruts`) did change in 0.12, and there's at least one significant bug in it that is not in git.

I also note that your "remove PerWorkspace" version unconditionally applies `avoidStruts` to all layouts including the fullscreen layout; if you were expecting it to not apply there (I can't tell since you didn't say what issues you were having), reconsider where you are putting `avoidStruts`.

davama:
Basically when under git-xmonad my dzen bars are covered by the window clients and `ToggleStruts` no longer works. This does not happen when using haskell-xmonad (bars are respected and togglestruts works). 

I had gone through a whole session on the mailing list trying to figure out what it was exactly (back in March) and that's where things were left. (`PerWorkspace` being the issue)

My current xmonad.hs file uses `PerWorkspace` which is in my profile. (link on first post)
my current layouts looks like so:
```
[...]

PR #80, opened 2016-09-02
Make strut cache global

f1u77y:


liskin:
Thanks! Please have a look at https://github.com/liskin/xmonad-contrib/commits/managedocks-global-cache, I made some additional changes, improved your commit message and added another commit.

f1u77y:
@liskin I agree with almost all of your improvements except using `refresh` instead of sending messages. The point is that sending messages to inactive layouts, it's just useless(sorry, my fault), that's why I think the most right way is to use only `sendMessage`.

liskin:
Well, I did some research and found no modules in xmonad-contrib that use messages solely for the purpose of refreshing the layout without carrying a specific change to its configuration, while there are some that use `refresh`. That's why I used `refresh`, since it's clearly simpler (by lines of code, at least).

Regarding #76, I have no idea. Does `PerWorkspace`actively prevent messages being sent to the layout when it's active? Then it's a `PerWorkspace` bug. But it seems strange.

f1u77y:
I think code is cleaner when something is run only when it needs to be run.

> Does `PerWorkspace` eactively prevent messages being sent to the layout when it's active?

It prevents messages being sent to the layout when it has never been active because it retranslates message to its first sub-layout when it hasn't been run initially while it should retranslate it to the second one.

liskin:
> I think code is cleaner when something is run only when it needs to be run.

Okay, whatever. :-)

> It prevents messages being sent to the layout when it has never been active because it retranslates message to its first sub-layout when it hasn't been run initially while it should retranslate it to the second one.

Oh. Well, I never claimed that #76 was fixed by this, anyway. Point c) in https://github.com/liskin/xmonad-contrib/commit/fab2c2ad0e1720cd97d27a41c14b30ef9464b4d8 is a different issue, one that is now fixed indeed.

lakshmivyas:
@f1u77y , @liskin - Thank you, both.  I just applied the patches (from @liskin's branch) and everything is awesome again :+1:

davama:
Any chance this could be merged?
Thanks

byorgey:
I don't really understand the issues or changes here, but if everyone involved is happy with it and you just need someone with a commit bit to click "merge" I would be happy to.  I'll merge in a few days unless I hear any objections before then.

f1u77y:
@byorgey c48d81e37845184b62623aa08545e23859828f86 is just an enchancement that keeps docks state in `StateExtension` instead of per-layout state. It should not fix any issues(but it fixes one caused by X.L.Perworkspace bug(#76); will create a ticket with more information ASAP).

Another(e38fb3bdb8bba7e09cc598fd37a06545ce18229f) enchancement is described well enough in commit message, I think.

(oh, and how did I not notice that this StateExtension should be persistent; will fix)

byorgey:
OK, thanks, makes sense.  I'll wait for the `StateExtension` fix then, ping me when you think it's ready to be merged.

liskin:
Oh no, no no no.

c48d81e is everything but just an enhancement. That's the whole fix. And it too has a commit message that explains what is fixed and how.

And no, it should definitely not be persistent. If the strut cache somehow gets out of sync, an xmonad restart is all that needed to reset it. If you make it persistent, that won't be possible. Also, there's really no point in making a **cache** persistent, is it?

liskin:
(So just to avoid any misunderstandings, I do think that it is currently ready to be merged and any additional changes may make it not ready. I did express this opinion by https://github.com/xmonad/xmonad-contrib/pull/80#pullrequestreview-120363)

f1u77y:
@liskin 
Ok got it, and I've forgot about `docksStartupHook` that makes `PersistentExtension` completely useless. 

I also think that `modifyXS` should be moved to `X.U.ExtensibleState` because it's more common then just for internal usage in this module(I used exatly that function for fixing X.L.Minimize at least)

f1u77y:
@byorgey it's ready to be merged

geekosaur:
On Mon, Oct 24, 2016 at 6:29 PM, Bogdan Sinitsyn <notifications@github.com>
wrote:

> @byorgey <https://github.com/byorgey>
>
> c48d81e
> <https://github.com/xmonad/xmonad-contrib/commit/c48d81e37845184b62623aa08545e23859828f86>
>
> is just an enchancement that keeps docks state in StateExtension instead
> of per-layout state.


This actually trades one bug for another: currently you lose the strut
cache, and therefore existing struts, on mod-shift-space.
This would stop that, but also break the workaround that replaces the
binding with one that rebuilds the cache afterward: IIRC it would keep the
old struts because it can't be cleared from the outside, and then merge a
fixed one, leaving you with a semi-broken mixture.

-- 

f1u77y:
@geekosaur as liskin mentioned, rebuilding of the cache now can be achieved by restarting xmonad

f1u77y:
As liskin mentioned, rebuilding of the cache now can be achieved by 
restarting xmonad

PR #85, opened 2016-09-15
Ignore the namedscratchpad workspace in DynamicWorkspaceOrder extension.

lakshmivyas:
NamedScratchpads are ubiquitous.  It'd be nice to have a "Secret"
workspace bucket in XMonad core so that all extensions can ignore
those workspaces (scratch pads or otherwise).  Since that feature
is not available, adding an explicit dependency seems like the
approach with least friction for now.


byorgey:
I'm not sure how I feel about this.  It's a bit confusing that, for example, `getSortByOrder` will still include the named scratchpad workspace.  So e.g. if someone uses `getSortByOrder` to generate a list of workspaces for their status bar, the NSP workspace will show up, but then with this patch functions like `withNthWorkspace` will ignore it.   Also, maybe some people actually like/depend on the current behavior?  Who knows.  I don't like this sort of ad-hoc one-off change to the behavior of an existing library, unless it is clearly fixing a bug.  Is there a different way we could structure this to make the new behavior opt-in?

lakshmivyas:
@byorgey - I agree - It is a definite hack.  Closing this for now.

PR #86, opened 2016-09-19
Export XMonad.Actions.Workscreen (WorkscreenId)

aiya000:
This is fixing for problem that
WorkscreenId was shown in XMonad.Actions.Workscreen document
(ex: viewWorkscreen),
but never shown WorkscreenId definition.

Is this the intension ?


nlewo:
I don't think it was intentional and it's better to export it.

LGTM

aiya000:
Thanks :smile:

byorgey:
Thanks!

PR #88, opened 2016-10-01
Added Super to mkKeyMap keys

elvis-sik:
I am using this :smile:

### DIFF
--- a/XMonad/Util/EZConfig.hs
+++ b/XMonad/Util/EZConfig.hs
@@ -194,6 +194,7 @@ removeMouseBindings conf mouseBindingList =
 -- > <Insert>
 -- > <Break>
 -- > <Space>
+-- > <Alt>
 -- > <Super>
 -- > <F1>-<F24>
 -- > <KP_Space>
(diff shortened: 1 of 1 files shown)


elvis-sik:
Added `xK_Alt_L` as well. It must be mentioned that both `alt` and `super` have left and right variants. I called `xK_Alt_L` and `xK_Super_L` simply `Alt` and `Super`, respectively.

byorgey:
Could you explain how exactly are you using this?  On first glance it seems a bit strange to have modifier keys in the list of special keys.

elvis-sik:
I stopped using it on that same day because [it turned out not to be a reliable way to get the behavior I wanted](http://unix.stackexchange.com/questions/313715/xmonad-control-timing-between-sequential-keybindings/).

Anyway, I was inspired by [Spacemacs](https://github.com/syl20bnr/spacemacs), which I like a lot :smile: 

Spacemacs is based around `evil mode` in Emacs, so we edit vim-style in it, but every command in (vim) `normal mode` is prefixed by `SPC` (akin to `leader` in vim). So to go the left window (which would be called a buffer in vim), I press `SPC w h`, sequentially.

My original idea was to replicate this in my WM, and press `Super w h` to go left. As I couldn't find out how to do this kind of sequential keybinding using that `(xK_Alt_L, xK_p)` interface, I decided to hack this PR quickly to prototype my idea.

In the end, after experimenting a bit, I found it awkward, mostly because of the issue I explained in the stack exchange question, but also because `Super`'s placement in the keyboard is much more uncomfortable than `Space`, in my opinion.

byorgey:
I see.  Yes, I can see how that spacemacs-like behavior would be nice if you like that sort of interface, but offhand I do not know how to achieve this in xmonad.  So, should this PR be closed?

PR #93, opened 2016-10-13
Added XMonad.Hooks.FocusGuard

SirBoonami:
Disclaimer: I have positively no idea what I'm doing as my experience with X11 is absolutely zero. This module is the result of an afternoon of poking around in other source code and testing the outcome, and it seems to work well for me. From an expert's point of view it might be incredibly stupid.

### DIFF
--- a/XMonad/Hooks/FocusGuard.hs
+++ b/XMonad/Hooks/FocusGuard.hs
@@ -1,0 +1,83 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module       : XMonad.Hooks.FocusGuard
+-- Copyright    : (c) 2016 Felix Hirn <felix.hirn@wyvernscale.com>
+-- License      : BSD
+--
+-- Maintainer   : Felix Hirn <felix.hirn@wyvernscale.com>
+-- Stability    : unstable
+-- Portability  : unportable
+--
+-- Filters out certain events by certain windows to prevent them from
+-- stealing focus. Especially useful when using some variation of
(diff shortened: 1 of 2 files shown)


SirBoonami:
I'll have to retract this for now as I've run into some issues with it. Sometimes it's pretty hard to tell if a change is really active or not with the length and complexity of the toolchain involved.

byorgey:
OK, let us know if you get it working!

SirBoonami:
Okay, I finally got around to extensively testing it and the issues I encountered were unrelated.

I realize that in the meantime, a PR with a similar goal, but different approach (#110 by @liskin, marking the focus-grabbing window in question as urgent instead of selecively ignoring requests from certain applications) has been made, but I think my solution still provides a desirable alternative for more obnoxious cases.

Merging both hooks into a single module might be even better though, what do you think?

liskin:
Actually your code does almost exactly the same thing as mine. How did I not notice this pull request here? :-(
(To my defense, I had the code for years, just last month decided it's time to clean it up and push upstream.)

Anyway, I intend to read all @geekosaur's suggestions on my pull requests and write some new code. Choosing between ignoring windows and marking them urgent will definitely be there. But I can't promise when I get to it.

SirBoonami:
I closed it a while ago because I ran into some issues, which were, as it turned out, unrelated, so I just reopened it. I think the best solution for now is to wait for your PR to be accepted, then add my hook (maybe with some slight modifications) on top.

SirBoonami:
Abandoned in favor of #128.

PR #95, opened 2016-10-16
Feature/add submapdefaultwithkey

Fuco1:
This is useful for when we want to decide what to do in the default
action based on the key that failed to match in the submap.


byorgey:
Nice idea, thanks!

Fuco1:
Thanks :) I have another feature lined up which depends on this.  PR is coming!

byorgey:
Great, looking forward to it. =)

PR #97, opened 2016-10-19
fix broken XPConfig docs

samdoshi:
The docs for `autoComplete` and `showCompletionOnTab` look like they've been jumbled up.

https://hackage.haskell.org/package/xmonad-contrib-0.12/docs/XMonad-Prompt.html#t:XPConfig


byorgey:
Good catch, thanks!

PR #98, opened 2016-10-21
X.A.WindowBringer: Add a config object with a X.C.Default implementation

IvanMalison:
This provides a less complicated interface to the specification of custom behavior. In particular it allows the specification of a custom window titling function.

This is the first non-trivial piece of haskell I've ever written, so comments (even nit-picky ones) are welcome. I'm somewhat embarrassed to admit that I was actually unable to run this code because I couldn't figure out how to make my from compiled from source xmonad-contrib linkable. Specifically, I get errors like:

```
    35  17 error           • Couldn't match expected type ‘XConfig l0’
               with actual type ‘xmonad-0.12@xmonad-0.12-2GDBWwSJPDc58Xg9q7fgK4:XMonad.Core.XConfig
                                   a0’
   NB: ‘xmonad-0.12@xmonad-0.12-2GDBWwSJPDc58Xg9q7fgK4:XMonad.Core.XConfig’
         is defined in ‘XMonad.Core’
             in package ‘xmonad-0.12@xmonad-0.12-2GDBWwSJPDc58Xg9q7fgK4’
       ‘XConfig’
         is defined in ‘XMonad.Core’
             in package ‘xmonad-0.12@xmonad-0.12-9UBNv1UcrEbBAIGXxBv7Um’
 • In the second argument of ‘($)’, namely
     ‘ewmh
      $ pagerHints
          (def
             {modMask = mod4Mask, terminal = "urxvt",
              manageHook = manageDocks <+> manageHook def,
              layoutHook = myLayoutHook, logHook = myLogHook,
              handleEventHook = docksEventHook <+> fullscreenEventHook,
```

I suspect that this has to do with the fact that I somehow compiled xmonad-contrib against a different version of xmonad, but I'm not really sure how to fix this. I am cabal installing as follows 

```
cabal install --force-reinstalls --flags="-use_xft"
```

If anyone could help me out with that I would greatly appreciate it. In any case, everything seems to compile and I'm hopefully that this functions as I expect it to.

### DIFF
--- a/XMonad/Prompt.hs
+++ b/XMonad/Prompt.hs
@@ -149,8 +149,8 @@ data XPConfig =
         , changeModeKey     :: KeySym     -- ^ Key to change mode (when the prompt has multiple modes)
         , defaultText       :: String     -- ^ The text by default in the prompt line
         , autoComplete      :: Maybe Int  -- ^ Just x: if only one completion remains, auto-select it,
-        , showCompletionOnTab :: Bool     -- ^ Only show list of completions when Tab was pressed
                                           --   and delay by x microseconds
+        , showCompletionOnTab :: Bool     -- ^ Only show list of completions when Tab was pressed
         , searchPredicate   :: String -> String -> Bool
                                           -- ^ Given the typed string and a possible
                                           --   completion, is the completion valid?


IvanMalison:
@byorgey I've managed to compile everything and make sure it is working. Any chance you can take a look at this?

byorgey:
Sure, I will try to take a look.

XMonad/Actions/WindowBringer.hs
```diff
@@ -45,83 +46,106 @@ import XMonad.Util.NamedWindows (getName)
...
+data WindowBringerConfig = WindowBringerConfig { menuCommand :: String
+                                               , menuArgs :: [String]
+                                               , windowTitler :: X.WindowSpace -> Window -> X String
+                                               }
 
--- | Make dmenu case insensitive
-defaultArgs :: [String]
-defaultArgs = ["-i"]
+instance Default WindowBringerConfig where
+    def = WindowBringerConfig{ menuCommand="dmenu"
```
byorgey: It's customary style to put spaces on either side of `=`.  Personally, I think it also looks nice to vertically align the `=` signs (though it's not necessary).

XMonad/Actions/WindowBringer.hs
```diff
@@ -45,83 +46,106 @@ import XMonad.Util.NamedWindows (getName)
 -- For detailed instructions on editing your key bindings, see
 -- "XMonad.Doc.Extending#Editing_key_bindings".
 
--- | Default menu command
-defaultCmd :: String
-defaultCmd = "dmenu"
+data WindowBringerConfig = WindowBringerConfig { menuCommand :: String
```
byorgey: It would be good to include some Haddock documentation here, both on the data type itself and on the individual fields.  Maybe something like this:
```
-- | Configuration record for WindowBringer functions.
data WindowBringerConfig = WindowBringerConfig
  { menuCommand   :: String     -- ^ The shell command used to xyz.
  , menuArgs      :: [String]   -- ^ Foo bar
  ...
```

I personally like to put the opening brace indented on a separate line like this, it results in more consistent indentation and leaves more horizontal space for including comments and such.

XMonad/Actions/WindowBringer.hs
```diff
@@ -45,83 +46,106 @@ import XMonad.Util.NamedWindows (getName)
...
-       keyValuePair ws w = flip (,) w `fmap` decorateName ws w
+       keyValuePair ws w = flip (,) w `fmap` titler ws w
 
 -- | Returns the window name as will be listed in dmenu.
 --   Tagged with the workspace ID, to guarantee uniqueness, and to let the user
 --   know where he's going.
 decorateName :: X.WindowSpace -> Window -> X String
 decorateName ws w = do
-  name <- fmap show $ getName w
+  name <- show `fmap` getName w
```
byorgey: If you're using `fmap` infix, you should use `(<$>)` instead.  (I see there were already some infix `fmap`s in this code before, perhaps it was written before `(<$>)` became common.)

XMonad/Actions/WindowBringer.hs
```diff
@@ -45,83 +46,106 @@ import XMonad.Util.NamedWindows (getName)
...
+windowMap = windowMap' decorateName
+
+-- | A map from window names to Windows, given a windowTitler function.
+windowMap' :: (X.WindowSpace -> Window -> X String) -> X (M.Map String Window)
+windowMap' titler = do
   ws <- gets X.windowset
   M.fromList `fmap` concat `fmap` mapM keyValuePairs (W.workspaces ws)
  where keyValuePairs ws = mapM (keyValuePair ws) $ W.integrate' (W.stack ws)
-       keyValuePair ws w = flip (,) w `fmap` decorateName ws w
+       keyValuePair ws w = flip (,) w `fmap` titler ws w
```
byorgey: I think `flip (,) w` is more legibly written `(,w)` (by turning on the `TupleSections` language extension).

XMonad/Actions/WindowBringer.hs
```diff
@@ -45,83 +46,106 @@ import XMonad.Util.NamedWindows (getName)
...
 
 -- | Calls dmenuMap to grab the appropriate Window, and hands it off to action
 --   if found.
-actionMenu :: String -> [String] -> (Window -> X.WindowSet -> X.WindowSet) -> X ()
-actionMenu menuCmd menuArgs action = windowMap >>= menuMapFunction >>= flip X.whenJust (windows . action)
+actionMenu :: WindowBringerConfig -> (Window -> X.WindowSet -> X.WindowSet) -> X ()
+actionMenu WindowBringerConfig{ menuCommand = cmd
+                              , menuArgs = args
+                              , windowTitler = titler} action = windowMap' titler
+               >>= menuMapFunction >>= flip X.whenJust (windows . action)
```
byorgey: I'm not a big fan of the indentation/layout here, but I also can't think of anything else I really like.  Matching on the giant record argument (though entirely necessary) makes things awkward.  Perhaps just move `= windowMap' titler` down onto the next line so it doesn't get lost in the noise.

XMonad/Actions/WindowBringer.hs
IvanMalison: I had it as `<$>` before and it failed to build on older versions of haskell. Is there some import that is needed in older versions?

XMonad/Actions/WindowBringer.hs
IvanMalison: I didn't like this either, and I was hoping there might be a better way, I guess there just isn't.

IvanMalison:
Thanks for the review. I'll implement your suggestions right now and let you know when I'm done.

IvanMalison:
Note that 067004a fails CI while 05d7493 does not. Do I need to do something to make <$> work in older versions of ghc?

f1u77y:
@IvanMalison `import Data.Functor((<$>))`

IvanMalison:
Okay looks like the soln was to import Control.Applicative((<$>))

IvanMalison:
@byorgey This should be ready for merge now. Thanks!

geekosaur:
On Fri, Oct 21, 2016 at 12:06 AM, Ivan Malison <notifications@github.com>
wrote:

> I suspect that this has to do with the fact that I somehow compiled
> xmonad-contrib against a different version of xmonad, but I'm not really
> sure how to fix this. I am cabal installing as follows


Make sure you do not have multiple installs of the xmonad package. This can
be caused by having both an OS package and a cabal package installed, or
one in global and one in user or sandbox. ghc handles this situation
terribly, and you get that type mismatch because it's trying to match the
XConfig from one of the installed xmonad libs with the one from the other.

(If you look closely, the two libraries have different hashes. This usually
reflects compiler-generated names that leaked into the .hi file for
cross-module inlining.)

-- 

geekosaur:
On Tue, Oct 25, 2016 at 12:06 AM, Ivan Malison <notifications@github.com>
wrote:

> I had it as <>gt; before and it failed to build on older versions of
> haskell. Is there some import that is needed in older versions?
>

Control.Applicative, before 7.10.


-- 

IvanMalison:
Oops, I didn't mean to close this. I think it was closed automatically when I removed the relevant commits from the master branch in my repo. Let me open another request.

PR #103, opened 2016-10-26
X.A.WorkspaceNames: Export WorkspaceNames type

IvanMalison:
The getWorkspaceNames names that is exported by this module adds the tag with a colon as a prefix.

With WorkspaceNames type itself users can access the actual names that are assigned to each workspace.


liskin:
Wouldn't this be a better abstraction: https://github.com/liskin/xmonad-contrib/commit/9ee89c429912483ca6f90e00eee89ed921b83696
(I've had that for a while, dunno why I didn't open a pullreq back then. I'll try to do it this weekend together with some other changes)

IvanMalison:
Closed in favor of #108

PR #106, opened 2016-10-27
X.A.WindowGo: Add arg for custom focus fn to raiseNextMaybe

IvanMalison:
Motivation: I prefer to use a greedier focus function that switches the workspace on the current screen to the focused window.

``` haskell
greedyFocusWindow w ws = W.focusWindow w $ W.greedyView
                         (fromMaybe (W.currentTag ws) $ W.findTag w ws) ws
```

This pull request exposes a version of raiseNextMaybe to which a custom focus function, such as the one that I prefer, can be provided.


byorgey:
I like the new function, but could you please add a Haddock comment to document it?

byorgey:
Thanks!

PR #108, opened 2016-10-29
X.A.WorkspaceNames: add get(Current)WorkspaceName

liskin:
(a cleaner version of #103, possibly)

### DIFF
--- a/XMonad/Actions/WorkspaceNames.hs
+++ b/XMonad/Actions/WorkspaceNames.hs
@@ -25,6 +25,8 @@ module XMonad.Actions.WorkspaceNames (
     renameWorkspace,
     workspaceNamesPP,
     getWorkspaceNames,
+    getWorkspaceName,
+    getCurrentWorkspaceName,
     setWorkspaceName,
     setCurrentWorkspaceName,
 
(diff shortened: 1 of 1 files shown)


IvanMalison:
Yeah agreed. Thanks!

byorgey:
Thanks!

IvanMalison:
@liskin @byorgey After trying to use this in my own config, I've decided that this interface is actually harder to use in some ways because it requires access to the xmonad at all times to get the name of any particular workspace. I don't really see the advantage of hiding the underlying data structure in this case, since by exposing it we can do a bunch of stuff in pure functions that way.

What about exposing a function that exposes the underlying mapping?

liskin:
Something like a `getWorkspaceNames :: X (WorkspaceId -> Maybe String)` ?

liskin:
Oh, there already is a function `getWorkspaceNames` that's `X (WorkspaceId -> String)`. Maybe that should've been called `getWorkspaceIdsWithNames` or something as it's highly specialized for my xmobar usecase.

IvanMalison:
getWorkspaceNames annoyingly adds the tag and a colon to the workspace name.

liskin:
Okay, see https://github.com/xmonad/xmonad-contrib/pull/114

Issue #36, opened 2016-02-08
X.L.LayoutBuilder should support resizing of subboxes


pjones:
This is a placeholder issue while I refactor the `X.L.LayoutBuilder` module.  Changes will include:
- Bring in features from `X.L.LayoutBuilderP` (done)
- Make layout boxes resizable with a message

My work is happening here:

https://github.com/pjones/xmonad-contrib/blob/xmonad/feature/layoutb/XMonad/Layout/LayoutBuilder.hs

I merged `X.L.LayoutBuilderP` into `X.L.LayoutBuilder` since they were nearly identical files and it doesn't make sense to add resizing to both.  `X.L.LayoutBuilderP` is now deprecated.

pjones:
I now think that `LayoutBuilder` should be deprecated so I'm closing this.

PR #113, opened 2016-11-05
Refactor all X.L.LayoutBuilderP functionality into X.L.LayoutBuilder

pjones:
X.L.LayoutBuilderP is nearly identical to X.L.LayoutBuilder.  Originally
I wanted to add the ability to dynamically resize the layout boxes so it
make a lot of sense to join these two modules together so I wouldn't
have to do it in both.  Even though I never got around to that I still
think it's a good idea to merge these two modules into one.

I believe I was able to merge these without creating any
backward-compatibility issues.  I've been sitting on these changes since
2015 and they work for me without having to change older parts of my
config (relating to X.L.LayoutBuilder).

If anyone wants to work on dynamically resizing layout boxes the issue I
created for it is #36.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,13 @@
     EWMH taskbars and pagers. Useful for `NamedScratchpad` windows, since
     you will usually be taken to the `NSP` workspace by them.
 
+### Minor Changes
+
+  * `XMonad.Layout.LayoutBuilder`
+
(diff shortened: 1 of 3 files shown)


byorgey:
Looks good to me, though I admit I can't really verify whether any backwards compatibility issues etc. have been introduced.  Hopefully other folks using LayoutBuilder will be able to test it before 0.13.

PR #121, opened 2016-11-26
X.A.DynamicWorkspaceNames: Add dynamic workspace names code

IvanMalison:
This module provides a LayoutModifier that automatically names workspaces based on their contents. It exports most of its functions so that users can easily write their own such layoutmodifiers to do something similar with a slightly different naming function.

### DIFF
--- a/XMonad/Actions/DynamicWorkspaceNames.hs
+++ b/XMonad/Actions/DynamicWorkspaceNames.hs
@@ -1,0 +1,73 @@
+{-# LANGUAGE TypeSynonymInstances, MultiParamTypeClasses, FlexibleInstances #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Actions.DynamicWorkspaceNames
+-- Copyright   :  (c) Ivan A. Malison
+-- License     :  BSD-style (see LICENSE)
+--
+-- Maintainer  :  <IvanMalison@gmail.com>
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- Automatically label workspaces using the WorkspaceNamesModule.
+-----------------------------------------------------------------------------
+module XMonad.Actions.DynamicWorkspaceNames
+  (
+   NameWorkspacesHook(..),
+   currentWSName,
+   updateWorkspaceName,
+   setWorkspaceNames,
+   getClass,
+   getNamesFromWindows,
+   getClassesNameForWorkspace
+  ) where
+
+import           Data.List
+import           Data.Maybe
+import           Graphics.X11.Xlib.Extras
+import           Control.Monad
+
+import           XMonad
+import qualified XMonad.StackSet as W
+import           XMonad.Actions.WorkspaceNames
+import           XMonad.Layout.LayoutModifier
+
+-- | Get the current name for the given workspace.
+currentWSName :: WindowSpace -> X String
+currentWSName ws = fromMaybe "" <$> (getWorkspaceNames' <*> pure (W.tag ws))
+
+-- | Update the name of a workspace only if it is not the same as the current
+-- name.
+updateWorkspaceName :: (WindowSpace -> X String) -> WindowSpace -> X ()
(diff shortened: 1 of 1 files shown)


IvanMalison:
If anyone could help me come up with a way to make NameWorkspacesHook hold the function that will be used to name workspaces I'd appreciate it. I think that it's likely impossible because NameWorkspacesHook to be a Read and Show instance and this is pretty hard when there is a field that is a function.

pjones:
You can't (easily) make a Read or Show instance if one of the fields is a function.  To do that you'd have to make a custom Show that doesn't export the function and a custom Read that sets the function to some default function.

Can you refactor the function out of the record?

IvanMalison:
@pjones It is factored out right now, its just that in order to supply a custom renaming strategy, the user will have to implement their own instance of the type class using the provided functions, which is unfortunate.

pjones:
@IvanMalison Please add an entry to `CHANGES.md` so we can get this reviewed and merged.  Thanks.

XMonad/Actions/DynamicWorkspaceNames.hs
```diff
@@ -0,0 +1,73 @@
...
+getClass w = fmap resClass $ withDisplay $ io . flip getClassHint w
+
+-- | Apply the given action to each window in the given workspace.
+getNamesFromWindows :: (Window -> X String) -> WindowSpace -> X [String]
+getNamesFromWindows windowNamer w = mapM windowNamer $ W.integrate' $ W.stack w
+
+-- | Build a '|' delimited string combining the class hints of the windows in
+-- each workspace.
+getClassesNameForWorkspace :: WindowSpace -> X String
+getClassesNameForWorkspace = (intercalate "|" <$>) . getNamesFromWindows getClass
```
liskin, line 63: I feel like this could use some sorting or at least deduplicating.

Or is it not needed in the intended usecase? I'm curious what this really looks like in your workflow. Would you mind sharing a screenshot or something? :-)

XMonad/Actions/DynamicWorkspaceNames.hs
```diff
@@ -0,0 +1,73 @@
...
+currentWSName :: WindowSpace -> X String
+currentWSName ws = fromMaybe "" <$> (getWorkspaceNames' <*> pure (W.tag ws))
+
+-- | Update the name of a workspace only if it is not the same as the current
+-- name.
+updateWorkspaceName :: (WindowSpace -> X String) -> WindowSpace -> X ()
+updateWorkspaceName nameWorkspace workspace = do
+  currentName <- currentWSName workspace
+  newName <- nameWorkspace workspace
+  when (currentName /= newName) $ setWorkspaceName (W.tag workspace) newName
```
liskin, line 45: This "only set if changed" functionality could go into `setWorkspaceName` itself.
And I feel like these `currentWSName`, `updateWorkspaceName` and `setWorkspaceNames` functions belong into `X.A.WorkspaceNames` rather than here.

liskin (commented):
This looks very good, just some minor nitpicks.

(And again, I would've looked at this earlier had I been pinged. ;-))

slotThe:
@IvanMalison friendly ping in case you missed the review :)

liskin:
Note that there is some overlap between this module and https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Hooks/DynamicIcons.hs. Right now it's not entirely straightforward to plug DynamicIcons into EwmhDesktops, so not all use cases of DynamicWorkspaceNames are covered, but once all the EwmhDesktops refactorings around #399 are finished, I'd expect DynamicIcons to be as powerful as this.

XMonad/Actions/DynamicWorkspaceNames.hs
IvanMalison: The idea was to be able to see exactly which windows were in each workspace.

IvanMalison:
I don't use this, and I haven't for a long time, so I'm just going to close.

PR #125, opened 2016-11-30
CustomKeys.hs: Fix documentation

vvv:
Fix code example in documentation. (Wrong implementation of `delkeys`.)
Thanks to Lasse R.H. Nielsen (@lrhn) for reporting the problem!


lrhn:
Thank you. The version without `zip` now works for me.

PR #126, opened 2016-12-02
Fix #120 - Make Actions.WindowGo.raiseNextMaybe span over all workspaces

### LINKED ISSUE #120: Actions.WindowsGo/raiseNextMaybe cycles between 2 workspaces only
As reported some years ago in https://code.google.com/archive/p/xmonad/issues/284 raiseNextMaybe cycles only between 2 workspaces, making impossible to cycle through the entire list of windows.

The reason is that raiseNext calls allWindows from xmonad StackSet, which relies on a list of workspaces built as follows: the current workspace is put first, followed by visible then hidden workspaces. Thus the ordering of workspaces is not preserved while switching between workspaces, which is why raiseNext will 
cycle only between the current and another workspace.

I found the following fix:
- create a allWindowsSorted function in Xmonad.StackSet which relies on a list
of workspaces sorted by their tags
- call W.allWindowsSorted instead of W.allWindows in raiseNext.

The patch is given below for 0.12. Let me know if it is an appropriate way to fix the bug. If so, I can create 
the pull requests for xmonad and xmonad-contrib that apply on current master.
Otherwise, let me know what solution you would prefer. 
It would be nice that 0.13 contains a fix.

[xmonad-fix-raiseNextMaybe.patch](https://github.com/xmonad/xmonad-contrib/files/605921/xmonad-fix-raiseNextMaybe.patch.txt)

pauleve:
As discussed in #120, this patch introduces workspaceSorted and allWindowsSorted to provide an ordering of windows which does not change when changing the current workspace.

### DIFF
--- a/XMonad/Actions/WindowGo.hs
+++ b/XMonad/Actions/WindowGo.hs
@@ -38,13 +38,14 @@ module XMonad.Actions.WindowGo (
 
 import Control.Monad
 import Data.Char (toLower)
+import qualified Data.List as L (nub,sortBy)
 import Data.Monoid
 import XMonad (Query(), X(), ManageHook, WindowSet, withWindowSet, runQuery, liftIO, ask)
 import Graphics.X11 (Window)
 import XMonad.ManageHook
 import XMonad.Operations (windows)
 import XMonad.Prompt.Shell (getBrowser, getEditor)
-import qualified XMonad.StackSet as W (allWindows, peek, swapMaster, focusWindow)
+import qualified XMonad.StackSet as W (peek, swapMaster, focusWindow, workspaces, StackSet, Workspace, integrate', tag, stack)
 import XMonad.Util.Run (safeSpawnProg)
 {- $usage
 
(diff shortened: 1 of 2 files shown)


pauleve:
Thanks for your feedback. I fixed the indentation, and added a comment to explain the purpose of workspacesSorted.

XMonad/Actions/WindowGo.hs
```diff
@@ -66,12 +67,20 @@ appropriate one, or cover your bases by using instead something like:
 For detailed instructions on editing your key bindings, see
 "XMonad.Doc.Extending#Editing_key_bindings". -}
 
+workspacesSorted :: Ord i => W.StackSet i l a s sd -> [W.Workspace i l a]
+workspacesSorted s = L.sortBy (\u t -> W.tag u `compare` W.tag t) $
+                       W.workspaces s
```
pauleve: Thanks for your feedback. I fixed the indentation, and added a comment to explain the purpose of workspacesSorted.

pjones:
Please include an entry in `CHANGES.md` and I will merge.

pjones:
Thank you!

PR #127, opened 2016-12-09
DynamicProperty: execute other hooks

bennofs:
All False short-cuts the default behavior for the event, which leads to a non-functioning window manager. Returning mempty ensures that the default action is still executed,


geekosaur:
Committed, thanks!

geekosaur:
...tbh I hadn't realized that had gone in; I still had it flagged locally for verification and submission >.>

PR #128, opened 2016-12-11
Add new module XMonad.Hooks.Focus .

sgf-dma:
Hi.

So, i've fixed documentation and here i am, trying to jump into leaving train
again.

Here is another way to solve #110, #93 and
https://github.com/xmonad/xmonad/issues/45 : a module, which extends
ManageHook EDSL with combinators working on focused window and current
workspace. Please, see documentation for examples and detailed description in
the module for more details.

I apologize to authors of two proposed above solutions for releasing this so
late, but this module provides a much more generic way to solve focus grab
problem. E.g. it allows to move activated window to current workspace, but not
grab focus, or grab focus, but not switch workspace.

Unfortunately, it has some pitfalls:

1. I've decided to apply `manageHook` to activated window too. It seems very
   natural to me, but it may lead to unexpected results for many people using
   `ewmh`, when `manageHook` previously working only for new windows, start
   working for activated windows too. It may be solved, by adding

        not <$> activated

    before those part of `manageHook`, which should not be called for
    activated windows.  But this lifts `manageHook` into `FocusHook` and it
    needs to be converted back later.
2. *(fixed)* This module is incompatible with `XMonad.Hooks.EwmhDesktops` in its current
   state, because `ewmh` also handles activated window and it'll overwrite any
   decisions made by `FocusHook`. And i think, the fixes proposed above are
   not very good: they effectively disable entire `handleEventHook` for
   activated window, but i'd prefer to disable only handling in `ewmh` (e.g.
   provide a different `handle`, whatever). Though, even proposed fixes can be
   used, if they'll be applied in correct order:

        handleFocusQuery ..
          . ignoreNetActiveWindow ..
          . ewmh

    so may be it's fine.
3. *(fixed)* It requires more general versions of `XMonad.Hooks.ManageHelpers`
   combinators, then currently provided. E.g. now we have

        composeOne :: [MaybeManageHook] -> ManageHook

    but i need

        composeOne :: (Monoid a, Monad m) => [m (Maybe a)] -> m a

    This may be easily fixed, though.
4. *(fixed)* `handleFocusQuery` adds `modMask` automatically to focus lock key.
   Probably, that should be changed (to not adding `modMask`) for consistency.
5. There is no tests even for pure functions yet.
6. *(fixed)* Probably, i shouldn't overwrite `_NET_WM_NAME` and just set it, if it's
   missed.

I can fix 3, 4 and 6 in a few days. The 1 is a decision on which entire module
is built and can be either accepted or not (with the module itself). The 2 needs
a decision to be made. I've cc-ed all participants of above issues.

@SirBoonami , @liskin , @unclechu , please, try this module for your use case.
Does it solve your problem?

/cc @liskin @geekosaur  @pjones  @f1u77y  @byorgey  @SirBoonami  @unclechu


unclechu:
@sgf-dma thanks! It works for me!
https://github.com/xmonad/xmonad/issues/45#issuecomment-266317556

sgf-dma:
I've pushed another commit, adding function `getWMName` to X.H.SetWMName and fixing point 6 from the first post: do not overwrite wm name, if user has set it.

sgf-dma:
Another commit changing the type of ManageHook combinators and fixing point 3 above.

I'll add edit the CHANGES.md, when all fixes will be here. I hope, that will be soon.

sgf-dma:
I've fixed the point 4 above (about not adding 'modMask' to key for toggling focus lock in `handleFocusQuery`), added changes into CHANGES.md and rebased all commits. The build fails, but i have no idea why.

I do not plan any more changes to X.H.Focus. Please, review the code.

And the only point left - is incompatibility with `X.H.EwmhDesktops`. I've seen several PRs modifying `EwmhDesktops` and discussions about rewriting it. So.. i'm unsure should i and how particularly fix this.

geekosaur:
On Thu, Dec 15, 2016 at 2:24 PM, sgf-dma <notifications@github.com> wrote:

> The build fails, but i have no idea why.


Spurious failures: it's unclear where the problem is, but it's fairly
common for Travis to encounter network failures while setting up the build
container. The logs for the 7.6.3 and 7.8.1 builds both show it failing to
install packages from Ubuntu's repo.

-- 

sgf-dma:
Finally, i think, i've found a way of fixing incompatibility with `X.H.EwmhDesktops`: i moved all ewmh-related code from `X.H.Focus` to `X.H.EwmhDesktops`. And now `manageHook` is applied to activated window. That means:
1. Window activation in `ewmh` do nothing by default (with empty `ManageHook`).
2. You can change activation behavior using `ManageHook` combinators. Now `X.H.EwmhDesktops` provides additional combinator `activated` to match on activated window.
3. You can use `X.H.Focus` combinators to create more complex focus/workspace switch strategies.

@unclechu , please, test your config with current master. Note the changes:
1. The type of `activated` combinator has changed from `FocusQuery Bool` to `Query Bool` and its now exported by `X.H.EwmhDesktops`.
2. There is no `handleFocusQuery` function now. You should use `ewmh` instead (to announce _NET_ACTIVATE_WINDOW and handle an event) and just add `FocusHook` (converted to `ManageHook` by `manageFocus`) to your `manageHook` (usually, it should be the leftmost).

I hope, there will be no more rebases.

And we may go even further with this: add `netActivated` property, which is now the value stored in extensible state, to `Query` and handle window activation in xmonad. That way only `xmonad` will call `manageHook`. Probably, that'll be more consistent.

unclechu:
@sgf-dma is `handleFocusQuery` still exist?
https://github.com/xmonad/xmonad-contrib/compare/master...sgf-dma:master#diff-8b1c3fd0d4a6765c16dfd18509182f9dR105

unclechu:
@sgf-dma it works, I did [some changes in my config for new API](https://github.com/unclechu/xmonadrc/commit/d821e7757706ea73a5ba94a6e03367c81a08a94a).

sgf-dma:
@unclechu , thanks!

I've removed no longer relevant changes from CHANGES.md (mentioning `handleFocusQuery`) and tried to resolve merge conflict. And though the last have failed, now the resolving is straightforward: just use `theirs` version.

pjones:
@sgf-dma What is the status of this PR?  I'm thinking of waiting until after v0.13 to merge this.

sgf-dma:
I've fixed all issues i was aware of and was able to fix (see first post). But since there was a little feedback, all these issues was mine own. Anyway, for now, i think i've finished this PR (well, apart from conflicting CHANGES, but there is a little sense in fixing it until actual merge).

SirBoonami:
Sorry for not reacting to this earlier, I've been way too busy the last few months.

Anyway, I've looked over your code and while I've yet to test it, it looks pretty good so far. There are many different possible ways to handle window activation, which is always good.

I'd suggest adding a few drop-in hook presets (like the ones [DynamicLog](http://hackage.haskell.org/package/xmonad-contrib-0.13/docs/XMonad-Hooks-DynamicLog.html) provides) for the most common use cases, as I've found them to be

* Handy building blocks for your `xmonad.hs` if you just want a quick, generic solution without looking into the inner workings of the module,
* Good stepping stones to understand how the module is supposed to be used by looking at their implementation.

You've already provided a bunch of examples in the description text, it should be easy to turn them into presets and maybe put their explanations into regular comments so that newbies (like me lol) aren't immediately overwhelmed by your module when they open its documentation.

sgf-dma:
I've added more examples and predefined configurations:

- `activateSwitchWs` - default EWMH behavior (switch workspace and focus).
- `activateOnCurrentWs` - bring activated window on current workspace and switch focus.
- `activateOnCurrentKeepFocus` - the same, but leave focus unchanged.

@SirBoonami As i understand your module, it's equivalent to something like

    not <$> (className =? ...) --> activateSwitchWs

SirBoonami:
Alright, I tested your changes on three different settings:

* Ignore all focus events (new default)
* Pull the window's workspace to the front (old default)
* Ignore requests for some nasty windows (my approach)

Ignoring the three hours spent in cabal hell, everything went smoothly and behaved exactly as expected. From what I can tell, this seems ready for production.

byorgey:
So it seems there is a lot of momentum behind this PR, several users who were complaining about/working on similar things have endorsed it, and it seems well thought out and well-documented as far as I can tell.  I just don't think I have the time/knowledge/desire to test it myself.  Are there any remaining objections to merging this?  ( @geekosaur , @pjones , what do you think? )

pjones:
I don't have any issues with this PR.  I think it's a great idea.

byorgey:
Merged!  Thanks @sgf-dma and others for the hard work putting this together and testing it.

PR #131, opened 2016-12-25
Fix CHANGES.md after b9d8f6c .

sgf-dma:
Fix CHANGES.md after "X.A.Submap: establish pointer grab to avoid freezing X."


pjones:
Thanks!

PR #147, opened 2017-02-14
New function: `X.P.Window.windowMultiPrompt'

pjones:
### Description

Like 'windowPrompt', but uses the multiple modes feature of
@Prompt@ (via 'mkXPromptWithModes').

Given a list of actions along with the windows they should work
with, display the appropriate prompt with the ability to switch
between them using the @changeModeKey@.

For example, to have a prompt that first shows you all windows, but
allows you to narrow the list down to just the windows on the
current workspace:

> windowMultiPrompt config [(Goto, allWindows), (Goto, wsWindows)]

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,25 @@
 # Change Log / Release Notes
 
+## 0.14 (No Release Date Announced)
+
+### Breaking Changes
+
+These are changes that users should review and prepare for before
+upgrading XMonad.
+
+### New Modules
+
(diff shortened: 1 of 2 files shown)


byorgey:
This looks good to me.  The problem with requiring users to edit `CHANGES.md` is that it creates merge conflicts whenever any PRs are submitted in parallel (*i.e.*, all the time).  It's annoying, but on the other hand I'm not sure what a better solution would be.

PR #159, opened 2017-03-28
Add Emacs-style Prefix handling

Fuco1:
### Description

This patch adds an Emacs-like prefix argument and a wrapper function you can use to turn any `X` action into one aware of the prefix.  See the Usage notes in the file for more details.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -305,6 +305,11 @@
     that are exclusive on the same screen. It also allows to remove this
     constraint of being mutually exclusive with another scratchpad.
 
+  * `XMonad.Actions.Prefix`
+
+    A module that allows the user to use an Emacs-style prefix
+    argument (raw or numeric).
+
 ### Bug Fixes and Minor Changes
 
   * `XMonad.Layout.Tabbed`
--- a/XMonad/Actions/Prefix.hs
+++ b/XMonad/Actions/Prefix.hs
@@ -26,6 +26,8 @@ module XMonad.Actions.Prefix
        , usePrefixArgument
        , useDefaultPrefixArgument
        , withPrefixArgument
+       , isPrefixRaw
+       , isPrefixNumeric
        , ppFormatPrefix
        ) where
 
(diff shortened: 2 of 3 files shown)


pjones:
Pretty interesting idea @Fuco1.  With the default key binding being C-u, doesn't that mean we'd break Emacs since there would be no way to use C-u?

Fuco1:
That's the neat part! Unless there is a keybinding registered with xmonad that handles the prefix, xmonad recognizes there's no way to continue and just sends `C-u` to the window that would normally recieve it.  That's one of the reasons I made the `submapWithDefaultKey` patch some months ago.

So if you have `M-k` in xmonad but not in Emacs, `C-u M-k` would pass the prefix to whatever is bound to `M-k`.  If there is no `M-k` in xmonad, these two would be send to Emacs (in the expected order).

There is one slight problem when you have a map like `C-c C-u` in Emacs (for example org mode uses this), then you have to hit `C-g` to get out of the "xmonad grab" (or any other non-recognized key really) to get `C-c C-u` sent to Emacs.

Fuco1:
I've been using this for about three months now, here are some examples from my current config if people are interested about the possibilities: https://github.com/Fuco1/dotfiles/blob/master/xmonad/.xmonad/xmonad.hs#L117

pjones:
Perhaps you can add a link to your config in the docs for this module.

IvanMalison:
> There is one slight problem when you have a map like C-c C-u in Emacs (for example org mode uses this), then you have to hit C-g to get out of the "xmonad grab" (or any other non-recognized key really) to get C-c C-u sent to Emacs.

Oooh that's a pretty nasty gotcha. What would be the harm in using a different prefix key by default in xmonad? Super-u isn't defined by default...

Fuco1:
@IvanMalison no harm I guess, I just wasn't sure what to use and `C-u` is readily understandable for most potential users (which I guess would come from Emacs).  We can change it of course.

Personally, the mentioned `C-c C-u` of org is likely the only "offending" binding I've encountered, mostly because people somehow subconsciously avoid binding `C-u` to anything, even in nested maps.  But it caught me by surprise the first time, I have to admit :D

I should probably add a note about this to the readme, as this shadowing can theoretically happen with any binding in any application.  Or we can figure out some way to make it a non-issue, but no solution comes to mind right now.

slotThe:
@psibi This pr seems done, I guess @Fuco1 can just rebase once more and we merge it before 0.17?  Seems like it simply got forgotten about by the people involved.

Fuco1:
fwiw I've been using this ever since without any problems.  I can attempt a rebase, hopefully there won't be too many conflicts.

psibi:
@slotThe Thanks for the ping. I will merge this in a couple of days unless there is objection. Also, I think two things needs to be done before we merge it:

- [ ] Rebase and fix the current conflict
- [ ] Add the following gotcha in the docs:

```
There is one slight problem when you have a map like C-c C-u in Emacs (for example org mode uses this), then you have to hit C-g to get out of the "xmonad grab" (or any other non-recognized key really) to get C-c C-u sent to Emacs.
```

Fuco1:
I did the rebase on upstream master and added a caveat section to usage in the documentation.

psibi:
Thanks!

PR #167, opened 2017-04-14
Cycle workspace by screen

IvanMalison:
Cycle through previously viewed workspaces in the order they were viewed most
recently on the screen where cycling is taking place.

The extremely long name is needed to differentiate this module from `XMonad.Actions.CycleRecentWS` which does something quite similar, with the main difference being that the history for different screens is not indendent (though it has the advantage of not depending on workspacse history)

I also provided the repeatableAction and update WorkspaceHistory as part of this change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,6 +35,15 @@
     called for activated windows.  But this lifts `manageHook` into
     `FocusHook` and it needs to be converted back later using `manageFocus`.
 
+  * `XMonad.Actions.CycleWorkspaceByScreen`
+
+    A new module that allows cycling through previously viewed workspaces in the
+    order they were viewed most recently on the screen where cycling is taking
+    place.
+
+    Also provides the `repeatableAction` helper function which can be used to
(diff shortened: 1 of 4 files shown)


pjones:
@IvanMalison Two things:

1. Please squash your commits. See [this file](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md) for more details.

2. What's going on with all the changes in the `.cabal` file?

IvanMalison:
> Please squash your commits. See this file for more details.

Sure. Do you want me to put everything from History and CycleRecent together or can I keep those separate.

> What's going on with all the changes in the .cabal file?

I just ran M-x sort lines on all the modules since they are clearly *SUPPOSED* to be alphabetically sorted. It seems like a few mistakes were made here and there.

pjones:
Go ahead and squash everything into a single commit.  And thanks for the note about sorting.  That makes sense.

IvanMalison:
@pjones Merge?

pjones:
Thanks!

PR #169, opened 2017-04-20
Use minimize in scratchpads

f1u77y:
### Description

Use X.A.Minimize instead of NSP workspace for managing scratchpads. Rationale is described in #90.

### Checklist

### DIFF
--- a/XMonad/Util/NamedScratchpad.hs
+++ b/XMonad/Util/NamedScratchpad.hs
@@ -25,16 +25,12 @@ module XMonad.Util.NamedScratchpad (
   namedScratchpadAction,
   allNamedScratchpadAction,
   namedScratchpadManageHook,
-  namedScratchpadFilterOutWorkspace,
-  namedScratchpadFilterOutWorkspacePP
   ) where
 
 import XMonad
 import XMonad.Hooks.ManageHelpers (doRectFloat, isInProperty)
 import XMonad.Actions.Minimize
 import XMonad.Actions.CopyWindow
-import XMonad.Actions.DynamicWorkspaces (addHiddenWorkspace)
-import XMonad.Hooks.DynamicLog (PP, ppSort)
 
 import Control.Monad (filterM)
 import Data.Maybe (listToMaybe)
@@ -147,7 +143,7 @@ someNamedScratchpadAction f confs n
   where
     toggleMinimize :: Window -> X ()
     toggleMinimize window = runQuery isMinimized window >>= \case
-        True -> maximizeWindow window
+        True -> maximizeWindow window >> windows (W.focusWindow window)
         False -> minimizeWindow window
 
     isMinimized :: Query Bool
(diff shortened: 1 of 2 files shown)


pjones:
I like this idea a lot.  

When you get around to writing the entries in `CHANGES.md` be sure to describe the steps necessary to migrate any configuration that might stop compiling.

f1u77y:
I have two questions about some non-trivial changes in this area:
1. Some modules do still have and export their functions for filtering out workspaces. I think that practically it's required only for "NSP" workspace but I of course can be wrong, so if someone has idea about what to do with these functions, leave feedback please. 
2. (Maybe it's a bit offtopic) Activating and then deactivating NamedScratchpad may lead to enexpected window activation(we may activate not the same window we came from, which is expected by user). That should be in the old version too.

pjones:
I have a few named scratch pads and when I deactivate one, the window that was focused before activating the scratch pad will be focused again.  This is pretty important for me.  

For example, one of my scratch pads is my password manager.  When I close the password manager scratch pad I want to be able to paste a password into the previously focused window without having to change the current focus to the correct window.

skewerr:
Could this mess with X.A.Minimize's per-workspace stacks? I'm a bit tired right now to test it with multiple workspaces and minimized windows(the way I've been using them for a while), but I'll leave this here and come back if/when I eventually do.

My question is: if I'm in workspace \#1, minimize a window A with a keybind to `withFocused minimizeWindow`, use a keybind to invoke a named scratchpad that spawns window B into the stack, then I send it away from the same workspace \#1, when I use a keybind to `withLastMinimized maximizeWindow`, will I see the scratchpad window B or window A? What about other workspaces, in case I send the scratchpad away from workspace \#1 and invoke `withLastMinimized maximizeWindow` from workspace \#2?

b0o:
@f1u77y Thank you for this patch, I was having tons of issues using `marshallPP` from `X.L.IndependentScreens` and your patch fixed my issue - turned out `IndependentScreens` doesn't know how to handle the `NSP` workspace since it expects all workspaces to follow its naming convention.

NickHu:
I would like to get this merged too, since the `NSP` workspace always seemed like a hack anyway

eprst:
LGTM, but how does it play with the task bars? Say if I have 10 hidden scratchpads, will I see all 10 in the taskbar panel, e.g. if using xmonad with xfce?
We could probably add `_NET_WM_STATE_SKIP_TASKBAR` to `_NET_WM_STATE`, but it is supposed to be managed by the app, not WM.

geekosaur:
Supposed to be, but that didn't stop me from writing http://hackage.haskell.org/package/xmonad-contrib-0.15/docs/XMonad-Util-NoTaskbar.html.

byorgey:
Reviewing the discussion here, it looks like one potentially serious objection that hasn't been addressed (if I'm understanding correctly) is that currently, when you pull up a named scratchpad and then banish it again, focus returns to the most recently focused window.  But with this change the focus will go someplace random instead.  Is that correct?  I think I would be annoyed by the change.

ivanbrennan:
> ...currently, when you pull up a named scratchpad and then banish it again, focus returns to the most recently focused window.
@byorgey I'm not experiencing that with the current implementation. I _want_ it to behave that way, but at the moment, banishing the scratchpad causes the lowest window in the stack to be focused, rather than the most recently focused window.

byorgey:
@ivanbrennan strange.  I just tested this again with my setup to be sure, and I am indeed seeing the behavior I described --- after banishing a named scratchpad the focus returns to whichever window was focused previously.  The behavior you describe definitely seems wrong. Could you open a new ticket with a minimal config that shows that behavior?

ivanbrennan:
@byorgey It seems to be the use of [scratchpads in combination with sublayouts](https://github.com/xmonad/xmonad-contrib/pull/376#issuecomment-687109644) that causes the behavior I'm seeing. Here's a minimal config I used to repro:
```haskell
import XMonad
import XMonad.Layout.SubLayouts (subLayout)
import qualified XMonad.Util.NamedScratchpad as NS

keys' = const . M.fromList $
    ((mod4Mask, xK_Shift_R), NS.namedScratchpadAction scratchpads "scratchpad")
  ]

scratchpads = [NS.NS name command (appName =? name) NS.defaultFloating]
  where
    name = "scratchpad"
    command = terminal def ++ " -name " ++ name

layoutHook' = subLayout [] tiled tiled
  where
    tiled = Tall 1 (3/100) (1/2)

main = xmonad def
  { keys = keys',
    manageHook = NS.namedScratchpadManageHook scratchpads,
    layoutHook = layoutHook'
  }
```

slotThe:
I will close this, since it'll probably need to be rewritten when we tackle (parts of) #591 anyways

PR #170, opened 2017-04-24
Only modify the last word during tab completion

### LINKED ISSUE #164: X.Prompt: Tab completing a second word erases the first
### Problem Description

I inadvertently introduced a bug in commit 57c00b1086359b0061ae5ce33aafc8b615e43ddb while attempting to improve the feedback of tab completion.

The bug: If `alwaysHighlight` is enabled and the user presses the completion key, Prompt will attempt to expand the completion and highlight the matching menu item.  If you then enter a space, type some more characters, then press the completion key again, the entire input line will be replaced with the completion of the second word.  Instead, the first word should remain unaffected while the second and subsequent words are completed successfully. 

Easy way to reproduce:

  * Configure the shell prompt to use `alwaysHighlight`

  * Bring up the shell prompt

  * Tab complete a shell command name that expects an argument

  * Enter a space, start typing the first argument (probably a file name), and hit tab

  * Notice that the command name was removed from the input buffer and replaced with the expanded argument.

mavant:
Fixes #164.

It's not clear to me why these functions take in XPState as a param only
to then produce some XP () values. It seems like they could just as well
call `get`.

### Description

When tab-completing in alwaysHighlight mode, rather than setting the entire command string, just delete the last word and insert the completion in its place.
Since the entire command is no longer set simply to a completion, dispatch on the last word of the command instead.
Motivated by https://github.com/xmonad/xmonad-contrib/issues/164.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -337,6 +337,11 @@
 
     - Added `filterOutWs` for workspace filtering.
 
+  * `XMonad.Prompt
+
+    - Accommodate completion of multiple words even when `alwaysHighlight` is
+      enabled.
+
 ## 0.16
 
 ### Breaking Changes
--- a/XMonad/Prompt.hs
+++ b/XMonad/Prompt.hs
@@ -698,9 +698,8 @@ handleCompletion cs = do
 
     let updateWins  l = redrawWindows l
         updateState l = case alwaysHlight of
-            False                                           -> simpleComplete l st
-            True | Just (command st) /= highlightedCompl st -> alwaysHighlightCurrent st
-                 | otherwise                                -> alwaysHighlightNext l st
+            False -> simpleComplete l st
+            True  -> hlComplete l st
 
     case cs of
       []  -> updateWindows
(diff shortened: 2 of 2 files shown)


pjones:
Thanks.  This is on my list of things to pull into my config and test.

mavant:
Figuring out how to test on Nix was definitely educational. Kinda makes me miss Arch.

pjones:
@mavant The current merge conflicts are probably being caused commit d873053 where master was merged into this feature branch.  Typically, merging master into a feature branch is frowned upon because it makes it really hard to see what changes are being made by this branch vs. changes that came from previous commits on master.

If you could drop commit d873053 and rebase instead that would be wonderful.  I have more details in the xmonad CONTRIBUTING.md file if you are interested.  Thanks.

mavant:
Sure thing. Sorry, this was the first time I'd used the web UI to resolve conflicts, and it didn't occur to me that doing so would explicitly create a merge commit. Pushing a rebase in a sec.

slotThe:
#164 is still an issue and this fix still works; no idea why it was never merged.  @mavant the only thing that's left to be done is a rebase onto current `master`.  If you're not interested in this anymore (which I can definitely understand, after being ignored for three years :>) just say so and I'll do this myself

mavant:
@slotThe For the next few days my only internet access is via cell phone, so please feel free to rebase this.

slotThe (approved):

slotThe:
Thanks!

PR #178, opened 2017-05-05
add new module X.U.ExclusiveScratchpads

bforte:
### Description

This is a new module that extends the idea of NamedScratchpads: For example, a lot of the time I just need a scratchpad for doing a quick calculation (eg. a `ghci` shell) after that I might use another scratchpad (eg. `octave` shell), with this module the workspace doesn't get polluted since I can group related scratchpads that share the same position on the workspace.

I use it a lot and thought it might be worth being shared, note that I'm not sure if this meets all your standards as this is my first (real) contribution. I tested the module for about a week now with my current setup and it works the way it is supposed to, I also I tested it with `xmonad-testing`, so I think it should do fine.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,13 @@
 
 ### New Modules
 
+  * `XMonad.Util.MutexScratchpads`
+
+    Named scratchpads that can be mutually exclusive: This new module extends the
+    idea of named scratchpads such that you can define "families of scratchpads"
+    that are mutually exclusive on the same screen. It also allows to remove this
+    constraint of being mutually exclusive with another scratchpad when.
+
   * `XMonad.Hooks.Focus`
 
     A new module extending ManageHook EDSL to work on focused windows and
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,11 +34,11 @@
 
 ### New Modules
 
-  * `XMonad.Util.MutexScratchpads`
+  * `XMonad.Util.ExclusiveScratchpads`
 
     Named scratchpads that can be mutually exclusive: This new module extends the
     idea of named scratchpads such that you can define "families of scratchpads"
-    that are mutually exclusive on the same screen. It also allows to remove this
+    that are exclusive on the same screen. It also allows to remove this
     constraint of being mutually exclusive with another scratchpad when.
 
   * `XMonad.Hooks.Focus`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,7 +39,7 @@
     Named scratchpads that can be mutually exclusive: This new module extends the
     idea of named scratchpads such that you can define "families of scratchpads"
     that are exclusive on the same screen. It also allows to remove this
-    constraint of being mutually exclusive with another scratchpad when.
+    constraint of being mutually exclusive with another scratchpad.
 
   * `XMonad.Hooks.Focus`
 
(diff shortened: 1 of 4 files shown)


crocket:
Why is it called mutex?

bforte:
Because the point is that they are mutually exclusive. First I named it competing scratchpads but I thought that's a stupid name. I'm not good with naming, haven't thought about ExclusiveScratchpads might be a better name.. Should I rename it?

crocket:
[Mutex](https://en.wikipedia.org/wiki/Mutual_exclusion) is the name of a concurrency control strategy.
It confuses people who knows this. Renaming is recommended.

bforte:
Yes, I knew that - this was the reason I named it like this because it would be clear what it does from the name, seems like it missed the point. I'll rename it to avoid further confusion, sorry about that!

crocket:
If you use `nsp` instead of `"NSP"` in your module, I recommend that you export `nsp`, too.
I'm paranoid about breaking changes. If `nsp` is not exported and the value of `nsp` changes in the future, XMonad configuration might break.

bforte:
There's no reason to change `nsp` in the future, I just thought it might be a good thing to do in case someone wants to change it, makes it easier to do that and also when you're coming from NamedScratchpads.

The NSP workspace is going to disappear: I saw [#169](https://github.com/xmonad/xmonad-contrib/pull/169) and I'm going to rewrite it using X.A.Minimize, I did not know about that before I started writing this. As soon as I get some time, I'll rewrite the module. So we can put this on hold I guess..

pjones:
@bforte Thanks for the new module.  Let us know when the new version is ready and we'll get it merged.

bforte:
@pjones Should I open another PR?

NickHu:
I've been using this for a while now, and I noticed that when you call the `scratchpadAction`, and it spawns a new window, while the last window is focused, the order of all the windows on the workspace seems to shift around a bit. Can anyone else confirm?

geekosaur:
Welcome to the `StackSet`. You will indeed see this, and you will find there's no reliable workarounds, much less fixes.

NickHu:
I've been using this module now for about 2 years and it's been stable; is there any chance of it getting merged?

byorgey:
@NickHu thanks for the nudge.  This looks good to me.

crazyboycjr:
Hi,

Thank you guys, I love this module very much. Here's how I typically use it. I used to prepare three transparent and fullscreen terminals (I use Alacritty), and use F1, F3, F4 to switch among them. I have been worked like this with Tilda since 2016. However, I find that when I'm focusing on one scratchpad and try to switch to another (e.g. by pressing F3), the new one fails to get the focus. This is a little bit annoying because I have to minimize and maximize it manually to regain the focus (e.g. by pressing F3 twice).

It took me a little while to dive into the source code to find the reason. It turns out that there are two [lines](https://github.com/xmonad/xmonad-contrib/blob/252e6e4d0074dcd0eda531177ac2e17939a9d3d7/XMonad/Util/ExclusiveScratchpads.hs#L150) of code related to this behavior. I think it should be `hideOthers` first, then `toggleWindow w`. Would you @bforte @NickHu mind taking a look at this and updating the code if necessary? I appreciate it very much!

Thank you!

byorgey:
Hi @crazyboycjr , thanks for looking into this! I think the best way forward would be for you to open a pull request with the changes that you think should be made (if you need help figuring out how to do that just let me know, happy to help).  Then we can use the pull request as a place to discuss the changes.

PR #185, opened 2017-05-23
Improve conversion of Char to KeySym in XMonad.Util.Paste

aplaice:
### Description

The issue with the current implementation is that in order to paste a character, it's trying to convert a `Char` into a `KeySym` using `stringToKeysym`, which is not really designed for that, as it's supposed to take the name of the character ("colon" or "agrave") — a string. "Luckily", for ascii alphanumeric characters ("a", "1") the name of the character co-incides with the character itself, so the Char is converted into a String and fed to `stringToKeysym`. Unfortunately, this doesn't work even for ascii punctuation, so the output from `stringToKeysym` is fed as default value to `parseKey`. `parseKey` is slightly better, but it does not fare well with non-ascii characters or even with a space " ".

The method suggested in the patch exploits the fact that for Latin-1 characters (Unicode code point below 256 (0x100)), the keysym of the character is the same as its code point, while for all other characters, 0x1000000 plus the code point, corresponds to a matching keysym. (The latter statement is a slight simplification — see [here](http://www.cl.cam.ac.uk/~mgk25/ucs/keysyms.txt) for details.) I think that the `maybe` and the `parseKey` part of `pasteChar` is now no longer necessary, but it does no harm, so I've left it in.

One remaining, huge, problem with XMonad.Util.Paste is that `keysymToKeycode` (used in the basic `sendKeyWindow`) only works if the given key is in your current keymap (which sort-of makes sense, but tools like `xdotool` manage to circumvent this) and hence you still cannot send arbitrary unicode characters. However, trying to remedy this would be far more complicated.

### Checklist

I'm not sure what to write in `CHANGES.md`.

### DIFF
--- a/XMonad/Util/Paste.hs
+++ b/XMonad/Util/Paste.hs
@@ -68,13 +68,14 @@ pasteString = mapM_ (\x -> if isUpper x || x `elem` "~!@#$%^&*()_+{}|:\"<>?" the
 
    > pasteChar shiftMask 'F'
 
-   Note that this function makes use of 'stringToKeysym', and so will probably
-   have trouble with any 'Char' outside ASCII.
+   Note that this function will probably have trouble with any 'Char'
+   outside ASCII.
 -}
 pasteChar :: KeyMask -> Char -> X ()
-pasteChar m c = sendKey m $ maybe (stringToKeysym [c]) fst
+pasteChar m c = sendKey m $ maybe (unicodeToKeysym c) fst
                 $ listToMaybe $ readP_to_S parseKey [c]
 
+-- | Send a key with a modifier to the currently focused window.
 sendKey :: KeyMask -> KeySym -> X ()
 sendKey = (withFocused .) . sendKeyWindow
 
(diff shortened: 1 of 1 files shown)


geekosaur:
Given that the X event uses a keycode, not a keysym, the only way `xdotool` can circumvent this is to find an unused keycode and assign it to the keysym it wants to paste, then undo that assignment sometime afterward. The "sometime" is a problem, since if it does it too soon, the receiving program won't be able to use the mapping to convert the keycode back to a keysym. And neither `xdotool` nor any other program can tell when the receiving program has processed the event and it's safe to remove the mapping.

aplaice:
That is probably more complicated than would be worth implementing in XMonad, for relatively little gain.

slotThe (approved):
LGTM, I would agree that trying to do any kind of circumvention here sounds like much more trouble than it's worth.  @geekosaur what do you think?

geekosaur:
I already commented on it above and see no reason to change my assessment.

geekosaur:
To be more clear, I agree that above circumvention is taking things a bit too far, especially when one can use `xdotool` if they really want it.

aplaice:
Thanks for merging!

PR #186, opened 2017-05-24
Fix render order of LayoutHints and MultiColumns

ankaan:
### Description

Before this fix, when using layoutHintsToCenter together with
MultiColumns, in certain situations XMonad would render the border of
the focused window below a border of unfocused windows. This looks odd
and is here fixed by changing MultiColumns to always place the focused
window in front (even though they should not really overlap) and making
LayoutHints preserve the order returned from the underlying layout,
except for the focused window that is placed on top.

This is a good idea since layoutHintsToCenter requires the focused
window to be on top for good rendering, even if that is not really
required when the underlying layout is used on its own. This way
layoutHintsToCenter requires less of the layout that is modified and
MultiColumns is more compatible with future layout modifiers that are
not so considerate.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,19 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Layout.LayoutHints`
+
+    Preserve the window order of the modified layout, except for the focused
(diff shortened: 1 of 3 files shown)


pjones:
Thanks!

PR #190, opened 2017-06-06
Replace X with MonadIO in XMonad.Util.Dmenu

crocket:
### Description

MonadIO is compatible with xfork which prevents dmenu prompts from freezing XMonad.
Without xfork, if I try to give focus to another window while dmenu is waiting, XMonad freezes until I kill dmenu in virtual terminal or ssh session.

### DIFF
--- a/XMonad/Util/Dmenu.hs
+++ b/XMonad/Util/Dmenu.hs
@@ -41,28 +41,32 @@ import XMonad.Util.Run
 -- <http://www.jcreigh.com/dmenu/dmenu-3.2-xinerama.patch>
 dmenuXinerama :: [String] -> X String
 dmenuXinerama opts = do
-    curscreen <- (fromIntegral . W.screen . W.current) `fmap` gets windowset :: X Int
-    runProcessWithInput "dmenu" ["-xs", show (curscreen+1)] (unlines opts)
+    curscreen <-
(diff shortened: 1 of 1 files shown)


crocket:
This closes https://github.com/xmonad/xmonad-contrib/issues/188

XMonad/Util/Dmenu.hs
```diff
@@ -41,34 +41,38 @@ import XMonad.Util.Run
...
 -- | like 'dmenu' but also takes the command to run.
-menu :: String -> [String] -> X String
+menu :: MonadIO m => String -> [String] -> m String
 menu menuCmd opts = menuArgs menuCmd [] opts
 
 -- | Like 'menu' but also takes a list of command line arguments.
-menuArgs :: String -> [String] -> [String] -> X String
-menuArgs menuCmd args opts = fmap (filter (/='\n')) $ runProcessWithInput menuCmd args (unlines opts)
+menuArgs :: MonadIO m => String -> [String] -> [String] -> m String
+menuArgs menuCmd args opts = fmap (filter (/='\n')) $
```
byorgey: This breaks compilation on GHC < 7.10 since it uses `fmap` but only has a `MonadIO` constraint (`Functor` was not a superclass of `Monad` until GHC 7.10).  The simplest fix is to just use `liftM` instead of `fmap`.

byorgey:
Looks great, thanks!

PR #193, opened 2017-06-10
X.L.LayoutHints stop windows from overlapping

ankaan:
### Description

This is an improvement on the pull request "Fix render order of
LayoutHints and MultiColumns" (#186) and addresses the actual underlying
problem.

It turned out that windows can sometimes overlap also. This happens when
a window is exactly in the center along an axis. There is a special
case in the code for this, but it was not handled properly.

I removed the code that places the focused window on top since it is no
longer required, but I still preserve the window order of the underlying
layout. This interferes even less with the underlying layout.

I also removed some code paths that were no longer necessary due to this
change and generalized some types so that I could debug the code more
easily.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -60,11 +60,10 @@
 
   * `XMonad.Layout.LayoutHints`
 
-    Preserve the window order of the modified layout, except for the focused
-    window that is placed on top. This fixes an issue where the border of the
-    focused window in certain situations could be rendered below borders of
-    unfocused windows. It also has a lower risk of interfering with the
-    modified layout.
+    - Preserve the window order of the modified layout, this lowers the risk of
+      interfering with the modified layout.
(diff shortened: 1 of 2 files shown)


ankaan:
I realized that the special case that I initially removed actually made things more visually pleasing, so I added it back but made it work properly.

slotThe (approved):
LGTM, but I don't really use layout hints, so I can't adequately review it in detail.

@liskin you were flagged by that bot in the first run; do you still use this functionality and can review the change?

@ankaan can you perhaps rebase on top of current `master`?

liskin:
I do use LayoutHints but I don't use LayoutHintsToCenter so I'm probably unaffected, and it's been almost 10 years since I last touched this module, so I'm as qualified to review it as anyone else. Might be nice to add unit test for the problem being fixed, or something like that which will help potential reviewers to understand it.

PR #196, opened 2017-06-25
Add XMonad.Util.SessionStart

Ongy:
### Description

I added a `XMonad.Util.SessoinStart` module to include a way to check if we are currently starting the session, or restarting xmonad in the startupHook.
This is a more general approach than `spawnOnce` which e.g. allows to use `spawnOn`.

Currently this requires to manually set a flag in the ExtensibleState (at the end of startupHook). If this is moved to the core repository at some point, this can be done automatically when the state file is read.

### Checklist

  - [✓] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [✓] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

  - [✓] I updated the `CHANGES.md` file

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,6 +56,14 @@
     Also provides the `repeatableAction` helper function which can be used to
     build actions that can be repeated while a modifier key is held down.
 
+  * `XMonad.Utils.SessionStart`
+
+    A new module that allows to query if this is the first time xmonad is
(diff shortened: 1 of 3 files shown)


byorgey:
Thanks!

PR #198, opened 2017-06-28
In XMonad.Prompt.Unicode, replace unsafePerformIO w/ ExtensibleState

mekeor:
Using unsafePerformIO is a sin in Haskell. In this module,
XMonad.Prompt.Unicode was used as a dirty hack for caching the contents of
the UnicodeData file.

This patch removes the usage of unsafePerformIO within the module and instead
makes usage of XMonad.Util.ExtensibleState to save the contents of the
UnicodeData file.

Furthermore, it allows to specify a custom path to that UnicodeData
file (unicodePromptFromFile). It does not rename old identifiers though.

### Description

behind them.

### Checklist

### DIFF
--- a/XMonad/Prompt/Unicode.hs
+++ b/XMonad/Prompt/Unicode.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+
 {- |
 Module      :  XMonad.Prompt.Unicode
 Copyright   :  (c) 2016 Joachim Breitner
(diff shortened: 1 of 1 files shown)


mekeor:
I moved some code snippets around because I wanted the code to go from general/abstract to specific/concrete.

NickHu:
@mekeor Wanted to bring to your attention #181 incase you hadn't seen it. It's a shame that we both decided to implement this on the same day...

Please correct me if I'm wrong, but I think your implementation is broken - there is no call to `XS.put` and hence it doesn't actually cache any unicode data, so it re-reads the file every time.

mekeor:
@NickHu Oh. I hadn't seen that you continued to work on your solution. So, anyway, you're right: My implementation is indeed broken. I forgot to do what I originally wanted to do. So, shall I just close this merge-request and we try to merge yours?

NickHu:
@mekeor I think that would be best

PR #206, opened 2017-07-22
Export XMonad.Prompt.insertString.

foreverbell:
### Description

Background: I am trying to add some shortcuts for XPrompt, for example,
when doing `Ctrl-V`, text in my clipboard will be pasted into prompt. The
best way to do this is to use the function `insertString`, but I surprisingly
find that it is not exported.

The current work around is to use `setInput`, but it loses track of the cursor position.
```haskell
myPromptKeymap = M.union defaultXPKeymap $ M.fromList
  [ ((controlMask, xK_v), setInput =<< liftIO getClipboard) 
  ]
```

As I have seen that `deleteString` is also exported, so I think there is no downside to
export this function. 

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -108,6 +108,10 @@
 
     Make type of ManageHook combinators more general.
 
+  * `XMonad.Prompt`
+
+    Export `insertString`.
+
   * `XMonad.Prompt.Window`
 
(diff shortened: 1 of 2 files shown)


byorgey:
Makes sense, thanks!

PR #212, opened 2017-09-26
Generalised hybridNavigation to hybridOf

LSLeary:
### Description

Previously in #77 I wrote a very simple hybrid navigation strategy for Navigation2D, taking advantage of what was already written. It prioritised target windows found by lineNavigation over those found by centerNavigation, and due to relevant functions/constructors not being exposed by the module this wasn't able to be changed on the user end. This became a noticeable problem for me recently when I started using a layout with one window in a narrow vertical master pane and the rest of the screen in a grid, resulting in window configurations like the below:
configuration

In these cases lineNavigation—and hence hybridNavigation—move focus between the leftmost and rightmost windows, not able to reach the others. I don't use multiple screens so I can't confirm, but this may also be behind the comment @pjones left on a commit in the original PR. The behaviour should be expected so I don't consider it a bug as such, but changing the order of prioritisation to prefer centerNavigation would prevent the middle row getting skipped. Electing to generalise and allow either order (as well as hybrids with any future navigation strategies), I wrote and exposed hybridOf. It takes two exposed Navigation2D instances and returns a third that attempts the first, falling back on the second if it finds no target window.

hybridNavigation is hence redefined as `hybridOf lineNavigation centerNavigation` and deprecated.

### Checklist

    - [x] Actively tested against current git versions of xmonad & xmonad-contrib, just not within xmonad-testing. Hopefully that's OK, since it's such a minor change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -66,6 +66,10 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Actions.Navigation2D`
+
+    Generalised (and hence deprecated) hybridNavigation to hybridOf.
+
   * `XMonad.Layout.LayoutHints`
 
     Preserve the window order of the modified layout, except for the focused
--- a/XMonad/Actions/Navigation2D.hs
+++ b/XMonad/Actions/Navigation2D.hs
@@ -43,6 +43,7 @@ module XMonad.Actions.Navigation2D ( -- * Usage
                                    , Navigation2D
                                    , lineNavigation
                                    , centerNavigation
+                                   , hybridOf
                                    , hybridNavigation
                                    , fullScreenRect
                                    , singleWindowRect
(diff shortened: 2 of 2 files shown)


LSLeary:
It seems the Travis builds failed because X11-1.8 failed to install. I don't imagine that has anything to do with the actual contents of this PR.

pjones:
This is great, thank you!

We need a volunteer to fix Travis then we can get this merged.

IvanMalison:
@pjones The quick fix for travis is the same as what I did in https://github.com/xmonad/xmonad/pull/137

pjones:
@IvanMalison thank you.  I updated the Travis config and restarted the build.

pjones:
@LSLeary Would you please pull with rebase from master to pick up the change to `.travis.yml`.  Thank you.

LSLeary:
@pjones done and passed.

PR #213, opened 2017-09-27
Added X.A.InputMode

lisael:
- it defines an action that switches to another key binding
      map.
    - implemented as recursive X.A.Submap

### Description

We often struggle to find binding-free keys. I, for myself, was reluctant to bind 16 keys just to move and resize floating windows ( yes 16, because I wanted 2 speeds).

`X.A.Submap`s may help sometimes but in that case it was just an no-go, so I hacked a recursive Submap, like so:

```haskell
resizeMode =
    [ ((0, xK_h), do withFocused (keysResizeWindow (-15,0) (0,0))
                     submap . M.fromList $ resizeMode)
    , ((0, xK_l), do withFocused (keysResizeWindow (15,0) (0,0))
                     submap . M.fromList $ resizeMode)
    , ((0, xK_k), do withFocused (keysResizeWindow (0,-15) (0,0))
                     submap . M.fromList $ resizeMode)
    , ((0, xK_j), do withFocused (keysResizeWindow (0,15) (0,0))
                     submap . M.fromList $ resizeMode)
    , ((controlMask, xK_h), do withFocused (keysResizeWindow (-1,0) (0,0))
                               submap . M.fromList $ resizeMode)
    , ((controlMask, xK_l), do withFocused (keysResizeWindow (1,0) (0,1))
                               submap . M.fromList $ resizeMode)
    , ((controlMask, xK_k), do withFocused (keysResizeWindow (0,-1) (0,0))
                               submap . M.fromList $ resizeMode)
    , ((controlMask, xK_j), do withFocused (keysResizeWindow (0,1) (1,0))
                               submap . M.fromList $ resizeMode)
    , ((0,         xK_Tab), do focusDown
                               submap . M.fromList $ resizeMode)
    , ((0,      xK_Escape), return ())
    ]
```

I eventually made a function generating this mess from a simple keybinding list, for usability, and heres my package proposal.

Note that it's my second real haskell program —the first one being my `xmonad.hs` :) —. I hope it's good enough.

### Checklist

### DIFF
--- a/XMonad/Actions/InputMode.hs
+++ b/XMonad/Actions/InputMode.hs
@@ -1,0 +1,65 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Actions.InputMode
+-- Copyright   :  (c) Bruno Dupuis <lisael@lisael.org>
+-- License     :  BSD3-style (see LICENSE)
+--
+-- Maintainer  :  Bruno Dupuis <lisael@lisael.org>
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- A module that allows the user to create a so called imput mode, that is  a
+-- persistent "XMonad.Actions.Submap".
+--
+-----------------------------------------------------------------------------
+
+module XMonad.Actions.InputMode (
+                             -- * Usage
+                             -- $usage
+                             inputMode
+                            ) where
+import XMonad hiding (keys)
+import qualified Data.Map as M
+import XMonad.Actions.Submap
+
+{- $usage
+
+
+
+
+First, import this module into your @~\/.xmonad\/xmonad.hs@:
+
+> import XMonad.Actions.InputMode
+
+Create a sub-mapping of keys. Example using "XMonad.Actions.FloatKeys":
+
+> moveMode =
+>     [ ((0, xK_h), withFocused (keysMoveWindow (-15,0) ) )
+>     , ((0, xK_l), withFocused (keysMoveWindow (15,0) ) )
+>     , ((0, xK_k), withFocused (keysMoveWindow (0,-15) ) )
+>     , ((0, xK_j), withFocused (keysMoveWindow (0,15) ) )
+>     , ((0,         xK_Tab), focusDown )
+>     ]
+
+Add a switch in your key bindings:
+
+>    , ((modm,                 xK_g), inputMode moveMode)
+
+So, you cat hit `modm+g` to activate the move mode and use
+h, j, k, l to move a window around. Hit 'Escape' to restore your normal
+bindings.
+
+For detailed instructions on editing your key bindings, see
+"XMonad.Doc.Extending#Editing_key_bindings".
+
+-}
+
+-- | Given a list of key bindings, return an action that temporay modifies
+--   your bindings. Hit `Escape` to switch back to normal key bindings.
+inputMode :: [( (KeyMask, KeySym), (X () ) )] -> X ()
+inputMode xs = do
+    submap . M.fromList $ modeMap where
+        modeMap = [ ( binding , do x
+                                   (submap . M.fromList $ modeMap) )
+                    | ( binding , x ) <- xs
+                  ] ++ [((0, xK_Escape), return ())]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,6 +63,12 @@
 
     Currently needs manual setting of the session start flag. This could be
     automated when this moves to the core repository.
+    
+ * `XMonad.Actions.InputMode`
+
+    A new Module that allows to declare a temporary keybindings Map. It's
+    built on top of `XMonad.Actions.Submap` by creating a recursive
+    submap that re-activate itself after each action.
 
 ### Bug Fixes and Minor Changes
 


XMonad/Actions/InputMode.hs
```diff
@@ -0,0 +1,65 @@
...
+"XMonad.Doc.Extending#Editing_key_bindings".
+
+-}
+
+-- | Given a list of key bindings, return an action that temporay modifies
+--   your bindings. Hit `Escape` to switch back to normal key bindings.
+inputMode :: [( (KeyMask, KeySym), (X () ) )] -> X ()
+inputMode xs = do
+    submap . M.fromList $ modeMap where
+        modeMap = [ ( ( km, ks ), do x
```
lisael: no need to pattern-match `( km, ks )`, it will be used as is. I correct this.

lisael:
I've got a few concerns about the module, now:

- how does this mix with `EZConfig` keybindings (I don't use it, but it seems to be pretty popular)
- maybe the function does a bit to much things by calling `submap . M.fromList` by itself. It could be a pure `[ ( (KeyMask, KeySym), (X () ) ) ] -> [ ( (KeyMask, KeySym), (X () ) ) ]` function and let the user create the submap. This should also solve my concerns about EZConfig.

@byorgey,  any thoughts?

LSLeary:
For what it's worth, I'm testing this (since it happens to be rather useful). It has minimal imports, so rather than pulling the commit to my contrib repo I just rewrote the function in my config. Only the aesthetics are different but this is my version; use it if you like it better.
```haskell
inputMode :: [((KeyMask, KeySym), X ())] -> X ()
inputMode bindings = submap modeMap
  where modeMap = M.fromList
                $ ((0, xK_Escape), return ())
                : [ (maskedKey, action  >> submap modeMap)
                  | (maskedKey, action) <- bindings
                  ]
```
I don't think there's any real point in splitting out the pure logic in this case.

Re EZConfig, you certainly can't pass keybindings to `inputMode` in the form you would to `additionalKeysP`. If the module contained and exposed a `fromP :: String -> Maybe (KeyMask, KeySym)` function that did the parsing it would be easy to write `inputModeP`, but I couldn't find anything that tidy. In any case it's an extraneous format that there's no need to support; `[((KeyMask, KeySym), X ())]`  is standard and sensible. And since inputMode xs is just an X (), it'll work correctly on being called from a keybinding set up through any method.

LSLeary:
OK, so after a while of use I've found that it is pretty sweet, but the fact that pressing any key not in the submap exits the mode is irritating in my usecase. I don't think `inputMode` should be _replaced_ with a version that does not have this behaviour—some people will definitely want it as is. Another version should be written and exposed alongside.

It happens to be quite easy to change; `submap` is an alias of `submapDefault (return ())`, where `return ()` is the default action to take on receiving a keypress not in the submap. By replacing `submap` with `submapDefault (inputMode bindings)` you get a version that stays in the mode given a keypress not in the map.

The comments/usage would probably need to be rewritten, but my current implementation looks like this:

```haskell
-- Not exposed.
type KeyBinding = ((KeyMask, KeySym), X ())

-- | Enters a mode in which given keybindings take effect; exit on unbound keys.
softInputMode :: [KeyBinding] -> X ()
softInputMode = inputMode (return ()) 0 xK_Escape

-- | Enters a mode in which /only/ given keybindings take effect; exit on escape.
hardInputMode :: [KeyBinding] -> X ()
hardInputMode bindings = self
  where self = inputMode self 0 xK_Escape bindings

-- | Enters a mode in which given keybindings take effect, and unmapped keys
--   perform def. Exit on (exitMask, exitKey).
inputMode :: X () -> KeyMask -> KeySym -> [KeyBinding] -> X ()
inputMode def exitMask exitKey bindings = self
  where self = submapDefault def . M.fromList
             $ ((exitMask, exitKey), return ()) : map2 id (>> self) bindings

-- Should be standard, right? Seems useful. Not exposed.
map2 :: (a -> c) -> (b -> d) -> [(a, b)] -> [(c, d)]
map2 f g l = zip (map (f . fst) l) (map (g . snd) l)
```

Re my contributing to this PR I don't know what the standard procedure is (I suppose I could submit a PR to your PR), but I honestly don't mind if you just copy & paste and/or rewrite the above.

lisael:
@LSLeary nice! I don't mind adding a "persistent" mode, I review your suggested code and test it when I find enough time (I still have to test the PR with xmonad-testing too)

LSLeary:
I found some weird behaviour you should look into it when you're testing. Everything else I've tried has worked 100%, but sending `kill` in input mode (either version) produces some unusual results. The immediate effect is that the remaining windows aren't retiled to fill the space immediately, and no window has focus. If you exit the mode it retiles and focuses. If you try to move to another window with Nav2D, that also exits the mode and focuses. I can do BinarySpacePartition operations like rotating and swapping, and well as visit other workspaces, but then when I exit the mode I find the windows that were in the workspaces I visited have vanished. ps shows their processes are still running.

TheMC47:
@LSLeary  @lisael  I found this PR yesterday and I think it's an awesome feature! Do you want to continue working on it? @LSLeary  could you provide a minimal configuration to the problem you found? Maybe I can help testing it

lisael:
@TheMC47 Sorry, I don't use Xmonad anymore, and I didn't write any Haskell since this PR, I'm afraid I can't help :/.

Feel free to fork my repo and iterate over this

slotThe:
@TheMC47 if you still want to pursue this, it may be better to clean up https://gist.github.com/LSLeary/6741b0572d62db3f0cea8e6618141b2f and use that instead

TheMC47:
Thank you @slotThe! I remember I spent a couple of hours playing around with that gist, but I was confused about grabbing/ungrabbing. If there are any docs explaining how that works in X, it would be great

geekosaur:
https://www.x.org/wiki/Development/Documentation/GrabProcessing/
https://tronche.com/gui/x/xlib/input/pointer-grabbing.html (also covers passive key grabs)

In short, a passive key or pointer grab tells the X server to watch for the specified key or pointer chord, and when it occurs establish an active keyboard or pointer grab. The active grab then redirects all key and/or pointer events to the grabbing client.

TheMC47:
Thanks! I'll check it

TheMC47:
I'm running into problems with using `GrabAll` from the gist: keys with no modifiers aren't grabbed. In the example below, pressing `M-F` works, but `D` doesn't. Replacing that with `GrabBound` makes both work

```haskell

main = xmonad …
  . modal passiveMode
  …
  def …


passiveMode :: XConfig Layout -> Mode
passiveMode cnf = Mode
  "DEFAULT"
  GrabBound
  (M.insert (modMask cnf, xK_n) (setTo (echoMode cnf) >> refresh) (keys cnf cnf))


echoMode :: XConfig Layout -> Mode
echoMode cnf = Mode "ECHO" GrabAll $ M.fromList
  [ ((noModMask, xK_d)  , xmessage "D")
  , ((modMask cnf, xK_d), setTo (passiveMode cnf) >> refresh)
  , ((modMask cnf, xK_f), xmessage "F")
  ]
```


Note: I don't particularly see the utility in `GrabAll`, and don't really know how to debug it more. Might just drop it in a first PR and then see what we can do

TheMC47:
Closed by https://github.com/xmonad/xmonad-contrib/commit/0891575518e2ee80d4551c9be3e606b79415c11e

PR #218, opened 2017-11-21
Extended the sendMessage interface of X.L.Gaps to allow arbitrary modifications

LSLeary:
### Description

I added `ModifyGaps (GapSpec -> GapSpec)` to the `GapMessage` type so that arbitrary modifications can be made to gaps at runtime. I utilised that to implement and expose:
  - `weakModifyGaps`, taking a simpler function of the form `Direction2D -> Int -> Int`
  - `modifyGap`, taking a function of the form `Int -> Int` and a `Direction2D`
  - `setGaps`, taking and instituting a `GapSpec`
  - `setGap`, taking an `Int` and a `Direction2D`

Ideally I would also like to remove the `IncGap` and `DecGap` constructors and re-implement them as top-level functions `incGap` and `decGap` using `ModifyGaps`, but that would break people's configs.

#### Motivation

Previously the closest thing to `sendMessage $ setGap n d` was `sendMessage (DecGap 9999 d) >> sendMessage (IncGap n d)`; an awful hack. Being able to set any or all gaps in a clean way is a basic functionality that should have been available from the start.

The full generalisation also allows behaviours like toggling gaps between two non-zero states, and shifting around or reshaping the tiled region of the screen whilst satisfying dynamic dimension constraints. I mainly use it to reposition windows if I want them to maintain a specific width—good for programming on a strict budget of 80 characters per line, or reading an ebook or some manga.

### Checklist

Edit: Just BTW, I had to remove the double quotes from the arithmetic evaluation expressions on lines 83 & 84 of the xmonad-testing/bin/run-in-xephyr.sh script to get it to work. Gave me some error about primitives.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -71,6 +71,11 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Layout.Gaps`
+
+    Extended the sendMessage interface with `ModifyGaps` to allow arbitrary
+    modifications to the `GapSpec`.
+
   * `XMonad.Actions.Navigation2D`
 
     Generalised (and hence deprecated) hybridNavigation to hybridOf.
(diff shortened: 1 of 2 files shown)


byorgey:
This looks great! Thanks!

PR #220, opened 2017-12-12
Added XMonad.Actions.SelectWindow

larkery:
### Description

A small action which lets you quickly select a window on the screen by pressing a key. I made a video to explain this:  https://gfycat.com/BaggyDishonestFoal

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,6 +69,11 @@
     Currently needs manual setting of the session start flag. This could be
     automated when this moves to the core repository.
 
+  * `XMonad.Actions.SelectWindow`
+
+    A new module that lets you quickly select any visible window by pressing
+    a key displayed in a hint on top of that window.
+
 ### Bug Fixes and Minor Changes
 
   * `XMonad.Layout.Gaps`
--- a/XMonad/Actions/SelectWindow.hs
+++ b/XMonad/Actions/SelectWindow.hs
@@ -1,0 +1,129 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Actions.SelectWindow
+-- Copyright   :  (c) 2017 Tom Hinton <xmonad@larkery.com>
+-- License     :  BSD3
+--
+-- Maintainer  :  Tom Hinton <xmonad@larkery.com>
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- Select windows using the keyboard quickly.
+-----------------------------------------------------------------------------
+
+module XMonad.Actions.SelectWindow (selectWindow,
+                                     selectWindowColors,
+                                     selectWindowKeys) where
+
+import Graphics.X11.Xlib.Extras (getWindowAttributes,
+                                 WindowAttributes (..))
+
+import XMonad.Util.Font
+import XMonad.Util.XUtils
+import Data.List (sortOn)
+import qualified Debug.Trace as D
+
+import XMonad
+import qualified XMonad.StackSet as W
+import Control.Arrow ((&&&))
+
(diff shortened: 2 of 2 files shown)


msk-:
This is embarrassing. I've just made a PR for a very similar feature, and thought I'd have a look around to see how long I should expect to wait for a response. And here this is! The main differences I can see between our PRs, by looking at your video:
 - I make a giant splash window that totally covers the window it's to select. This might be a matter of taste, but I can definitely see the appeal of your approach. I can't say I thought about this too much.
 - I take a single list of keys and generate a chord for each window; independent of the number of screens in use. Funny, I have a TODO in my code for mapping keys to screens.

If it's all the same to you, I'd be pretty happy to subsume the aforementioned functionality in my PR. I'd do the following:
 - Provide an option to control the splash size (perhaps just full-screen xor text-size)
 - Provide some ability to map from keys to physical screens
 - Anything I've missed?

In case it's important: I've been working on this on and (mostly) off since June, but just got some time to finish it off over the holidays.

larkery:
I don't really mind - if you make your thing meet my needs I would use it happily, as it saves me from doing maintenance and I'm not in it for the kudos :)

At the moment the main feature I depend on from mine is that the key chosen is deterministic for the physical locations of windows in front of me, so most of the time (when I have fewer than 8 windows visible) I don't need to read the hint display because I can predict which window will have which key. This makes it fast to use; if I always have to read the hints it becomes worse than just moving the cursor and so of no value. This is also why I don't need anything other than single keys.

I suppose the maintainers may also have an opinion when they get to review our commits; I am not sure what policy is used for accepting or rejecting pull requests.

msk-:
Cool. My use-case was largely selecting one from many logging terminals without using the cursor; often requiring multiple keys. I guess if one of us can sensibly produce the functionality of the other it makes sense to not have two somewhat overlapping modules.

I'll see whether I can produce the features I've listed above, as some of them are easy and others I'd meant to do anyway, and also duplicate your key/screen assignment mechanism, or create something similar. I quite like the idea of predictable keys for a given screen+window in certain circumstances.

Btw, when I pointed out I'd been working on this for a while, I was trying to make it clear I hadn't done this intentionally!

msk-:
Hello again,

Perhaps you might like to test my changes. They should meet your needs. Let me know if not. Use:

`(selectWindow def { sKeys=[[xK_a,xK_d,xK_f], [xK_j,xK_k,xK_l]], maxChordLen=1 }) >>= (flip whenJust (windows . W.focusWindow)`

Also, as some of this is your design, and a wee bit (the sorting stuff) is your code, I'd like to acknowledge this. Well, supposing you'd appreciate it! Would a "thanks to larkery/Tom Hinton (GH/other URL here)" suffice?

larkery:
Hi, I'll have a go on this at the weekend or maybe on an evening this week - I can't do too much config twiddling at work as it's not billable... I'm happy to be acknowledged, thanks!

msk-:
Cool, no hurry at my end, so whenever works for you is great.

larkery:
I had a go: it has some more options than mine so is likely useful to more people, i'll remove this request and use your module when it comes downstream in my package manager.

I think you're missing some exports from your code (e.g. for textSize and the like), and the compiler gives me some warnings about some fields not being initialized and some case branches being overlapping which might be worth checking up on.

msk-:
Hey, thanks for trying it out, and thanks for letting me know about those
issues. I'd sort of quit my life for a couple of months to ski, amongst
other things. I'll check it out and fix it up in a couple of weeks when I
get back. Thanks again!

PR #222, opened 2018-01-04
Actions.EasyMotion feature added

partiallyordered:
### Description

Added a new feature to select visible windows with key chords.

This is pretty much the first Haskell I've written, so any critique/suggestions/improvements will be very much welcome. Thanks!

### Checklist

  - [*] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)


msk-:
See #220, turns out @larkery and I have implemented similar features.

anthraxx:
Any news on this? This looks incredibly nice and it would be awesome to see it getting pulled

msk-:
I'm the worst. A few days ago I fixed up the compile warnings. I haven't had a chance to ensure I didn't break anything yet. Aiming to do that tomorrow. Sorry to anyone who's been following this :(.

slotThe:
This is incredible; @msk- are you still interesting in getting this merged?  If yes then I will try to find some time this as thoroughly as I can

msk-:
@slotThe yup, still interested, a review would be great please :+1:

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+                         where f' = min 0.0 $ max f 1.0
+
+-- | Handles overlay display and window selection. Called after config has been sanitised.
+handleSelectWindow :: EasyMotionConfig -> X (Maybe Window)
+handleSelectWindow c = do
+  f <- initXMF $ font c
+  th <- textExtentsXMF f (concatMap keysymToString (concat $ sKeys c)) >>= \(asc, dsc) -> return $ asc + dsc + 2
+  XConf { theRoot = rw, display = dpy } <- ask
+  XState { mapped = mappedWins, windowset = ws } <- get
+  let currentW = W.stack . W.workspace . W.current $ ws
```
slotThe, line 229: Because you're only using `currentW` to get to the current focus below, you can do 
```suggestion
  let currentW = W.peek $ ws
```

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+      -- Bucket windows by screen, unless the user has provided only a single list of keys
+      winBuckets :: X [[Overlay]]
+      winBuckets = sequence $ fmap (sequence . fmap buildOverlay) $
```
slotThe, line 232: `sequence . fmap` already has a name: `mapM`

The "newer" way is to use the slightly more general `traverse = sequenceA . fmap`.

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+      displayF = displayOverlay f (bgCol c) (borderCol c) (txtCol c) (borderPx c)
+      buildOverlay :: Window -> X Overlay
+      buildOverlay w = do
+        wAttrs <- io $ getWindowAttributes dpy w
+        let r = overlayF c th $ makeRect wAttrs
+        o <- createNewWindow r Nothing "" True
+        return Overlay { rect=r, overlay=o, win=w, attrs=wAttrs, chord=[] }
+  overlays <- fmap concat sortedOverlays
+  status <- io $ grabKeyboard dpy rw True grabModeAsync grabModeAsync currentTime
+  if (status == grabSuccess)
```
slotThe, line 250: Some of these would perhaps be more readable in a `where` clause (and with space separating individual functions)

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+    then do
+      resultWin <- handleKeyboard dpy displayF (cancelKey c) overlays []
+      io $ ungrabKeyboard dpy currentTime
+      mapM_ (deleteWindow . overlay) overlays
+      io $ sync dpy False
+      releaseXMF f
```
slotThe, line 256: This has the same problem that was fixed for X.Prompt and X.A.TreeSelect in #447, i.e. if I use the following config 

``` haskell
import XMonad
import XMonad.Actions.EasyMotion (selectWindow)
import XMonad.Util.EZConfig (additionalKeysP)

main :: IO ()
main = xmonad $ def{ clickJustFocuses = False }
  `additionalKeysP`
    [("M-n", selectWindow def >>= (`whenJust` windows . W.focusWindow))]
```

and  spawn two terminals, then press `M-n` and click into the non-focused window, xmonad will grab the keyboard and mouse and never release them.  This should be fixable with a similar setup as #447 in your `event` function

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+  case duplicatedKeys of
+    0 -> handleSelectWindow conf { sKeys = map sanitiseKeys (sKeys conf) }
+    _ -> return Nothing
+    where
+      keyList = concat (sKeys conf)
+      keyListLen = length keyList
+      uniqueKeyCount = length . toList . fromList $ keyList
+      duplicatedKeys = uniqueKeyCount - keyListLen
+      -- make sure the key lists don't contain: 'cancelKey, backspace
+      sanitiseKeys = toList . fromList . L.filter (not . flip elem [cancelKey conf, xK_BackSpace])
```
slotThe, line 276:
```suggestion
      sanitiseKeys = toList . fromList . L.filter (`notElem` [cancelKey conf, xK_BackSpace])
```

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+        Selected o -> return . Just $ win o
+        _ -> whenJust currentW (windows . W.focusWindow . W.focus) >> return Nothing -- return focus correctly
+    else releaseXMF f >> return Nothing
+
+-- | Display overlay windows and chords for window selection
+selectWindow :: EasyMotionConfig -> X (Maybe Window)
+selectWindow EMConf { sKeys = [] } = return Nothing
+selectWindow conf =
+  -- Make sure there are no duplicates in our key lists
+  case duplicatedKeys of
```
slotThe, line 267: I wonder why we don't just remove duplicates every time?  Might need to write that down in the docs, but right now I don't see a reason not to

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+      sanitiseKeys = toList . fromList . L.filter (not . flip elem [cancelKey conf, xK_BackSpace])
+
+-- | Take a list of overlays lacking chords, return a list of overlays with key chords
+appendChords :: Int -> [KeySym] -> [Overlay] -> [Overlay]
+appendChords _ [] _ = []
+appendChords maxLen ks overlays =
+  zipWith (\o c -> o { chord=c }) overlays chords
+    where
+      chords = replicateM chordLen ks
+      tempLen = -((-(length overlays)) `div` length ks)
```
slotThe, line 285: Can you describe, in words, what this number is supposed to represent?

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+
+-- | Create a rectangle from window attributes
+makeRect :: WindowAttributes -> Rectangle
+makeRect wa = Rectangle (fi (wa_x wa)) (fi (wa_y wa)) (fi (wa_width wa)) (fi (wa_height wa))
+
+-- | Display an overlay with the provided formatting
+displayOverlay :: XMonadFont -> String -> String -> String -> Int -> Overlay -> X ()
+displayOverlay f bgC brC textC brW Overlay { overlay = w, rect = r, chord = c } = do
+  showWindow w
+  paintAndWrite w f (fi (rect_width r)) (fi (rect_height r)) (fi brW) bgC brC textC bgC [AlignCenter] [L.foldl' (++) "" $ map keysymToString c]
```
slotThe, line 331:
```suggestion
  paintAndWrite w f (fi (rect_width r)) (fi (rect_height r)) (fi brW) bgC brC textC bgC [AlignCenter] [concatMap keysymToString c]
```

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+event :: Display -> IO (EventType, KeySym)
+event d = allocaXEvent $ \e -> do
+  maskEvent d (keyPressMask .|. keyReleaseMask) e
+  KeyEvent {ev_event_type = t, ev_keycode = c} <- getEvent e
+  s <- keycodeToKeysym d c 0
+  return (t, s)
+
+-- | A three-state result for handling user-initiated selection cancellation, successful selection,
+--   or backspace.
+data HandleResult a = Exit | Selected a | Backspace
```
slotThe, line 298: This is not exported and you're only using it in conjunction with `Overlay`; is there any reason why `a` is a type variable here?

slotThe (changes requested):
Awesome!

So the first round is complete (this will probably take multiple iterations, because the module itself is quite complex; sorry about that)

Some general things I've noticed: 

  1. Instead of merging `master` into your feature branch, you should rebase on top of `master` instead.  Otherwise the history will be incredibly hard to read.  See [contributing](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md#rebasing-and-squashing-commits) for more information
  2. You should also mention your module in `XMonad.Doc.Extending`, so users can find it

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+-- >    import XMonad.Actions.EasyMotion (selectWindow, EasyMotionConfig(..))
+-- >    , ((modm, xK_f), (selectWindow def {sKeys = [[xK_f, xK_d], [xK_j, xK_k]}) >>= (flip whenJust (windows . W.focusWindow)))
+--
+-- To customise font:
+--
+-- >    import XMonad.Actions.EasyMotion (selectWindow, EasyMotionConfig(..))
+-- >    , ((modm, xK_f), (selectWindow def {font = "xft: Sans-40"}) >>= (flip whenJust (windows . W.focusWindow)))
+--
+-- The font field provided is supplied directly to the initXMF function. The default is
+-- "xft:Sans-100". Some example options:
```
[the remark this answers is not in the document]
partiallyordered: Haven't tested this, but a scan of `initXMF` looks as though it'll fall back to `-misc-fixed-*-*-*-*-10-*-*-*-*-*-*-*`. Using core fonts is a pretty bad experience, in my testing. For whatever reason, core fonts cause the overlay to be very slow to appear. In any case, if someone still wants to use the module with core fonts, well, great. The fallback font is _tiny_ for the purpose of this module- a better default is in order at least. It looks as though I might be able to test whether XMonad was compiled with XFT using `#ifdef XFT` and use a conditional default. What do you think?

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+--
+-- >    "xft: Sans-40"
+-- >    "xft: Arial-100"
+-- >    "xft: Cambria-80"
+--
+-- Customise the overlay by supplying a function to do so. The signature is 'Position' ->
+-- 'Rectangle' -> 'X' 'Rectangle'. The parameters are the height in pixels of the selection chord
```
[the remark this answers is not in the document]
partiallyordered: Is this a convention, or does it have some bearing on the generated docs?

XMonad/Actions/EasyMotion.hs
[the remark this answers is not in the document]
slotThe: Yeah I think that's what we ended up deciding on in #349 as well; as much as I dislike CPP it's probably the correct choice in this situation

XMonad/Actions/EasyMotion.hs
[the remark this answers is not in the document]
slotThe: This is the haddock syntax for monospacing a piece of text (see [here](https://www.haskell.org/haddock/doc/html/ch03s08.html)); so it does have bearing on the generated docs. 

You can also put it on it's own line entirely like this: 

``` haskell
-- Customise the overlay by supplying a function to do so. The signature is @
-- 
-- > 'Position' -> 'Rectangle' -> 'X' 'Rectangle'
-- 
-- The parameters are the height in pixels of the selection chord
```

Whatever you prefer, really, but I do think we should follow the convention that code be monospaced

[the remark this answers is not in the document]
partiallyordered: Useful link, thank you

XMonad/Actions/EasyMotion.hs
partiallyordered: Latent shame and self-loathing. Hah. For the record, I appreciate how kindly you asked this, though I would've probably taken "this is not a good variable name" in good faith.

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,332 @@
...
+data EasyMotionConfig =
+  EMConf { txtCol      :: String                             -- ^ Color of the text displayed
+         , bgCol       :: String                             -- ^ Color of the window overlaid
+         , overlayF    :: Position -> Rectangle -> Rectangle -- ^ Function to generate overlay rectangle
+         , borderCol   :: String                             -- ^ Color of the overlay window borders
+         , sKeys       :: [[KeySym]]                         -- ^ Keys to use for window selection
+         , cancelKey   :: KeySym                             -- ^ Key to use to cancel selection
+         , font        :: String                             -- ^ Font for selection characters (passed to initXMF)
+         , borderPx    :: Int                                -- ^ Width of border in pixels
+         , maxChordLen :: Int                                -- ^ Maximum chord length. Use 0 for no maximum.
```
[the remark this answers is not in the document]
partiallyordered: Is `Maybe Int` appropriate? Feels sorta funny, not sure.

XMonad/Actions/EasyMotion.hs
partiallyordered: Sheesh, I've made.. some.. changes.. I don't have a good feel for Haskell aesthetics yet. E.g. `appendChords` is off by itself, it's a little larger than the others I've put in a `where` clause, but not by much. I'll resolve this particular item for now because the code has changed quite a bit since the above comment. More feedback in this vein very welcome, though.

partiallyordered:
@slotThe thanks a lot for the review, really. It's been a reintroduction to Haskell for me. But that's been good, if a little bit of a trial coming from my js day job. I've addressed most of the feedback, but I suspect you were right about another round- there have been a couple of significant changes and I'm expecting I have more to learn (about.. err.. perhaps monad transformers..?).

Regarding commit history, I'm not sure it'll be very useful, even if linear- unless you object I'll probably just squash it all into "added EasyMotion module" or similar.

Thanks again, it's been a good (re-)learning experience so far.

XMonad/Actions/EasyMotion.hs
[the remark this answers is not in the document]
slotThe: I would say that that's fine.  If this feels too contrived to you, you can probably just define some type that's isomorphic to `Maybe Int` yourself and give it a better name, like `data MaxLength = NoMaximum | Max Int` or something similar.

XMonad/Actions/EasyMotion.hs
slotThe: I think `appendChords` being a top-level function is not that bad; don't really want `handleSelectWindow` to become even bigger!  Btw, because I was just looking at this, you should be able to write 

``` Haskell
zipWith Overlay overlayWins chords
```

instead of having to explicitly deal with record syntax there.

Regarding the style I'll jot down a few things in the second review; particularly about those 150+ column lines ;)

slotThe:
> @slotThe thanks a lot for the review, really. 

My pleasure!

> I've addressed most of the feedback, but I suspect you were right about another round- there have been a couple of significant changes and I'm expecting I have more to learn (about.. err.. perhaps monad transformers..?).

Every time you've written a function within the `X` monad you were using monad transformers!  Easy peasy :)  I have some real-life stuff to take care of right now, so the review will probably be delayed by a few weeks; though I promise I'll get to it by early April!

> Regarding commit history, I'm not sure it'll be very useful, even if linear- unless you object I'll probably just squash it all into "added EasyMotion module" or similar.

Yeah I think this is the best approach;  it's nice (for me) to see the progression on what has changed, but the changes were so extensive that I'll probably end up going through the whole module again anyways.  And after merging this I think these would not be _that_ helpful for someone reading over the commit logs.

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,378 @@
...
+
+-- | Create overlay windows the full width of the window they select, the minimum height to contain
+--   their chord, and a proportion of the distance from the top of the window they select
+bar :: RealFrac f => f -> Position -> Rectangle -> Rectangle
+bar f th r = Rectangle { rect_width  = rect_width r
+                       , rect_height = fi th
+                       , rect_x      = rect_x r
+                       , rect_y      = rect_y r + round (f' * (fi (rect_height r) - fi th)) }
+                         -- clamp f in [0,1] as other values will appear to lock up xmonad
+                         -- as the overlay will be displayed off-screen
```
slotThe, line 243: I feel like this may be nicer as documentation for the definition of `f'`

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,378 @@
...
+                    sortedOverlayWindows :: X [OverlayWindow]
+                    sortedOverlayWindows = sortOverlayWindows <$> buildOverlayWindows dpy th visibleWindows
+                PerScreenKeys m -> fmap concat $ sequence $ M.foldr (:) [] $ M.mapWithKey (\sid ks -> buildOverlays ks <$> sortedOverlayWindows sid) m
+                  where
+                    screenById :: ScreenId -> Maybe (W.Screen WorkspaceId (Layout Window) Window ScreenId ScreenDetail)
+                    screenById sid = L.find ((== sid) . screen) (W.current ws : W.visible ws)
+                    visibleWindowsOnScreen :: ScreenId -> [Window]
+                    visibleWindowsOnScreen sid = L.filter (`elem` (toList mappedWins)) $ W.integrate' $ (screenById sid) >>= W.stack . W.workspace
+                    sortedOverlayWindows :: ScreenId -> X [OverlayWindow]
+                    sortedOverlayWindows sid = sortOverlayWindows <$> (buildOverlayWindows dpy th $ visibleWindowsOnScreen sid)
```
slotThe, line 269: I would indent this (just for readability) a little more like 

``` haskell
  overlays <- case sKeys c of
    AnyKeys ks -> buildOverlays ks <$> sortedOverlayWindows
     where
      visibleWindows :: [Window]
      [...]
```

Much less wasted horizontal space.  Also, Haskell people are quite fond of aligning things instead of having long lines, i.e. instead of 

``` Haskell
PerScreenKeys m -> fmap concat $ sequence $ M.foldr (:) [] $ M.mapWithKey (\sid ks -> buildOverlays ks <$> sortedOverlayWindows sid) m
```

you could write

``` haskell
PerScreenKeys m ->
  fmap concat
    $ sequence
    $ M.foldr (:) []
    $ M.mapWithKey (\sid ks -> buildOverlays ks <$> sortedOverlayWindows sid) m
```

This is your module, so do what you feel like fits best; we're way beyond enforcing any kind of style guide around here :)

XMonad/Actions/EasyMotion.hs
```diff
@@ -0,0 +1,378 @@
...
+    :: X [Overlay]
```
slotThe, line 270: With `{-# LANGUAGE ScopedTypeVariables #-}` you can make this type signature much more readable and write `overlays :: X [Overlay] <- case ...` directly.

slotThe (approved):
This looks really good now!  One could try to make this looks more like "standard" Haskell but honestly I'm at this point more interested in finally getting this merged in a timely fashion and pushing a quick fix pr afterwards if there's any need for that.  I just noticed a few more things that I feel like you should at least have a chance to see before this is merged, but it's nothing pressing really.  

I'll merge this in a few days if nothing comes up; thank you for the contribution!

partiallyordered:
@slotThe thanks again. I'm super, super busy this week. I'd like to make some of the changes you've suggested but most likely I'll get to them early next week. If you'd prefer to merge before then, go ahead. (I guess you'll squash, so I won't?).

P.S., my comment about learning monad transformers was based on a suspicion that I could be doing something a little more elegant than using `io` for about half of the lines in a couple of the main bits of handling, e.g.
https://github.com/xmonad/xmonad-contrib/blob/d52f49066405fe298ea5f6407ebabebbc1604fa0/XMonad/Actions/EasyMotion.hs#L271-L277
and
https://github.com/xmonad/xmonad-contrib/blob/d52f49066405fe298ea5f6407ebabebbc1604fa0/XMonad/Actions/EasyMotion.hs#L351-L361
Any tips would be very welcome (totally acknowledging you have no obligation!). Even "yeah, read this other module" would be great. Don't worry, I won't go messing around with what's already in the PR now, just want to learn better for future usage.

slotThe:
> @slotThe thanks again. I'm super, super busy this week. I'd like to make some of the changes you've suggested but most likely I'll get to them early next week. If you'd prefer to merge before then, go ahead. 

That's totally fine.  If you want to look at this a bit more then I'll just wait until you're satisfied—just don't feel pressured to do anything if you don't want to or don't have the time!

> (I guess you'll squash, so I won't?).

Sure, sounds like a plan.  

> P.S., my comment about learning monad transformers was based on a suspicion that I could be doing something a little more elegant than using `io` for about half of the lines in a couple of the main bits of handling, e.g.
> 
> https://github.com/xmonad/xmonad-contrib/blob/d52f49066405fe298ea5f6407ebabebbc1604fa0/XMonad/Actions/EasyMotion.hs#L271-L277
> 
> and
> 
> https://github.com/xmonad/xmonad-contrib/blob/d52f49066405fe298ea5f6407ebabebbc1604fa0/XMonad/Actions/EasyMotion.hs#L351-L361
> 
> Any tips would be very welcome (totally acknowledging you have no obligation!). Even "yeah, read this other module" would be great. Don't worry, I won't go messing around with what's already in the PR now, just want to learn better for future usage.

To be honest, lifting `IO` actions is just something that you sometimes have to do when working with monad transformers (thankfully, `IO` is the only thing you have to lift).  One could argue that this should perhaps be separated as much as possible, but that seems rather hard for contrib in particular; `X` effects want to mingle with what X11 does quite often.  

Arguably, all of the functions that the X11 library exports should have been `MonadIO` instead of plain `IO` (perhaps with a `specialise` pragma for `IO` to avoid unecessary runtime dictionary passing) so we wouldn't have to do this kind of lifting at all.

XMonad/Actions/EasyMotion.hs
partiallyordered: I'm not sure if you meant dedented, I suspect this comment has suffered from autoindent (and my subsequent negligence). I've dedented it, let me know if you meant something else.

XMonad/Actions/EasyMotion.hs
partiallyordered: Regarding the first suggestion: I prefer that too- I suspect this is another autoindent casualty, except in this case I probably would've assumed my editor indent was accepted style, rather than just having made the mistake of not noticing it!

Regarding the second: much nicer, thanks.

XMonad/Actions/EasyMotion.hs
partiallyordered: Took a while with the compiler patiently repeating at me:
> `overlays` IS `[Overlay]`, NOT `X [Overlay]`
before I finally listened to it. Hah.

I can't remember if it was this type signature or another one, but I definitely had some frustration fighting correct type signature indentation. And in fact I tried this style to see if it would work, but of course it didn't. `ScopedTypeVariables` is _great_ to know about, thank you.

partiallyordered:
Phew, sorry I'm so overdue (in the past couple of weeks, and for.. years..) addressing this. Thanks again for the feedback, and thanks for answering my questions. I'm happy for this to be merged whenever you are.

slotThe:
:heart:

XMonad/Actions/EasyMotion.hs
slotThe: You're completely right; thanks!

XMonad/Actions/EasyMotion.hs
slotThe: Whoops, sorry about that :)

It's a bit unintuitive that `ScopedTypeVariables` should introduce this behaviour I agree, but it definitely is a very useful extension once you know about it

PR #229, opened 2018-03-01
Add three-dimensional group layout

### LINKED ISSUE #214: Layout.Groups should export Uniq and gen
### Problem Description

I want to customize the Groups layout, but can't do it because `Layout.Groups` is missing too many exports, such as `Uniq`, `gen`, and the `Groups` type constructors. Instead of fixing this by adding all the exports, I think Layout.Groups should simply export a new `group3` function.

See this link for one such use (the author ended up forking XMonad to do it):

http://lynnard.me/blog/2013/12/30/more-xmonad-goodies/

### Configuration File

```haskell
module Main (main) where

import XMonad
import XMonad.Util.Stack
import XMonad.Layout.Groups
import Data.Maybe (fromJust)

group3 l l2 l3 = Groups g l3 start (U 2 0)
   where g = group l l2
         start = fromJust $ singletonZ $ G (ID (U 2 1) g) emptyZ

main :: IO ()
main = xmonad def
```

### Checklist

  - [x] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

ccrusius:
This is as per the work done in
http://lynnard.me/blog/2013/12/30/more-xmonad-goodies/, where the
functionality is explained in detail.

This also fixes #214. The original suggestion in that bug report is
not enough. Even if we export `gen` and `Uniq`, we would still have to
export the `Groups` type constructor. I thought it better to simply
allow the user to create a three-dimensional group instead.

### Description

behind them.

### Checklist

### DIFF
--- a/XMonad/Layout/Groups.hs
+++ b/XMonad/Layout/Groups.hs
@@ -22,6 +22,7 @@ module XMonad.Layout.Groups ( -- * Usage
                               -- $usage
                               -- * Creation
                               group
+                            , group3
                               -- * Messages
                             , GroupsMessage(..)
                             , ModifySpec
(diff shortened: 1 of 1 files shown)


byorgey:
Thanks!

PR #231, opened 2018-03-08
Added pretty printer for empty visible workspaces

miguelclean:
### Description
Simple extensions of the pretty printer to differentiate between empty and non-empty visible workspaces. Analogical to the existing functionality for hidden workspaces. Particularly useful if some displays managed by xmonad are turned off temporarily.
### Why would I want this?
My TV is often turned off. Still I want to know if its workspace contains any windows, without having to switch workspaces.
### Checklist
- [x] I've read CONTRIBUTING.md
- [x] I tested my changes with xmonad-testing
- [x] I updated the CHANGES.md file

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,13 @@
     `getXineramaPhysicalWsCompare` now need a extra argument of type `ScreenComparator` defined in
     `XMonad.Actions.PhysicalScreen` (see changelog of this module for more information)
 
+  * `XMonad.Hooks.EwmhDesktops`
+
+    - Simplify ewmhDesktopsLogHookCustom, and remove the gnome-panel specific
+      remapping of all visible windows to the active workspace (#216).
+    - Handle workspace renames that might be occuring in the custom function
+      that is provided to ewmhDesktopsLogHookCustom.
+
 ### New Modules
 
   * `XMonad.Hooks.Focus`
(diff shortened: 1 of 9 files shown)


byorgey:
Thanks for the PR, apologies for the long delay in following up!  Just to make sure I understand, would this change the behavior of the pretty printing for existing users?

miguelclean:
Yes! Workspaces with visible windows will be now handled by two different functions, instead of the first function only:

* ppVisible  = wrap "<" ">"
* ppVisibleNoWindows  = id

Overriding the default behavior for both cases, will require overriding both functions in a custom configuration.

miguelclean:
If this is undesirable I can wrap the new function in a __Maybe__ type and set the default value to _Nothing_, falling back to the _ppVisible_ function. I assumed the simplicity of the current approach outweighs its small impact on existing users.

geekosaur:
It's a fairly large proportion of current xmonad users.

miguelclean:
With __small__ I did not refer to the number of users but my opinion on the severity of the impact. :)

My proposed solution will not _break_ existing configurations but change the way labels of workspaces are displayed (which for most users will probably be the color). This will apply only to non-active, visible workspaces without any windows.

Labels of workspaces falling into any of the following categories will not be affected:
* Current Workspace
* Invisible Workspace
* Visible Worskpace with Windows

Anyway, I see 3 options at this point:
* Leave everything as it is. (Cost: Not knowing if a turned off monitor contains windows)
* Use my initially proposed solution. (Cost: Labels for non-active, visible workspaces without windows will change for existing users)
* Use my other proposed solution with the Maybe wrapping. (Cost: uglier code of higher complexity)

Not sure how to proceed...

geekosaur:
We're still cleaning up from past incompatible changes; I understand that this doesn't actively break things, but it will cause a lot of support requests when people's bars "don't work" any more (and yes, unforutnately, that's how it's usually described. Most people are lousy at characterizing problem reports, and worse at checking change logs).

miguelclean:
I understand. Now I wrapped the new function in a Maybe type:
https://github.com/xmonad/xmonad-contrib/pull/241

It does not change the behaviour for existing users now unless they set the new function explicitly. Its value falls back to _Nothing_ and the old behaviour using _ppVisible_.

PR #233, opened 2018-03-11
Add BinaryColumn layout

ideasman42:
This is similar to 'Column' layout with some differences.

- Add/remove windows keeps window bounds.

  *(with `Column` all window frames re-size whenever windows are added or removed. I find it quite annoying).*
- Enforce minimum window size.

  *(this means you never get a bunch of unrecognizable 1-3 high pixel windows - even if you have many windows open).*

- Scaling the master window down causes an evenly spaced grid.

- Negative scale can be used to increase the size of the last window
  instead of the master window.

### Description

Using xmonad with monitors in portrait mode is well suited to a column grid.
There are however some constraints.

Not that this is different enough from the Column layout that I don't think it can be a drop in replacement. Scale works differently window size minimum has been added.

### Checklist

Attempted this but was getting error:
https://gist.github.com/ideasman42/7bcbaa036e72036275ccd43b5fdf6555

### DIFF
--- a/XMonad/Layout/BinaryColumn.hs
+++ b/XMonad/Layout/BinaryColumn.hs
@@ -10,9 +10,14 @@
 -- Portability :  unportable
 --
 -- Provides Column layout that places all windows in one column.
--- Each window is half the height of the previous.
+-- Each window is half the height of the previous,
+-- except for the last pair of windows.
 --
--- Note: Originally based on XMonad.Layout.Column with edits.
+-- Note: Originally based on 'XMonad.Layout.Column' with changes:
+--
+-- * Adding/removing windows doesn't resize all other windows.
+-- (last window pair exception).
+-- * Minimum window height option.
 --
 -----------------------------------------------------------------------------
 
@@ -26,11 +31,13 @@ import qualified XMonad.StackSet
 import qualified Data.List
 
 -- $usage
--- This module defines layot named BinaryColumn.
+-- This module defines layout named BinaryColumn.
 -- It places all windows in one column.
--- Windows heights are calculated that.
+-- Windows heights are calculated to prevent window resizing whenever
+-- a window is added or removed.
+-- This is done by keeping the last two windows in the stack the same height.
 --
--- You can use this module by adding folowing in your @xmonad.hs@:
+-- You can use this module by adding following in your @xmonad.hs@:
 --
 -- > import XMonad.Layout.BinaryColumn
 --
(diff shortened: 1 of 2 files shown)


XMonad/Layout/BinaryColumn.hs
```diff
@@ -0,0 +1,132 @@
...
+                             -- * Usage
+                             -- $usage
+                             BinaryColumn (..)
+                            ) where
+import XMonad
+import qualified XMonad.StackSet
+import qualified Data.List
+
+-- $usage
+-- This module defines layot named BinaryColumn.
```
byorgey: Should be "layout"

XMonad/Layout/BinaryColumn.hs
```diff
@@ -0,0 +1,132 @@
...
+-- It places all windows in one column.
+-- Windows heights are calculated that.
```
byorgey: This sentence seems to be incomplete?

XMonad/Layout/BinaryColumn.hs
```diff
@@ -0,0 +1,132 @@
...
+-- > import XMonad.Layout.BinaryColumn
+--
+-- Then add layouts to your layoutHook:
+--
+-- > myLayoutHook = BinaryColumn 1.0 32 ||| ...
+--
+-- The first value causes the master window to take exactly half of the screen,
+-- the second ensures that windows are no less than 32 pixels tall.
+--
+-- Shring/Expand can be used to adjust the first value by increments of 0.1.
```
byorgey: Shrink

XMonad/Layout/BinaryColumn.hs
```diff
@@ -0,0 +1,132 @@
...
+--
+-- 2.0 uses all space for the master window
+-- (minus the space for windows which get their fixed height)
+-- 0.0 gives an evenly spaced grid.
+-- Negative values reverse the sixes so the last
```
byorgey: Should be "sides" instead of "sixes"?

byorgey:
@ideasman42  Looks good to me, thanks!  Could you fix the documentation things I mentioned? (Just push another commit to the `layout-binary-column` branch.) Then I will merge.

ideasman42:
Thanks for feedback, there were indeed some odd mistakes! Corrected and added some more description of how this is different.

ideasman42:
Checks failing seem unrelated to this PR.

geekosaur:
Looks like we do builds on Ubuntu Precise, and hvr removed support for that distro.

byorgey:
Thanks! Sorry for the long delay in merging.

LSLeary:
@byorgey @ideasman42 
I happened to notice when testing something unrelated that I couldn't actually import this module. The xmonad-contrib package doesn't expose it; looks like it never got added to the cabal file. In the process of figuring that out I also noticed that the changelog erroneously lists it as `XMonad.Hooks.BinaryColumn`.

byorgey:
@LSLeary Thanks, good catch!  Both fixed now.

PR #240, opened 2018-04-04
Implement additional options for Dzen

ae-g-i-s:
### Description

Some fairly common options to `dzen2` are not yet wrapped in `XMonad.Util.Dzen`.
This pull request adds the following functions, meant to be used with `dzenConfig`:
* `fgColor` and `bgColor` to set foreground/background color
* `align` and `slaveAlign` to set text alignment (using the `Align` data type from `XMonad.Util.Font`)
* `lineCount` to enable the slave window and specify the number of lines it should contain

Rather trivial change, please check formatting/style since I don't often contribute Haskell.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -239,6 +239,13 @@
     passed to `dzen2` invocation. The behaviour of current `dzen` function is
     unchanged.
 
+  * `XMonad.Util.Dzen`
+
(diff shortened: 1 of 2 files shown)


byorgey:
Looks good to me, thanks!

PR #242, opened 2018-04-20
Border Control

orbisvicis:
#### `XMonad.Layout.NoBorders`

The layout now maintains a list of windows that never have borders, and a list of windows that always have borders. Use `BorderMessage` to manage these lists and the accompanying event hook (`borderEventHook`) to remove destroyed windows from them. Also provides the `hasBorder` manage hook.

Two new conditions have been added to `Ambiguity`: `OnlyLayoutFloat` and `OnlyLayoutFloatBelow`; `OnlyFloat` was renamed to `OnlyScreenFloat`. See the documentation for more information.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -100,6 +100,13 @@
     `OnlyLayoutFloatBelow`; `OnlyFloat` was renamed to `OnlyScreenFloat`.  See
     the documentation for more information.
 
+    The type signature of `hiddens` was changed to accept a new `Rectangle`
+    parameter representing the bounds of the parent layout, placed after the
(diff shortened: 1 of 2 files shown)


orbisvicis:
Incompatibilities:

 * `hiddens` has a new type signature. I don't think there's any way around this; I need the extra information.
 * `OnlyFloat` renamed to `OnlyScreenFloat` because it is otherwise, ironically, ambiguous. I can always change it back, though what's *one* extra incompatible change?

byorgey:
Sounds OK to me.  Could you add something (ideally to both `CHANGES.md` and to the Haddock documentation) which concisely explains how to upgrade (*e.g.* for users who were using `hidden` before and will now have to modify their configs)?

PR #243, opened 2018-04-20
reSpaced

orbisvicis:
#### `XMonad.Layout.Spacing`

Rewrite `XMonad.Layout.Spacing`. Borders are no longer uniform but composed of four sides each with its own border width. The screen and window borders are now separate and can be independently toggled on/off. The screen border examines the window/rectangle list resulting from 'runLayout' rather than the stack, which makes it compatible with layouts such as the builtin `Full`. The child layout will always be called with the screen border. If only a single window is displayed (and `smartBorder` enabled), it will be expanded into the original layout rectangle. Windows that are displayed but not part of the stack, such as those created by 'XMonad.Layout.Decoration', will be shifted out of the way, but not scaled (not possible for windows created by XMonad). This isn't perfect, so you might want to disable `Spacing` on such layouts.

### Checklist

### DIFF
--- a/XMonad/Layout/Spacing.hs
+++ b/XMonad/Layout/Spacing.hs
@@ -31,7 +31,6 @@ module XMonad.Layout.Spacing
     ) where
 
 import           XMonad
-import qualified XMonad.StackSet                as W
 import           XMonad.Layout.LayoutModifier
 import qualified XMonad.Util.Rectangle          as R
 
@@ -55,7 +54,7 @@ data Border = Border
     , bottom    :: Integer
     , right     :: Integer
     , left      :: Integer
-    } deriving (Read,Show)
+    } deriving (Show,Read)
 
 -- | A 'LayoutModifier' providing customizable screen and window borders.
 -- Borders are clamped to @[0,Infinity]@ before being applied.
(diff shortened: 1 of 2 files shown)


LSLeary:
It's not up to me, but changes that break user configs generally won't be accepted without a good reason. You can rewrite internal details all you want, but you should then seek to embed the interface originally exposed by the module within the new interface, even if it's totally redundant.

So for example, my xmonad.hs that imports the `SpacingWithEdge` type and the functions `incSpacing`, `decSpacing` and `spacingWithEdge` should continue to compile and, generally, exhibit the same runtime behaviour (excluding any squashed bugs).

If you wish you can use the [DEPRECATED](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#warning-and-deprecated-pragmas) pragma with any exports that have been superseded, though only people who use -Werror or open their xmonad.hs in ghci will ever notice the warning, since `xmonad --recompile` suppresses warnings.

NB: This comment probably applies to some of your other PRs too; I haven't looked them over.

orbisvicis:
Believe it or not, only reason I didn't is I couldn't think of a good name for my new `spacing` function. Maybe I'll go with `spacingRaw`? Input appreciated. Other than that, the only thing I don't think should be backwards compatible is `modifierDescription` because it no longer really applies, specifically "WithEdge" and the number... what do you think?

Regarding the other PRs, backwards compatibility isn't so cut and dry so I'd rather not make changes now.

XMonad/Layout/Spacing.hs
```diff
@@ -1,139 +1,341 @@
...
+    EQ -> eq
+    GT -> gt
 
-    pureModifier (Spacing p) _ _ wrs = (map (second $ shrinkRect p) wrs, Nothing)
+-----------------------------------------------------------------------------
+-- Backwards Compatibility:
+-----------------------------------------------------------------------------
+{-# DEPRECATED spacing, spacingWithEdge, smartSpacing, smartSpacingWithEdge "Use spacingRaw instead." #-}
+{-# DEPRECATED setSpacing "Use setWindowSpacing/setScreenSpacing instead." #-}
+{-# DEPRECATED incSpacing "Use incWindowSpacing/incScreenSpacing instead." #-}
```
IvanMalison, line 304: Frankly insane to have deprecated all of these.

the fact that uniformBorder is not even exported for some reason is also insane.

PR #244, opened 2018-04-20
Rectangle

orbisvicis:
#### `XMonad.Util.Rectangle`

A new module for handling pixel rectangles.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -108,6 +108,10 @@
     A new layout based on Dishes, however it accepts additional configuration
     to allow multiple windows within a single stack.
 
+  * `XMonad.Util.Rectangle`
+
+    A new module for handling pixel rectangles.
+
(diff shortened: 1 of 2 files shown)


byorgey:
In various places we already use the `Rectangle` type which is defined in the `X11` package: http://hackage.haskell.org/package/X11-1.9/docs/Graphics-X11-Xlib.html#t:Rectangle .  How is this different / why do we need both?

orbisvicis:
This is a utility module for working with `Rectangle`s from `X11`. About half the operations are simpler with point/point rectangles rather than point/dimension rectangles. So `PointRectangle` is just a helper type for encapsulating the possible conversion results.

I created `Util.Rectangle` after noticing duplicated geometric operations in other modules, and a few of my pull requests depend on this module.

PR #245, opened 2018-04-20
Vimish Prompt

orbisvicis:
#### `XMonad.Prompt`

* Prompt ships a vim-like keymap, see `vimLikeXPKeymap` and `vimLikeXPKeymap'`. A reworked event loop supports new vim-like prompt actions.
* Prompt supports dynamic colors. Colors are now specified by the `XPColor` type, so update your `XPConfig` to match.
* Fixes `showCompletionOnTab`.
* The behavior of `moveWord` and `moveWord'` has changed; brought in line with the documentation and now internally consistent. The old keymaps retain the original behavior; see the documentation to do the same your XMonad configuration.

It should be noted that the bug I mention in the source was there before my changes.

### Checklist


orbisvicis:
Incompatibilities:

 * `XPConfig` colors bundled into `XPColor`. This can be reverted and `XPColor` kept & constructed for the state only.
 * `moveWord`/`moveWord'` behavior changed. I consider the old behavior a bug because it runs contradictory to the documentation and isn't internally consistent (characters moved depends on the direction). While I really don't think this should be reverted, the old keymaps do preserve the original behavior, which is "good enough".

XMonad/Prompt.hs
```diff
@@ -15,6 +15,17 @@
...
+-----------------------------------------------------------------------------
+-- Bugs:
+-- if 'alwaysHighlight' is True, and
+--  1 type several characters
+--  2 tab-complete past several entries
+--  3 backspace back to the several characters
+--  4 tab-complete once (results in the entry past the one in [2])
+--  5 tab-complete against this shorter list of completions
+-- then the prompt will freeze (XMonad continues however).
+-----------------------------------------------------------------------------
```
byorgey: Is this still accurate?

byorgey:
I can see both sides but I would lean towards keeping `XPConfig` the same and just bundling the color information into `XPColor` internally.  Is there any reason to change the `XPConfig` that I'm missing, other than "`XPColor` exists so we might as well use it"?

XMonad/Prompt.hs
orbisvicis: Yes, before/after this PR.

orbisvicis:
No reason. Backwards compatibility restored, but I'll keep this open for a last review.

byorgey:
Hey @orbisvicis , I was just going back through some old todos and saw this was never merged.  Sorry!  In the meantime it looks like there is now a conflict since some other change was made to the `XPConfig`.  Any thoughts?

byorgey:
Great, thanks!

Issue #191, opened 2017-06-06
X.L.NoBorders not working correctly together with X.L.Groups


ankaan:
### Problem Description

When using smartBorders on a X.L.Groups layout where the inner layout is Full, smartBorders fails to add borders back to anything but the master window in the group.

### How to reproduce

* Run xmonad with the below configuration file
* Press Down Arrow
* Press Shift + Right Arrow.

Expected result: Two windows side by side both with borders.
Actual result: Only the left window has a border (the one that was master in the group.)

If you instead spawn a bunch of windows using Return, and then repeatedly press Down Arrow,
in stdoud the number of windows without borders will be printed. It should stay at 1, but the output shows that the number of such windows grows. Only the master window gets it's border back when it is no longer focused.

The debug output shows that the the same issue actually happens with noBorders, but is not noticable since it always removes the border when a window is shown.

The issue is tied to the actions used to change focus, namely GE.focusMaster, GE.focusUp and GE.focusDown. The debug output shows that the old focused window does not gets its border back. But when xK_h and xK_l (windows W.focusUp and windows W.focusDown) is used in the below config, the borders are return correctly.

X.L.Groups is doing something odd when changing focus which makes X11 unable to change the border width, but I have not been able to figure out how to fix this.

### What I have tried so far

I have been trying to fix this by modifying X.L.NoBorders. When modifying redoLayout for ConfigurableBorder as following, things works ok (but with a warning since we no longer need "s".) The issue is that the user can no longer manually remove a border from a window using for example X.A.NoBorders. So this is not really a good solution.

```haskell
    redoLayout (ConfigurableBorder p s) _ mst wrs = do
        ws <- withWindowSet (\wset -> return (hiddens p wset mst wrs))
        asks (borderWidth . config) >>= setBorders (W.integrate' mst \\ ws)
        setBorders ws 0
        return (wrs, Just $ ConfigurableBorder p ws)
```

When I instead make changes so that we only remove windows from the remembered list if they are confirmed as having borders, or are no longer part of the stack, the problem remains. So there is something funky going on with checking if there is any border as well.

```haskell
    redoLayout (ConfigurableBorder p s) _ mst wrs = do
        s' <- filterM withoutBorder $ s `intersect` W.integrate' mst
        ws <- withWindowSet (\wset -> return (hiddens p wset mst wrs))
        asks (borderWidth . config) >>= setBorders (s' \\ ws)
        setBorders ws 0
[...]

PR #247, opened 2018-04-21
X.A.MessageFeedback fix refresh behaviour

ankaan:
### Description

This fixes the issue "X.L.NoBorders not working correctly together with
X.L.Groups" (#191).

Generaly when the users sends a message, they expect the layout to be
updated, not updating they layout is the exception. In the documentation
for the send method it mentioned that it worked like X.O.sendMessage,
even though it actually behaved more like X.O.sendMessageWithNoRefresh.
This PR changes this behaviour so that send actually does an update at
the end as expected.

This fixes the behaviour of X.L.NoBorders together with X.L.Groups, when
using X.L.G.Examples and X.L.G.Helpers to send messages.

I have added a new function for each old one that keeps the old
behaviour, but refreshing is the new default. The only place this module
is used in xmonad-contrib is in X.L.G.Helpers, which is the code we
actually wanted to fix.

An alternative to this is that keep the old behaviour in the existing
functions of X.A.MessageFeedback. But looking at X.O.sendMessage and
X.O.sendMessageWithNoRefresh from xmonad, this seems like the reasonable
path forward.

### Checklist

  - [*] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

### DIFF
--- a/XMonad/Actions/MessageFeedback.hs
+++ b/XMonad/Actions/MessageFeedback.hs
@@ -18,18 +18,26 @@ module XMonad.Actions.MessageFeedback (
                                       -- $usage
 
                                         send
+                                      , sendWithNoRefresh
                                       , tryMessage
                                       , tryMessage_
+                                      , tryMessageWithNoRefresh
+                                      , tryMessageWithNoRefresh_
                                       , tryInOrder
                                       , tryInOrder_
+                                      , tryInOrderWithNoRefresh
+                                      , tryInOrderWithNoRefresh_
                                       , sm
                                       , sendSM
                                       , sendSM_
+                                      , sendSMWithNoRefresh
+                                      , sendSMWithNoRefresh_
                                       ) where
 
-import XMonad.Core ( X (), Message, SomeMessage(..), LayoutClass(..), windowset, catchX )
+import XMonad.Core ( X (), Message, SomeMessage(..), LayoutClass(..), windowset, catchX, whenJust )
 import XMonad.StackSet ( current, workspace, layout, tag )
-import XMonad.Operations ( updateLayout )
+import XMonad.Operations ( updateLayout, windows )
+import qualified XMonad.StackSet as W
 
 import Control.Monad.State ( gets )
 import Data.Maybe ( isJust )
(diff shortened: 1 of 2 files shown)


XMonad/Actions/MessageFeedback.hs
```diff
@@ -63,37 +71,79 @@ import Control.Applicative ((<$>))
...
 -- | Convenience shorthand for 'XMonad.Core.SomeMessage'.
 sm :: Message a => a -> SomeMessage
 sm = SomeMessage
 
 
+-- | Send a 'SomeMessage' to the corrent layout and returns true if the message was handled.
 sendSM :: SomeMessage -> X Bool
 sendSM m = do w <- workspace . current <$> gets windowset
               ml' <- handleMessage (layout w) m `catchX` return Nothing
-              updateLayout (tag w) ml'
```
byorgey: Maybe I'm misunderstanding something --- why isn't `updateLayout` called any more?

byorgey:
Seems generally reasonable to me.  @geekosaur , do you have any opinion on this approach?

XMonad/Actions/MessageFeedback.hs
geekosaur: @byorgey, @ankaan doing an update (especially with refresh) from a message handler potentially can lead to infinite loops, some care is needed.

geekosaur: "update" here includes `X.O.windows` which will do a `refresh` and run `logHook` and potentially `manageHook` if there are new windows waiting to be managed.

ankaan: According to the old documentation, a refresh is expected. But that is not reflected in the actual old implementation.
The risk here is that some user has configured one of their hooks to use i.e. sendSM which could potentially trigger an infinite loop. By trial and error they would have discovered that this worked before the change. But what is actually the correct behaviour?

ankaan: Having to manually add refresh even on simple actions makes it harder to use the MessageFeedback module though. A refresh after each action is generally what is expected, which is reflected in the name of the methods. The long bulky name is the less used non-refresh variant. (As in xmonad core.)

The non-refresh is still there though if you need it and in that case you probably know you need it. It is harder to understand that you manually need to refresh after a simple action which when sending a normal message without feeback just works.

byorgey:
So I still think this (having built-in refresh be the default but having no-refresh versions available too) is a reasonable approach.  But after thinking about this some more my big concern is that since this PR simply changes the behavior of existing functions by adding a refresh call, anyone currently using these functions in a sensitive place (e.g. inside a message handler) could have their xmonad hang as soon as they update.  I am not sure it is reasonable to expect that such people will carefully read the changelog to see what changed, especially if their configuration still compiles without error.  But then when their xmonad starts to hang it could be very difficult for them to figure out what is wrong.

How prevalent do we think such usage might be?  How could we make sure such users have a smooth upgrade path?

ankaan:
The module has been around since 2009, so it is not a recent addition. But I think these functions are a bit more advanced than many others.

One way of somewhat keeping compatibility is to make the changes in new modules and deprecate the old. In this way we can keep backward compatibility, but the namespace will be cluttered up with a module that does the same thing, just very slightly differently. If the old module is later removed a compilation error will happen instead of sudden freeze. (This would involve both X.A.MessageFeedback and X.L.G.Helper.)

An alternative is to keep the old functions but name the new ones with refresh differently. But this will give us longer confusing function names that does not match what is in xmonad core.

I don't think this is much of a problem with X.L.G.Helper, since these are very much targeted towards setting up keyboard shortcuts. I don't see much use of these for automatic actions to take based on window events and such.

This leaves us with MessageFeedback triggered on window events and inside custom layouts not added to xmonad contrib.

Am I missing something? What other use cases are there where this code could be triggered based on a refresh?

orbisvicis:
Looks like I accidentally walked into this: rewrote X.A.MessageFeedback (#261) for X.L.Spacing (#260). @ankaan, what do you think?

LSLeary:
The documentation for the module is (was) ambiguous. `sendMessageWithNoRefresh` and `sendMessage` differ in more ways than just the refresh (the latter targets the current workspace) so the docs shouldn't necessarily be taken to imply that a refresh is being performed.

Providing alternative versions so that a refresh is up to the user is good, but changing the default behaviour imo is not. #261 sidesteps this issue by leaving deprecating the old functions without changing their behaviour, and providing a less ambiguous version of the module with desirable new functionality. As such I prefer (and have merged) it. Further discussion in #191.

PR #248, opened 2018-04-23
Fullscreen.hs: don't lay out windows obscured by fullscreen

jktomer:
## Description

There's no reason to return a rectangle for any window that is totally
obscured by a full-screen window, and not doing so has the nice property that
when hidden windows' borders overlap with a full-screen window's, the user
will not be confused by overlapping partially-drawn borders. It also makes the
Fullscreen modifiers combine much better with smartBorders.

## Checklist

- [x]  I've read CONTRIBUTING.md
- [ ]  I tested my changes with xmonad-testing
  - xmonad-testing has no configs with a Fullscreen layout; I have tested on my own system and this works as described.
- [x]  I updated the CHANGES.md file

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -135,6 +135,11 @@
     Added `sideNavigation` and a parameterised variant, providing a navigation
     strategy with fewer quirks for tiled layouts using X.L.Spacing.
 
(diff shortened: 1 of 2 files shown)


byorgey:
Thanks!

PR #253, opened 2018-04-29
StateFull: Fixing Full (and other layouts we lie to)

LSLeary:
### Description

_TL;DR: added a module providing a version of Full that doesn't misbehave when you focus a float._

Due to the `Stack` data type with which we pass windows to layouts, we're forced to claim that one of the windows being tiled has focus. Layouts like `Full` and `Accordion` depend strongly upon the correctness of this information, and misbehave when lied to (because a float has focus, or because the layout is child to some parent layout that hasn't given it the focus).

This module implements a layout transformer `FocusTracking` that just holds onto the most recent true focus it's seen and passes the information down to the transformed layout. This means the layout remembers where it was when it loses focus, and doesn't suddenly go jumping off to some other window. `StateFull` (the replacement for `Full`) is just a cute pattern synonym provided for the main use-case; it's equivalent to `focusTracking Full` or `FocusTracking Nothing Full`.

Honestly I don't know why this hasn't already been written; the idea and the implementation are both quite simple. If it has and I've just missed it, then point me to it and I can close this PR.

Edit: I realised the default `description` method wasn't static (on account of `show`ing internal state) so I added a better one.

### Checklist

    - I tested my changes by building and using them as my system xmonad-contrib alongside current git xmonad. I don't know if there's any point in it, but I did also try to test them with xmonad-testing ... it built fine, but trying to test in xephyr while working around the outer xmonad instance capturing my hundreds of keyboard and mouse bindings is way too painful, so I gave up instantly.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -76,6 +76,15 @@
 
 ### New Modules
 
+  * `XMonad.Layout.StateFull`
+
+    Provides StateFull: a stateful form of Full that does not misbehave when
+    floats are focused, and the FocusTracking layout transformer by means of
+    which StateFull is implemented. FocusTracking simply holds onto the last
+    true focus it was given and continues to use it as the focus for the
+    transformed layout until it sees another. It can be used to improve the
+    behaviour of a child layout that has not been given the focused window.
+
   * `XMonad.Hooks.Focus`
 
     A new module extending ManageHook EDSL to work on focused windows and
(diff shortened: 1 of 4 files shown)


byorgey:
Thanks!

liskin:
> Honestly I don't know why this hasn't already been written

If I'm not mistaken, it has: https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Layout/TrackFloating.hs
There seem to be minor differences in the condition when the remembered focus point is used and TrackFloating is implemented as a LayoutModifier, but practically they should behave the same in most situations.

So I wonder if perhaps we could consolidate the two somehow?

liskin:
Here's what I have so far: https://github.com/xmonad/xmonad-contrib/pull/418
I like @LSLeary's implementation more so I adapted TrackFloating (which I've been using for 10 years) to be more like it. I think keeping the one which is a LayoutModifier rather than a separate LayoutClass instance is better.

PR #254, opened 2018-05-15
Allow build with X11-1.9

mimi1vx:
### Description

Allow build of xmonad-contrib with X11-1.9

### Checklist

### DIFF
--- a/xmonad-contrib.cabal
+++ b/xmonad-contrib.cabal
@@ -64,7 +64,7 @@ library
                    random,
                    mtl >= 1 && < 3,
                    unix,
-                   X11>=1.6.1 && < 1.9,
+                   X11>=1.6.1 && < 1.10,
                    xmonad>=0.13   && < 0.14,
                    utf8-string,
                    semigroups


byorgey:
So what changed between 1.8 and 1.9?  Do we know it won't affect xmonad?  (@dmwit , any insights?)

geekosaur:
IIRC the main change is that there are some added required dependencies that used to be optional, e.g. `-lXScrnSaver`. No functionality that xmonad depends on is affected, iirc.

byorgey:
Thanks @dmwit ! We do call `queryTree` in a few places (https://github.com/xmonad/xmonad-contrib/blob/54ef9f6f8d4839b04eeee186a689485ca4e3fd80/XMonad/Util/WindowProperties.hs#L70, https://github.com/xmonad/xmonad-contrib/blob/ca9b7d9dfcaf972152731420fe9f1ef01d967f5d/XMonad/Hooks/ManageDocks.hs#L157) but if no one ever ran into a problem with it before, I'm not super motivated to figure out the right way to add an exception handler...

PR #260, opened 2018-06-06
Compatibility tweaks to X.L.Spacing following rewrite.

LSLeary:
### Description

When updating my system xmonad-contrib to git current I found the backwards compatibility of the rewritten spacing module insufficient. Overview of my changes and reasoning where applicable:
  - `SmartSpacing`, `SpacingWithEdge` and `SmartSpacingWithEdge` have been reintroduced as type synonyms of `Spacing`.
  - `spacing`, `smartSpacing`, `spacingWithEdge`, `smartSpacingWithEdge`, `incSpacing` and `setSpacing` have all been reverted to accept an `Int` rather than an `Integer`.
  - `incSpacing` and `setSpacing` have been reimplemented; as they were each use would have made changes to the layout over the course of two refreshes rather than just one. Moreover, they're no longer under deprecation; they expose a popular functionality with simplicity and efficiency, and that role has not been replaced.
  - The `ModifySpacing` message constructor has been reintroduced. I believe it formed an important part of the original interface.

With these changes almost no configs will be broken, which is nice. Hopefully I haven't missed anything.

@orbisvicis This is your rewrite; if you have any input please do weigh in. 

@byorgey This is only tangentially related, but in future should I just make this kind of change without bothering with a PR? What about more substantial changes, like a new module of my own writing? Or other people's PRs? I'm not sure what standard procedure is or should be.

### Checklist

    - As usual, changes tested in my system instance.

### DIFF
--- a/XMonad/Layout/Spacing.hs
+++ b/XMonad/Layout/Spacing.hs
@@ -24,19 +24,21 @@ module XMonad.Layout.Spacing
     , ModifySpacing (..)
     , spacingRaw
     , setSmartSpacing
+    , setSpacing
     , setScreenSpacing, setScreenSpacingEnabled
     , setWindowSpacing, setWindowSpacingEnabled
     , toggleSmartSpacing
     , toggleScreenSpacingEnabled
     , toggleWindowSpacingEnabled
+    , incSpacing
     , incWindowSpacing, incScreenSpacing
     , decWindowSpacing, decScreenSpacing
     , borderIncrementBy
       -- * Backwards Compatibility
       -- $backwardsCompatibility
+    , SmartSpacing, SpacingWithEdge, SmartSpacingWithEdge
     , spacing, spacingWithEdge
     , smartSpacing, smartSpacingWithEdge
-    , setSpacing, incSpacing
     ) where
 
 import           XMonad
@@ -175,6 +177,9 @@ instance Eq a => LayoutModifier Spacing a where
         = Just $ s { windowBorder = f wb }
         | Just (ModifyWindowBorderEnabled f) <- fromMessage m
         = Just $ s { windowBorderEnabled = f wbe }
+        | Just (ModifySpacing f) <- fromMessage m
+        = Just $ let modSp = borderMap (fromIntegral . max 0 . f . fromIntegral)
+                 in  s { screenBorder = modSp sb, windowBorder = modSp wb }
         | otherwise
         = Nothing
 
@@ -199,6 +204,7 @@ data ModifySpacing
     | ModifyScreenBorderEnabled (Bool -> Bool)
     | ModifyWindowBorder (Border -> Border)
     | ModifyWindowBorderEnabled (Bool -> Bool)
+    | ModifySpacing (Int -> Int)
     deriving (Typeable)
 
 instance Message ModifySpacing
@@ -252,6 +258,14 @@ decWindowSpacing = incWindowSpacing . negate
 decScreenSpacing :: Integer -> X ()
 decScreenSpacing = incScreenSpacing . negate
 
+-- | Set all borders to a uniform size.
+setSpacing :: Int -> X ()
+setSpacing = sendMessage . ModifySpacing . const
+
+-- | Increment both screen and window borders uniformly.
+incSpacing :: Int -> X ()
+incSpacing = sendMessage . ModifySpacing . (+)
+
 -- | Map a function over a 'Border'. That is, over the four individual borders.
 borderMap :: (Integer -> Integer) -> Border -> Border
 borderMap f (Border t b r l) = Border (f t) (f b) (f r) (f l)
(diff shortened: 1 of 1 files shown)


orbisvicis:
@LSLeary: have some suggestions coming soon.

byorgey:
@LSLeary Thanks for asking. I meant to send you and @orbisvicis a message about this but there wasn't any way to send a message along with the invite itself, and then it just slipped my mind.

There isn't really a standard procedure.  In general I would suggest something like the following:

1. For really simple things like typo fixes, etc., you should definitely feel free to just push without making a PR.
2. For anything more substantial (such as this PR itself) I would err on the side of opening a PR instead of just pushing.  This is not a "rule" but just a best practice that I try to follow myself.  I can't tell you how many times I have made a change or added a feature that I thought was "obvious", but opened a PR anyway, and ended up getting substantive feedback that improved the code in ways I hadn't even thought of.
3. However, there's a balance between getting feedback from others and simply holding up the process.  Feel free to say things like "If I don't hear anything in one week I'll just go ahead and merge."
4. In terms of other people's PRs, push access is push access.  If you review a PR and it looks good to you, go ahead and merge it.  In fact, the reason I added you and @orbisvicis is because it recently dawned on me that I was the only person with push access who was still actively reviewing PRs, and I simply don't have enough time and energy to review everything that is being submitted these days.

orbisvicis:
@LSLeary: what do you think about 2fde742e7a632f2f982cb03417abb6f0d2039f8d instead?

> ... have been reintroduced as type synonyms of Spacing.

I don't much see the point to this, since they'll still have incompatible signatures and that might make things confusing.

> ... have all been reverted to accept an Int rather than an Integer

+1

> `incSpacing` and `setSpacing` have been reimplemented; as they were each use would have made changes to the layout over the course of two refreshes rather than just one. Moreover, they're no longer under deprecation; they expose a popular functionality with simplicity and efficiency, and that role has not been replaced.

That's a good point but I can see wanting to send other combinations of `ModifySpacing` messages. So instead I implemented `sendMessages` in `XMonad.Actions.MessageFeedback`, see #261. The functions `setScreenWindowSpacing`, `incScreenWindowSpacing`, `decScreenWindowSpacing` replace `incSpacing` and `setSpacing` and aren't deprecated.

> The ModifySpacing message constructor has been reintroduced. I believe it formed an important part of the original interface.

I don't really agree with this:

* No logic should should reside within the message handlers, for maximum flexibility. It is easier to extend if the logic is encapsulated by the message senders.
* Messages should apply to the entire border structure rather than individual integer borders. Without this operations such as `borderIncrementBy` aren't possible (at least not without violating the above point).
* Any new implementation of `incSpacing` should continue to use `borderIncrementBy` for consistency with `incScreenSpacing` and `incWindowSpacing` (now that borders are 4 values rather than 1, a single `max` won't suffice).

orbisvicis:
@byorgey I do have eventual plans to fix old bugs but unfortunately I won't really have time do anything constructive here for quite a while.

LSLeary:
@orbisvicis Sorry it took me a while to get back to this; I'd been putting off some more important things for too long, so I prioritised them. 

> I don't much see the point to this, since they'll still have incompatible signatures and that might make things confusing.

If by incompatibility you mean there can be values with type signatures suggesting facts about the value that are not correct, then yes, these type synonyms will make that possible and it could cause confusion. But I believe that omitting the synonyms and hence breaking configs is far more likely to cause problems. Unless there is breakage they fail to prevent that I'm not seeing?

> The functions `setScreenWindowSpacing`, `incScreenWindowSpacing`, `decScreenWindowSpacing` replace `incSpacing` and `setSpacing` and aren't deprecated.

This all looks fine; I don't really care about the details so long as we're not wasting a refresh.

> * No logic should should reside within the message handlers, for maximum flexibility. It is easier to extend if the logic is encapsulated by the message senders.

I agree with this in principle, but I don't think it trumps immediate practical concerns.

> * Messages should apply to the entire border structure rather than individual integer borders. Without this operations such as `borderIncrementBy` aren't possible (at least not without violating the above point).
> * Any new implementation of `incSpacing` should continue to use `borderIncrementBy` for consistency with `incScreenSpacing` and `incWindowSpacing` (now that borders are 4 values rather than 1, a single `max` won't suffice).

These are details I'm not concerned with, I just trying to preserve a functional imitation of any previously exposed value or type. Personally I think including suboptimal code in the form of an additional constructor is the simplest and most practical way to achieve that, though perhaps you would want to alter the details of the implementation as per your above points.

LSLeary:
> What is the point of having the type without the constructors; where could it be used in a configuration?

The way I noticed the missing types was actually because it broke my own config when I updated. Anywhere someone used one of the old "constructor" functions like `spacingWithEdge` in a top level declaration with a type signature will break without the type synonym. The actual constructors don't appear to have ever been exposed, so we don't need to worry about them.

> Yes, but do the functions or the types represent the interface? For example, wrapper functions exist so there is no need to use `ModifySpacing` directly.

I'm not 100% on what you're asking here, but I'd generally say the functions/values represent the interface, with the type of each value almost inextricably linked to it. If it's a question of regular bindings or constructors forming the interface, that's just a matter of what gets exposed. Originally the `ModifySpacing` constructor had importance because all you could do without it was set/inc, and if you needed composability you couldn't do anything since the `sendMessage` was baked in to the functions.

> That said, I'd be pretty OK just by removing `max 0` from the `ModifySpacing` handler (so all my three points are satisfied across the module).

It's not a huge sticking point, but I'd rather the message handling still prevent borders being pushed into negatives. Rather than the `borderMap (fi . max 0 . f . fi)` I have now, would something like `borderMap (fi . f . fi)` composed with `borderIncrementBy 0` be fine? Not sure whether left or right or both is best. I know it doesn't keep the logic out of the message handling, but it's not like we're going to have a function wrapping this message and passing in the logic instead.

> Do you know how to deprecate that constructor without affecting a type of the same name?

As per our chat in #haskell, no. But I'd like to restate here that I think separating the new message constructors into their own type is the best solution. Layouts seamlessly handle multiple message types, and the `ModifySpacing` type and its constructor can both be relegated to the deprecated section, so I don't really see how this is heavy or otherwise what its shortfall is.

> Also someone is going to have to review #261 keeping in mind #247.

I'm not making any promises, but I can probably do that after I ~finish up a few other things I'm focusing on.

LSLeary:
@orbisvicis Yeah, that's what I meant re separating the constructors. This looks OK, but since it depends on #261 we can't merge it yet. Just so it has somewhere noticeable to sit in the meantime, do you want to put it in a PR? I'll close this in favour of it.

LSLeary:
Closed in favour of #273

PR #263, opened 2018-06-19
Wakeup reduction in EwmhDesktops

### LINKED ISSUE #227: Changing a window title quickly with ewmh enabled makes xmonad unresponsive
### Problem Description

When i use an app that can change window titles quickly (say, `paludis` or `tkabber`), xmonad doesn't respond for a while after such events, burning cpu cycles.

### Configuration File

I use `urxvt` and `for nvm in $(seq 0 1000); do echo -e '\e]2;test\a'; done` to reproduce it. On my config (https://github.com/l29ah/xmonad-config) this makes xmonad unresponsive for the next 30 seconds or so, with the minimal test config the time is about 5s (yet if i drop ewmh from mine it doesn't slow xmonad at all).

```haskell
module Main (main) where

import XMonad

main = xmonad $ ewmh defaultConfig
```

### Checklist

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

bgamari:
### Description

Previously the `EwmhDesktops` was painfully naive, sending property updates every time the log hook was invoked, even if none of the affected properties actually changed in value. This hurts significantly since it can cause a slew of unnecessary wakeups. Here we refactor the EWMH implementation to cache the current property value and only send updates if the value actually changes.

Previously `xmonad` was consistently sitting at 5% CPU usage when I was using my system (and `taffybar` even higher as it redraws its entire window when desktop names change). With this patch it sites at < 1%, exactly where it should be.

Fixes #227.

### Checklist

### DIFF
--- a/XMonad/Hooks/EwmhDesktops.hs
+++ b/XMonad/Hooks/EwmhDesktops.hs
@@ -90,6 +90,15 @@ data DesktopState
 instance ExtensionClass DesktopState where
     initialValue = DesktopState [] [] Nothing M.empty
 
+-- |
+-- The value of @_NET_ACTIVE_WINDOW@, cached to avoid unnecessary property
+-- updates.
+newtype ActiveWindow = ActiveWindow Window
+                     deriving (Eq)
+
+instance ExtensionClass ActiveWindow where
+    initialValue = ActiveWindow none
+
 toDesktopState :: ([WindowSpace] -> [WindowSpace]) -> WindowSet -> DesktopState
 toDesktopState f s =
     DesktopState
@@ -135,8 +144,9 @@ ewmhDesktopsLogHookCustom f = withWindowSet $ \s -> do
 
         mapM_ (uncurry setWindowDesktop) (M.toList $ windowDesktops s')
 
-    setActiveWindow
-
+    let activeWindow' = fromMaybe none (W.peek s)
+    whenChanged (ActiveWindow activeWindow') $ do
+        setActiveWindow activeWindow'
 
 -- |
 -- Intercepts messages from pagers and similar applications and reacts on them.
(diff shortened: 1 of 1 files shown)


bgamari:
Correct me if I'm wrong but I don't think the CI issue is due to my patch.

IvanMalison:
@bgamari Ahh. Thanks for investigating this. I was wondering why CPU usage was so high. Thanks for doing the legwork to track this down.

IvanMalison:
@bgamari I'm not convinced this is what is causing taffybar's high cpu usage.

As far as I can tell, taffybar only attempts interface updates when there is an actual change.

What led you to the conclusion that this was the cause?

bgamari:
@IvanMalison I observed a marked reduction in taffybar's CPU usage and wakeup count after applying this patch. Note that I still consider taffybar's CPU usage to be too high: I still see it saturate a good fraction of a CPU when switching desktops and a pretty minimal configuration still idles around 3%. However, this patch certainly helped and I'm afraid I don't have time to look farther at the moment.

IvanMalison:
>  Note that I still consider taffybar's CPU usage to be too high: I still see it saturate a good fraction of a CPU when switching desktops and a pretty minimal configuration

Yes, I agree, this is likely due to the fact that taffybar simply redraws the entire workspace widget on any ewmh update. It shouldn't be too hard to have it avoiding redraws for window icons when nothing has changed.

> minimal configuration still idles around 3%.

If I remove the clock and graph widgets i get 0% idle. I know you said that you don't have time to investigate, but could you try disabling those widgets and seeing if you see the same thing?

Continue discussion at https://github.com/taffybar/taffybar/issues/385 to avoid spamming xmonad-contrib

mgsloan:
LGTM :+1:

IvanMalison:
Is there a reason this hasn't been merged?

byorgey:
No reason, just not enough round tuits to go round, I guess...

PR #270, opened 2018-07-17
X.L.MultiToggle: Doc improvement

lucianposton:
Updated a doc that was missing some information that I would have found useful.

Minor doc change, so it doesn't update CHANGES.


byorgey:
Thanks for the update!

PR #274, opened 2018-07-21
Travis CI, and (dropping?) GHC 7.6/7.8 compatability

LSLeary:
### Description

Travis CI has been broken for way too long, so I had a go at it. Fair warning: I don't actually know what I'm doing, or what half the script is doing for that matter.

There's some statefulness in the setup that has contributed to random failures, so it's hard to know whether my changes have actually fixed that issue. It looked like the problem was `cabal configure [...]` failing with missing dependencies when the "cabal build-cache HIT" case of an earlier conditional meant that `cabal install --only-dependencies [...]` did not run, so I've moved that command out of the conditional. Presumably it shouldn't need to be run in that case, but it seems it does. I imagine there is a better fix; yell out if you know.

In the process of troubleshooting that, I saw what changes needed to be made for compatibility with earlier GHC versions, and made them in separate commits:
 - 7.8: Explicitly import applicative functions from `Control.Applicative`, and use the `DeriveDataTypeable` pragma. Nothing had to be sacrificed, so I see no reason to drop compatibility at this point.  
 - 7.6: Remove `PatternSynonym` usage from `X.L.StateFull`. That module has no particular need for it, but more generally I think pattern synonyms have potential to be quite useful (unlike GHC 7.6 compatibility). I'm also using them in another module I'm writing where their role is somewhat more substantial. __I'd prefer to drop this commit and remove 7.6 compat__.
 
### Checklist

    - Tested on my system.
    - Only the interface to `X.L.StateFull` has changed, so the changelog is only edited for correctness re that module.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 # This file has been generated -- see https://github.com/hvr/multi-ghc-travis
 language: c
 sudo: false
-dist: precise
 
 cache:
   directories:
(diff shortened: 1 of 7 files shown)


byorgey:
I think dropping GHC 7.6 support is totally fine!  If we support 7.8-8.4 that's 5 major versions which already seems above and beyond what's typical.

byorgey:
If you want to drop that commit then I'll take a closer look at the rest and merge it.

LSLeary:
OK @byorgey, I dropped the commit and added one explicitly removing 7.6 from xmonad-contrib.cabal and .travis.yml ... but it seems like the 7.8 version of PatternSynonyms is a bit too different from the modern one; it chokes on the type signature syntax. I took a look at the 7.8 manual and it seems like it might expect an empty context or two; something like this:
```haskell
pattern () => StateFull :: () => StateFull a
```
which isn't compatible with current PatternSynonyms. I don't want to use CPP so I've just removed the signature instead. Looks like it's building.

While you're here: I have another couple of modules that I want to get in master so they can be used, tested and further mature, but I wouldn't want them to go into a release if it were to happen within the next few months. What do you think about splitting off a 0.14-prerelease branch into which we don't merge all accepted PRs, and from which we remove any incomplete work?

PR #289, opened 2019-01-09
Improved tabbedRight and tabbedLeft in XMonad.Layout.Tabbed

berkeleyinc:
### Description

The leftTabbed and rightTabbed Layouts of XMonad.Layout.Tabbed are stretching their tabs over the whole screen height which makes the tab-text disappear.
This change will make the tabs have a fixed height (according to decoHeight). Similar to the TreeStyleTab Addon of Firefox.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Layout.Tabbed`
+
+    tabbedLeft and tabbedRight will set their tabs' height and width according to decoHeight/decoWidth
+
   * `XMonad.Prompt`
 
     Added `sorter` to `XPConfig` used to sort the possible completions by how
--- a/XMonad/Layout/Tabbed.hs
+++ b/XMonad/Layout/Tabbed.hs
@@ -14,7 +14,7 @@
 --
 -----------------------------------------------------------------------------
 
-module XMonad.Layout.Tabbed
+module Tabbed
     ( -- * Usage:
       -- $usage
       simpleTabbed, tabbed, addTabs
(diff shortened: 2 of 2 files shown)


byorgey:
Seems generally reasonable to me, though I don't use tabs and can't really follow the logic here.  Anyone else have a comment or able to try it out?  I'll merge in a few days if I don't hear anything.

LSLeary:
@byorgey @berkeleyinc The code for this module is somewhat opaque and my xmonad config is in the middle of a refactor so I can't easily test this atm, but I have a strong suspicion it should not be merged (at least not as-is).

The `Theme` used with a decoration allows the user to specify both a desired width and a height, but the implementation is free to use these values however it pleases. In particular, vertically and horizontally positioned decorations tend respectively to ignore width and height in favour of fitting to the window/screen dimensions instead. If I understand correctly, this PR is reversing that tendency for horizontally positioned tabs, however the existing behaviour is desirable and cannot be replicated by picking the right fixed height.

Ideally `Theme` would be refactored to make height and width fields optional so that the decoration can see whether or not the user wants a fixed width/height, but there might not be a good way to do that without breaking changes. It's probably easier to extend the tab decorations with an option and re-express the current interface within the extension.

berkeleyinc:
@LSLeary @byorgey 
To visualize the problem that I see, here is a before and an after picture:

- Example with current leftTabbed implementation: 
  <a href="https://ibb.co/JKqcHxb">xmonad-before</a>
- Example with changed leftTabbed implementation: 
  <a href="https://ibb.co/Kcn3Bfv">xmonad-after</a> 

In my opinion, the latter is preferred. Unless you (as a user) don't want tab titles at all. I don't see the use case for stretched leftTabbed / rightTabbed but if it should still be possible to have that, then how about we activate that stretched layout in case _decoHeight == 0_?

bergey:
This change looks good to me.  I've been using a very similar tab sizing of my own for 6+ years, and was just about to open a PR to upstream it.  Truncating window titles to 5 characters (when there's space for 30+) makes the layout useless to me.  Unless someone chimes in saying that they both use the current side tabs and like the current behavior, I think this should be merged, and defer broader redesigns of Theme or `updateDeco`.

I don't know why [updateDeco](https://github.com/xmonad/xmonad-contrib/blob/9fcea6cb55bceceec5487c07251afa32c7978315/XMonad/Layout/Decoration.hs#L403) fits the window title to (width - height / 2).  I don't know what other code uses `updateDeco`, or how hard it would be to change.

geekosaur:
I'd be wary of the various dwm compatibility modules, which were the original Decoration consumers. In particular, that code originated for dwm-style title bars.

bergey:
What's the status of this PR?  Would @berkeleyinc 's suggestion to

> activate that stretched layout in case decoHeight == 0?

be acceptable?

byorgey:
@bergey I don't know.  Honestly one of the problems here is that I'm more or less the only one who still occasionally looks at and merges xmonad PRs, and I literally have no idea what is going on with this one.  How about I merge it, give you a commit bit, and you promise to take a look if anyone complains about it in the future? =)

bergey:
@byorgey Yes, that works for me.  Thanks. =)

byorgey:
Done and done!

berkeleyinc:
Thanks a lot!!

Issue #217, opened 2017-11-19
XMonad.Prompt freezes if alwaysHighlight= True and maxComplRows is not Nothing


fabiodl:
### Problem Description

Especially for XMonad.Prompt.Prompt (but the bug holds for XMonad.Prompt.Window) I set alwaysHighlight to True (to avoid the necessity of pressing tab for selection) but to keep the prompt window reasonably small I set maxComplRows to Just 3
If I have enough choices, so that some of them are not shown, and I keep pressing tab, when I select the first "off screen" one, the prompt freezes. To be specific, the prompt stays there, other xmonad keybindings become available (for instance, switching workspace) while applications do not receive keystrokes anymore (the mouse can be used, however).
stdout reports

user error (Pattern match failure in do expression at XMonad/Prompt.hs:911:3-14)

Expected behavior: as when alwayHighlight is false, complete the prompt with the first invisible entry. 
### Configuration File

, ((myModKey .|. shiftMask, xK_b     ), windowPrompt def{maxComplRows=Just 3,alwaysHighlight = True} Bring allWindows)
### Checklist

  - I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

byorgey:
What version of xmonad are you using?

fabiodl:
0.13, installed from cabal

byorgey:
And when you run the version compiled from github, does it still give the same error `Pattern match failure in do expression at XMonad/Prompt.hs:911:3-14`?

fabiodl:
Sorry, for the git version it becomes row 914
user error (Pattern match failure in do expression at XMonad/Prompt.hs:914:3-14)

fabiodl:
I found out that when the error occurs, d seems not valid. Specifically, if I print d I get 
Display Prelude.!!: index too large. 
However, i have no idea where dpy is set

geekosaur:
https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Prompt.hs#L420 and later stuffed into the `XPState`. But I don't see why this would be relevant to `(!!)` unless there are two different uses of `d` with one shadowing the other. `dpy` is not meaningful to anything but Xlib and the X server.

fabiodl:
When the error does not occur, it contains what appears to be a pointer.

geekosaur:
[...]

PR #291, opened 2019-01-24
Fix XMonad.Prompt wraparound when maxComplRows not Nothing #217

mgsloan:
### Description

When tabbing through the prompt list of completions, if the list of completions does not fit due to `maxComplRows`, without this fix an exception would be thrown.  I am not certain why, but this exception occurs due to a pattern mismatch at https://github.com/mgsloan/xmonad-contrib/blob/dda242a4597c1ed0fa895a10b9296dcc709f4937/XMonad/Prompt.hs#L919

However, the root cause appears to be in the `nextComplIndex` function.  The problem is that it was assuming that dividing the total item count by the number of rows would yield the number of columns.  This did not hold when the number of rows was being limited.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,6 +35,10 @@
     Added `sorter` to `XPConfig` used to sort the possible completions by how
     well they match the search string (example: `XMonad.Prompt.FuzzyMatch`).
 
+    Fixes [issue #217](https://github.com/xmonad/xmonad-contrib/issues/217), where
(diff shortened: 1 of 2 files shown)


mgsloan:
Ping!  Any maintainer want to review this?  I've rebased it atop master.

byorgey:
Thanks for the PR and for the nudge!

mgsloan:
You're welcome, thanks for merging!

PR #292, opened 2019-02-13
Use absolute paths in DynamicProjects

wz1000:
### Description

This makes the input directory read from the prompt in `DynamicProjects` absolute wrt the current directory.

Without this patch, the directory set by the prompt is treated like a relative directory. This means that when you switch from a project with directory `foo` into a project with directory `bar`, xmonad actually
tries to `cd` into `foo/bar`, instead of `~/bar` as expected.

### Checklist


byorgey:
Thanks!

Issue #293, opened 2019-02-20
Enhancement: Cycle across windows on all visible workspaces


alhirzel:
### Enhancement

Provide a query to allow cycling across all visible windows (i.e. for multi-screen setups).

### Details

This was discussed in [this mailing list thread](https://mail.haskell.org/pipermail/xmonad/2017-February/015218.html) but did not seem to be resolved. The request is for a window switching method that allows for tabbing across all visible workspaces. I would like to submit the following `Query Bool` for use with `nextMatch` (from `XMonad.Actions.GroupNavigation`):

```haskell
isOnVisibleWS :: Query Bool
isOnVisibleWS = do
  w <- ask
  ws <- liftX $ gets windowset
  let allVisible = concat $ (maybe [] W.integrate) . W.stack . W.workspace <$> W.current ws:W.visible ws
      visibleWs = w `elem` allVisible
      unfocused = maybe True (w /=) $ W.peek ws
  return $ visibleWs && unfocused
```

Example usage:

```haskell
import XMonad.Actions.GroupNavigation (historyHook, nextMatch, Direction(..))
```

within the keys section

```
    -- use Alt+Tab and Alt+Shift+Tab to change focus to different windows across workspaces
    ((alt,           xK_Tab   ), nextMatch Forward isOnVisibleWS),
    ((alt .|. shift, xK_Tab   ), nextMatch Backward isOnVisibleWS),
```

mgsloan:
LGTM.

Side note is that it would be more efficient to compute `allVisibleWs` ahead of time.  Probably not entirely necessary since the number of workspaces is typically quite low, GHC may even be clever enough to inline and share the computation.

byorgey:
Looks good to me too.  Any ideas which module this should go in?  I guess we could just stick it in `XMonad.Actions.GroupNavigation` but it seems more generally applicable than that.  `XMonad.Hooks.ManageHelpers` is another place with several `Query Bool` things defined, though it doesn't seem super relevant.  In the end maybe `XMonad.Actions.GroupNavigation` is the best place for it, since that's the context in which folks are likely to want it.

@alhirzel , do you want to take a shot at opening a pull request that adds this?  We'd be happy to provide some guidance on how to do that if you need it.

alhirzel:
I took a swing at it in #294.

PR #294, opened 2019-02-28
Add isOnAnyVisibleWS function to Actions.GroupNavigation

alhirzel:
### Description

Adding `isOnAnyVisibleWS` to satisfy #293. The function itself has been tested as part of my `xmonad.hs`.

### Checklist

### DIFF
--- a/XMonad/Actions/GroupNavigation.hs
+++ b/XMonad/Actions/GroupNavigation.hs
@@ -142,7 +142,7 @@ orderedWorkspaceList ss wsids = rotateTo isCurWS wspcs'
     where
       wspcs      = SS.workspaces ss
       wspcsMap   = Fold.foldl' (\m ws -> Map.insert (SS.tag ws) ws m) Map.empty wspcs
-      wspcs'     = fmap (\wsid -> wspcsMap ! wsid) wsids
+      wspcs'     = fmap (wspcsMap !) wsids
       isCurWS ws = SS.tag ws == SS.tag (SS.workspace $ SS.current ss)
 
 --- History navigation, requires a layout modifier -------------------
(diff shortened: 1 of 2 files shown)


alhirzel:
(the force push was to correct a typo in a commit message)

XMonad/Actions/GroupNavigation.hs
```diff
@@ -16,17 +16,21 @@
...
 -- Also provides a method for jumping back to the most recently used
--- window in any given group.
+-- window in any given group, and predefined groups.
 --
 ----------------------------------------------------------------------
 
 module XMonad.Actions.GroupNavigation ( -- * Usage
                                         -- $usage
+
+                                        -- * Utilities
```
byorgey: Should this 'Utilities' heading be moved down right above the export of `isOnAnyVisibleWS`?  Right now `$utilities` will generate some documentation that says "handy queries for use with `nextMatch` etc." but it seems strange to put that at the top of a section containing e.g. `nextMatch` itself.

XMonad/Actions/GroupNavigation.hs
```diff
@@ -216,3 +220,22 @@ findM cond xs = findM' cond (viewl xs)
       if isMatch
         then return (Just x')
         else findM qry xs'
+
+
+-- $utilities
+-- #utilities#
```
byorgey: What is this `#utilities#` line for?

byorgey:
Looks great!  I just have a couple nitpicky questions/comments relating to the Haddock documentation.

XMonad/Actions/GroupNavigation.hs
alhirzel: I have never done haddock documentation, and followed an example. I assume it is for an HTML anchor but not sure; just followed the documentation of another `Actions.*` module.

alhirzel:
Yep, very welcome! I will address these and add the forgotten `import` when I get home tonight. Thank you for the feedback.

XMonad/Actions/GroupNavigation.hs
byorgey: Ah, I learned something new today!  You're right, it is to make an HTML anchor: https://www.haskell.org/haddock/doc/html/ch03s08.html#idm140354810720016

byorgey:
Thanks!

alhirzel:
@byorgey I forgot about the import problem; I just pushed the fix!

alhirzel:
#295

PR #295, opened 2019-03-03
Actions.GroupNavigation: fix bad import

alhirzel:
Follow-up to #294


alhirzel:
CI is happy now, sorry for the errors in the original PR!

byorgey:
Ah, good catch, thanks, I should have been paying more attention to CI as well.

PR #297, opened 2019-03-18
Hooks.WallpaperSetter: Preserve aspect ratio while scaling images

ajgrf:
### Description

Preserve the aspect ratio of wallpapers that xmonad sets. When previous versions would distort images to fit the screen size, it will now find a best fit by cropping instead.

I don't understand the imagemagick command invocation very well, so this was put together through trial and error. I've tested it with good results though, in a dual-head setup where each monitor has a different aspect ratio (1920x1080 and 1920x1200).

### Checklist


byorgey:
Thanks!

PR #302, opened 2019-05-08
Add filtering options to GridSelect

Geometer1729:
### Description

I wanted to be able to use GridSelect to chose between windows on the current workspace only.
So I added functions that accept a filter of type (Window -> X Bool) and only select between windows that return true.

### Checklist

I did test them by putting them in .xmonad/lib.

### DIFF
--- a/XMonad/Actions/GridSelect.hs
+++ b/XMonad/Actions/GridSelect.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE ScopedTypeVariables, GeneralizedNewtypeDeriving, TypeSynonymInstances, FlexibleInstances, OverlappingInstances #-}
+{-# LANGUAGE ScopedTypeVariables, GeneralizedNewtypeDeriving, TypeSynonymInstances, FlexibleInstances, MonoLocalBinds, IncoherentInstances, UndecidableInstances #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Actions.GridSelect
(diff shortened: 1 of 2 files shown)


XMonad/Actions/GridSelect.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE ScopedTypeVariables, GeneralizedNewtypeDeriving, TypeSynonymInstances, FlexibleInstances, OverlappingInstances #-}
```
byorgey: Why did the list of extensions change?  It's not obvious to me why the new functions necessitate these extensions.

XMonad/Actions/GridSelect.hs
```diff
@@ -38,6 +38,7 @@ module XMonad.Actions.GridSelect (
     withSelectedWindow,
     bringSelected,
     goToSelected,
+    goToSelectedFiltered,
```
byorgey: Did you mean to export more than just `goToSelectedFiltered`? The changelog message mentions `gridselectWindowFiltered` but that doesn't seem to be exported.

byorgey:
Thanks, looks useful!  I just have a couple minor questions/comments, see above.

XMonad/Actions/GridSelect.hs
Geometer1729: I think that is leftover from my first attempt to implement this feature, and I forgot to remove it. I will make a new pull request fixing that.

PR #305, opened 2019-06-09
Provide minimize layout that doesn't modify the layout name

IvanMalison:
### Description

behind them.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -89,6 +89,11 @@
     Add adwaitaTheme and adwaitaDarkTheme to match their respective
     GTK themes.
 
+  * `XMonad.Layout.Minimize`
+
+	Provide a version of the minimize layout modifier that does not modify the
+    layout name.
+
 
 ## 0.15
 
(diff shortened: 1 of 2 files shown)


geekosaur:
FWIW, this is what X.L.Renamed is for; we can't do this for *every* small layout modifier, therefore shouldn't do it for some privileged few.

IvanMalison:
> FWIW, this is what X.L.Renamed is for; we can't do this for _every_ small layout modifier, therefore shouldn't do it for some privileged few.

@geekosaur I already use renamed, and I still need this change, because I want to be able to name the individual layouts that I cycle through without having to rename everything.

Renamed is not great at picking out exactly how a few different layout modifiers applied/changed a name.

geekosaur:
`quietMinimize = renamed [CutWordsLeft 1] . minimize`

You don't have to apply it only at the top level, or etc.

IvanMalison:
> quietMinimize = renamed [CutWordsLeft 1] . minimize

Ah didn't know rename could do all of that.

PR #311, opened 2019-08-21
Added module X.A.PerWindowKeys

skewerr:
Allows one to shortcut actions that are performed based on queries.

### Description

Some XMonad users might want keybinds to perform different actions based on which window they are currently focusing. This adds two functions that allow them to perform either all actions for which a `Query Bool` holds true, or the first action for which that applies. One could also sequence the two for more complex behavior.

### Checklist

### DIFF
--- a/XMonad/Actions/PerWindowKeys.hs
+++ b/XMonad/Actions/PerWindowKeys.hs
@@ -1,0 +1,60 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Actions.PerWindowKeys
+-- Copyright   :  (c) Wilson Sales, 2019
+-- License     :  BSD3-style (see LICENSE)
+--
+-- Maintainer  :  Wilson Sales <spoonm@spoonm.org>
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- Define key-bindings on per-window basis.
+--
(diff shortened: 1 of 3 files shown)


XMonad/Actions/PerWindowKeys.hs
```diff
@@ -0,0 +1,60 @@
...
+-- you can use @'bindFirst' [(query1, action1), ..., (pure True,
+-- doThisIfTheOthersFail)]@.
+--
+-- For detailed instructions on editing your key bindings, see
+-- "XMonad.Doc.Extending#Editing_key_bindings".
+
+-- | Run an action if a Query holds true. Doesn't stop at the first one that
+-- does, however, and could potentially run all actions.
+bindAll :: [(Query Bool, X ())] -> X ()
+bindAll = sequence_ . map choose where
```
byorgey: Incidentally, `sequence_ . map choose` is the same as `mapM_ choose`.

byorgey:
This looks nice, thanks!  Can you also add the module in `xmonad-contrib.cabal`?

XMonad/Actions/PerWindowKeys.hs
skewerr: You're right, how did I miss that? Thanks for the heads up.

byorgey:
Thanks!

PR #315, opened 2019-09-27
A New Layout Combinator: TallMastersCombo

tidues:
### Description

This is a layout combinator that inherits the similar behaviors of the 'Tall' layout, and can be applied to merge any two sublayouts. Each sublayouts can be a 'Choose' composed layout, and the message FocusedNextLayout can switch to the next layout in the focused pane.
Moreover, it supports two master windows, a main master window, and a sub master window, which is the first window in the second pane. This combinator also works well with the Tabbed layout. Using the SwitchOrientation message, it can mirror the layout structures without affects the tabbed decorations. 

### Checklist

  - [ x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [x ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

  - [ x] I updated the `CHANGES.md` file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ tags
 # stack artifacts
 /.stack-work/
 /cabal.project.local
+
+stack.yaml.lock
+
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,14 @@
 
 ### New Modules
 
+  * `XMonad.Layout.TallMastersCombo`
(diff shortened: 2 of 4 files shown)


byorgey:
Looks great, thanks!

liskin:
This was merged post-0.16 but the CHANGES.md additions are in the 0.16 section. That's probably wrong.

PR #320, opened 2019-10-08
Remove legacy code regarding fmap

slotThe:
### Description

This removes a lot of legacy code involving `fmap`.

  1. Removes `Control.Applicative` imports for `(<$>), (<*>), (<$), pure`, and `Applicative` as these are in base now.
  2. Replaces all infix occurences of `fmap` and `liftM` with `<$>`.
  3. Replaces `liftM2` with `liftA2`.
  4. Replaces all remaining (prefix) occurences of `liftM` with `fmap` (I decided to not replace anything with `<$>` here, so as to preserve the original authors style).
  5. Applies the functor laws (in just one case) and replaces constructs of the form `fmap f $ fa` with `f <$> fa`.
  6. Replaces `join . fmap` (as well as things like `join $ io $ fmap`) with `=<<`.
  7. Cleans up all `Control.Monad` imports of `liftM` and `liftM2` (as those are no longer needed).

### Checklist

    Will try to get to this as soon as I can.

    I did not think this necessary.


byorgey:
Wow, thanks so much!

While skimming through the changes I noticed a few places with the keyword `do` followed by a single line of code.  It might be worth going through and getting rid of those too (not that I'm saying you have to do it!).

PR #321, opened 2019-10-09
Remove redundant do's

slotThe:
### Description

This pull request is part of an ongoing effort to remove legacy/redundant code from xmonad-contrib. 

  1. Removes all redundant `do`'s (places where `do` is followed by just a single line), as per @byorgey's request in #320. 

### Checklist

I did not think this necessary

### DIFF
--- a/XMonad/Actions/Launcher.hs
+++ b/XMonad/Actions/Launcher.hs
@@ -62,7 +62,7 @@ type ExtensionActions = M.Map String (String -> X())
 instance XPrompt CalculatorMode where
   showXPrompt CalcMode = "calc %s> "
   commandToComplete CalcMode = id --send the whole string to `calc`
-  completionFunction CalcMode = \s -> if (length s == 0) then return [] else do
+  completionFunction CalcMode = \s -> if (length s == 0) then return [] else
     lines <$> runProcessWithInput "calc" [s] ""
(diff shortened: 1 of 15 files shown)


geekosaur:
For what it's worth, I think the ones in the Debug… modules are stubs that I intended to fill in later, or otherwise just sticking to pattern so I could find all the debug operations more easily by pattern. There's more than just what looks pretty to think about here.

byorgey:
@geekosaur That's a fair point.  Are there particular modules where you'd prefer we didn't make this change?

geekosaur:
Just the Debug modules, since I more or less template the debug dumpers. At some point I'd like to expand and maybe bug-fix (although I thought I'd already sent some fixes in at one point) them, but it'd be no earlier than January and probably a few months after that.

geekosaur:
Also, aside from my caveat for which I might at some point be able to add pragmas to turn off the warnings, I'd be +1 on `hlint` for at least this repo. It tends to be more annoying than helpful for advanced users, so probably not the core; extras and testing might also do with it, though.

slotThe:
@geekosaur fair point indeed.  I have removed the changes for `DebugEvents` and `DebugWindow`.

byorgey:
Cool, thanks!

PR #324, opened 2019-10-29
New Feature for TallMastersCombo: Switch between Focused Windows across All Sub-layouts

tidues:
### Description

Each non-combinator sub-layout can contain one focused window, and this window will keep focused after switching to other sub-layouts. Two messages were added for switching between the focused windows across all sub-layouts.

### Checklist

### DIFF
--- a/XMonad/Layout/TallMastersCombo.hs
+++ b/XMonad/Layout/TallMastersCombo.hs
@@ -1,4 +1,11 @@
-{-# LANGUAGE PatternGuards, FlexibleContexts, FlexibleInstances, DeriveDataTypeable, TypeSynonymInstances, MultiParamTypeClasses #-}
+-- {-# LANGUAGE PatternGuards, FlexibleContexts, FlexibleInstances, DeriveDataTypeable, TypeSynonymInstances, MultiParamTypeClasses #-}
+{-# LANGUAGE PatternGuards,
+    FlexibleContexts, 
+    FlexibleInstances, 
+    DeriveDataTypeable, 
+    TypeSynonymInstances, 
+    MultiParamTypeClasses
+#-}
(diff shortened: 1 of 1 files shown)


byorgey:
Hi @tidues , does the `Choose` combinator continue to work?  I am not clear on whether this will break existing user configs or not.  Maybe `Choose` still works but if you want the behavior regarding focused windows you have to use the new `(|||)` operator?

geekosaur:
I'd also wonder how this interacts with the custom `(|||)` for the `JumpToLayout` messaage. Granting that it might be time to merge that one to core.

tidues:
Hi, @byorgey 

Currently, using this has to hide the (|||) operator from the XMonad module. Otherwise, there will be an ambiguity error. Other than this, everything else of the Choose combinator still works.

Basically, my (|||) operator is a wrapper for the original one with one extra feature that can monitor the current states of the Choose combinator. They serve for almost the same functionality, so I decided to use the same operator.

Actually, if the XMonad.Layout and expose the Choose(...) and LR(...) constructors, then I don't have to use a wrapper to monitor its states.

tidues:
Hi @geekosaur, 
I am not familiar with the JumpToLayout message. But every message that is not handled by the Choose combinator will pass down to the currently active layout. I don't if this will pose a problem?

geekosaur:
It's handled by a replacement `(|||)` operator, is the point. So now we have two different and apparently incompatible `(|||)` alternatives?

tidues:
I see. The JumpToLayout message and the corresponding (|||) operator are provided by the LayoutCombinator module. I know your concerns, but I don't think it is a problem since both are served as flexible layout combinators. Users will usually go with one or the other.

byorgey:
Thanks!

PR #325, opened 2019-11-18
Prompt: Use searchPredicate in more cases

jcpetruzza:
### Description

Prompts based on `mkComplFunList` and `mkComplFunList'` were not taking into account the `searchPredicate` funtion from `XPConfig`. This was rather confusing.

We fix it by passing `XPConfig` to these functions; although this is strictly more than they need, it makes the breaking change very easy to fix and is also more future-proof.

### Checklist

    I'm using the amended `workspacePrompt` on my own config. The rest of the changes were type-directed.
   
### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,12 @@
 
 ### Breaking Changes
 
+  * `XMonad.Prompt`
+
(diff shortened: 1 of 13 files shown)


byorgey:
Makes sense, thanks!

PR #328, opened 2019-12-02
X.L.MouseResizableTile borderWidth via focused win

atondwal:
When we calculate draggerGeometry, we should use the current width of
the border (if available), rather than the initial borderWidth.

Otherwise, we get the wrong width if the user has changed the borderWidth since XMonad started.


byorgey:
Thanks!

PR #329, opened 2019-12-15
Use `drop 1` instead of reimplementing it in terms of `tail`

josephcsible:
### Description

`tail` is a partial function and should be avoided where practical. In this case, it's being used to create a function exactly equivalent to `drop 1`, so just remove that function and use that instead.

### Checklist


byorgey:
Thanks!

PR #334, opened 2020-04-24
Add new variants of spawnPipe functions with encoding support

psibi:
The current handle returned by `spawnPipe` doesn't have any encoding
and it uses the function fdToHandle which returns a Binary
Handle. When spawnPipe is used with a program like xmobar, this can
easily lead to errors:

```
λ> h <- spawnPipe "xmobar"
λ> hPutStrLn h
"\35753Haskell\25110\32773Ghci\33021\27491\30830\26174\31034\27721\23383\24182\19988\35835\21462\27721\23383\21629\21517\30340\25991\26723"
<stdin>: hGetLine: invalid argument (invalid byte sequence)
```

One workaround is to use `hSetEncoding`. But from
reading GHC's source code - the entire Handle and write operations
seems stateful. So doing something like hPutStr and hSetEncoding can
theoretically lead to an undefined state as the first write will use
latin encoding and the second write will use the new encoding. More
details about it are present here:

* http://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.IO.Handle.Internals.html#writeCharBuffer
* http://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.IO.Buffer.html#CharBuffer

So having these new functions will ensure that we get the handles in
the proper encoding in the first place.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,12 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Util.Run` 
+  
+     Added two new functions to the module: `spawnPipeWithLocaleEncoding` and
+     `spawnPipeWithUtf8Encoding`. Using these function should be
+     preferred over `spawnPipe`.
+
   * `XMonad.Prompt.Window`
 
     Added 'allApplications' function which maps application executable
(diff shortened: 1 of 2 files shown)


byorgey:
No objections here.

PR #335, opened 2020-04-30
Add a doSink manage hook

sam-barr:
### Description

Added a new manage hook `doSink`, which does the opposite of `doFloat`.
It seemed like a gap in the available configuration functions that was easy to fill.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -513,7 +513,8 @@
 
   * `XMonad.Hooks.ManageHelpers`
 
-    Make type of ManageHook combinators more general.
+    - Make type of ManageHook combinators more general.
+    - New manage hook `doSink` for sinking windows (as upposed to the `doFloat` manage hook)
 
   * `XMonad.Prompt`
 
--- a/XMonad/Hooks/ManageHelpers.hs
+++ b/XMonad/Hooks/ManageHelpers.hs
@@ -226,3 +226,7 @@ doCenterFloat = doSideFloat C
 -- | Hides window and ignores it.
 doHideIgnore :: ManageHook
 doHideIgnore = ask >>= \w -> liftX (hide w) >> doF (W.delete w)
+
+-- | Sinks a window
+doSink :: ManageHook
+doSink = reader (Endo . sink)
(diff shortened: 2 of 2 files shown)


geekosaur:
The main problem with doSink is that autosink happens after the user manageHook so it tends to not work well.

sam-barr:
What is autosink? Searching for "xmonad" "autosink" returns literally zero results, and it does sound like something that would help with the problem I solved with `doSink`

geekosaur:
Sorry, I meant autofloat; I was trying to reply on my phone, with various problems along the way, and got a bit flustered.

If a window is marked as a dialog (ICCCM style, not EWMH; the latter requires an appropriate manageHook, found in X.H.ManageHelpers) or has a fixed size, xmonad floats it after running the manageHook. The action is appended to the manageHook, so it can't be tested for reasonably and overriding it can be tricky.

sam-barr:
I see, I guess that explains why `doSink` as a manage hook seems to work, but `isFloating --> doSink` doesn't.

byorgey:
So should we add some comments explaining a caveat about when this works and when it doesn't? (I confess I don't really understand the issue.)  I don't have a problem with adding this as is.

geekosaur:
Sounds OK to me. Although if you didn't understand the problem then
perhaps someone other than me should compose the caveat. :)

(The caveat is that an auto-floated window, one which has the same
size in all three ICCCM size hints, can't reliably be auto-sunk
because the float happens before running the ManageHook. This may
either fail or lead to "flickering" and possibly a size change if it's
later floated again. I might need to double-check that we don't also
check the WM_TRANSIENT_FOR property to float it; I'm somewhat badly
placed to do that currently.)

On 5/9/20, Brent Yorgey <notifications@github.com> wrote:
> So should we add some comments explaining a caveat about when this works and
> when it doesn't? (I confess I don't really understand the issue.)  I don't
> have a problem with adding this as is.
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/xmonad/xmonad-contrib/pull/335#issuecomment-626229609


-- 
brandon s allbery kf8nh
allbery.b@gmail.com

byorgey:
I'm merging this since it seems harmless and potentially useful even given the caveats, but it would be nice if the documentation explains the limitations.  @sam-barr , would you like to make another pull request adding a bit more documentation explaining when this will or won't work?

PR #339, opened 2020-05-14
Add ResizableThreeColumns layout

samtay:
Adds another layout.

### Description

I've added `XMonad.Layout.ResizableThreeColumns`, which is the same layout as `ThreeColumns` but allows the slave window heights to be resized, just as was implemented for the `ResizableTall` layout.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,11 @@
 
 ### New Modules
 
+  * `XMonad.Layout.ResizableThreeColumns`
+
+    A layout based on 'XMonad.Layout.ThreeColumns' but with each slave window's
+    height resizable.
+
   * `XMonad.Layout.TallMastersCombo`
 
     A layout combinator that support Shrink, Expand, and IncMasterN just as
--- a/XMonad/Doc/Extending.hs
+++ b/XMonad/Doc/Extending.hs
@@ -899,6 +899,10 @@ For more information on using those modules for customizing your
     More useful tiled layout that allows you to change a width\/height of window.
     See also "XMonad.Layout.MouseResizableTile".
 
+* "XMonad.Layout.ResizableThreeColumns":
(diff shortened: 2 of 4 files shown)


liskin:
This seems to be a really nice mesh of ThreeColumns and ResizableTall. I'm looking forward to using this if I ever get back to the office. :+1:

Just minor suggestion:

* you're adding 2 entries to XMonad/Doc/Extending.hs and no other module is documented twice in that file, so maybe that isn't a good idea

It would be also nice if, in addition to being able to resize slave windows vertically and increase/decrease number of windows in master column, one could also resize the slave column with currently focused window, and even increase/decrease number of windows in slave columns. This is an extra feature not present in either ThreeColumn or ResizableTall, because the former isn't resizable and the latter isn't three column, but it's quite natural, I think. :-)
(But it's probably okay to merge it even without it...)

samtay:
Oh thanks for the heads up! That duplication was totally an accident, probably should have looked over these changes more carefully :blush: should be fixed now.
 
Totally agree with your suggestion, I have had that exact same thought when using ThreeCol, it sometimes feels very limiting to only be able to inc/dec the master column count. This is the first time I've explored the underlying XMonad code, so I don't immediately know how to access like which window is currently focused to detect the current column, etc., but when I find more time I'll look into adding this, and if I do I'll definitely push it back up here.

byorgey:
So @liskin do you think this is now ready for merge then?

liskin:
@byorgey I was hoping to test it for a while, but it's going to be some time till I get to the office, so... maybe we should just merge it and, if any problems appear, fix it afterwards? The code looks really okay at first glance and I'm not sure I'll find the time and motivation to review it thoroughly until I can actually make use of it. :-)

byorgey:
You convinced me =)

PR #340, opened 2020-05-17
Add maxMagnifierOff to Magnifier layout

igor04:
### Description

It's useful to wrap layouts with `magnifierOff` and use Magnifier in cases when you needed to zoom window, but when you zoom window it makes sense to show it in max size without using `MagnifyMore` which is layout specific. We have `maximizeVertical` which work in a similar way but it uses only vertical space and ignores horizontal

### Checklist

### DIFF
--- a/XMonad/Layout/Magnifier.hs
+++ b/XMonad/Layout/Magnifier.hs
@@ -23,6 +23,7 @@ module XMonad.Layout.Magnifier
       magnifier,
       magnifier',
       magnifierOff,
+      maxMagnifierOff,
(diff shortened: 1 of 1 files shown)


duog:
Just a driveby comment: let's export the `Mag` constructor  so users can write novel combinators like this themselves. I've been frustrated by this before.

byorgey:
Thanks!

byorgey:
@duog , that's a great idea.  Would you be willing/able to open a pull request making that change?

Issue #338, opened 2020-05-10
Proposal: Migrate the usage of spawnPipe function to spawnPipeWithLocaleEncoding


psibi:
I recently merged a [MR](https://github.com/xmonad/xmonad-contrib/pull/334/) which introduced two functions:
* spawnPipeWithLocaleEncoding
* spawnPipeWithUtf8Encoding

As compared to the current `spawnPipe` function which provides a binary handle, these two function provides handle with proper encoding support. 

When a handle doesn't have any encoding associated with it, it seems to use latin encoding to write the data: http://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.IO.Handle.Internals.html#writeCharBuffer

Also, we can easily crash the programs created via `spawnPipe` functions:
```
λ> h <- spawnPipe "xmobar"
λ> hPutStrLn h
"\35753Haskell\25110\32773Ghci\33021\27491\30830\26174\31034\27721\23383\24182\19988\35835\21462\27721\23383\21629\21517\30340\25991\26723"
<stdin>: hGetLine: invalid argument (invalid byte sequence)
```

My proposal here is to convert the `xmobar-contrib's` usage of `spawnPipe` function to `spawnPipeWithLocaleEncoding` so that system's locale encoding is respected. An example function where it needs to be converted is the `statusBar` function:
https://www.stackage.org/haddock/lts-15.11/xmonad-contrib-0.16/src/XMonad.Hooks.DynamicLog.html#statusBar

I can volunteer to implement his, but I would like to hear input from others before I carry out he implementation. 

CC: @byorgey @geekosaur

geekosaur:
Makes sense to me. But there's a lot of user configs that use
spawnPipe directly; I wonder if we want to alias that to
spawnPipewithLocaleEncoding and create a spawnPipewithNoEncoding, and
leave it to users to convert as needed.

-- 
brandon s allbery kf8nh
allbery.b@gmail.com

psibi:
That sounds like a good idea. Infact, with that the implementation work would be easier. :-)

liskin:
This unfortunately breaks many configs that use `spawnPipe` and `dynamicLogString` together, because `dynamicLogString` unfortunately expects a binary handle and encodes its output as utf8, so this then results in double-utf8-encoded output.

[...]

PR #346, opened 2020-05-29
Make spawnPipe to use system's locale encoding

psibi:
Add a function named spawnPipeWithNoEncoding for people who might want
to use binary handles.

Fixes https://github.com/xmonad/xmonad-contrib/issues/338

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,9 +46,13 @@
 
   * `XMonad.Util.Run` 
   
-     Added two new functions to the module: `spawnPipeWithLocaleEncoding` and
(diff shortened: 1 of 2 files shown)


byorgey:
Makes sense to me.

liskin:
This unfortunately breaks many configs that use spawnPipe and dynamicLogString together, because dynamicLogString unfortunately expects a binary handle and encodes its output as utf8, so this then results in double-utf8-encoded output.

Assuming most people use spawnPipe together with dynamicLogString and using it separately is uncommon, I propose to revert this PR and improve the documentation instead. @psibi @geekosaur any other ideas?

Issue #348, opened 2020-05-31
Layout broken on Arch Linux by missing Xorg misc font


sergeykish:
### Problem Description

Initial investigation in https://github.com/xmonad/xmonad/issues/225

I follow https://hackage.haskell.org/package/xmonad-contrib-0.13/docs/XMonad-Layout-Tabbed.html

ER: it shows tabs
AR: one fullscreen window

Fix - set `fontName` (even on non existent).

Toggling between with and without `fontName`:

Mod-q (reload) does not affect
Shift-Mod-q (quit), `startx` affects

```
$ startx 2> errors
user error (createFontSet)
user error (createFontSet)
user error (createFontSet)
user error (createFontSet)
user error (Error in function getWindowAttributes)
user error (createFontSet)
user error (Error in function getWindowAttributes)
user error (createFontSet)
user error (Error in function getWindowAttributes)
```

Caused by default `fontName` https://github.com/xmonad/xmonad-contrib/blob/a58c1a6071ceaf83f284b5f641c949818c216318/XMonad/Layout/Decoration.hs#L106

Arch Linux does not recommend installing `xorg-fonts-misc` maybe caused by

> Note: pango 1.44 dropped support for FreeType in favor of HarfBuzz thus losing support for traditional BDF/PCF bitmap fonts

https://wiki.archlinux.org/index.php/Fonts#Bitmap

### Configuration File

```haskell
import XMonad
import XMonad.Layout.Tabbed

myTabConfig = def { inactiveBorderColor = "#FF0000"
                  -- , fontName = "xft:foo"
                  , activeTextColor = "#00FF00"}
myLayout = tabbed shrinkText myTabConfig

main = do
    xmonad $ def
        { layoutHook  = myLayout }
```

## Not limited to Tabs Layout

Same problem with https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/XMonad-Layout-DwmStyle.html

ER: DWM style - like Tall with name in top right corner
AR: fullscreen window

```haskell
import XMonad
import XMonad.Layout.DwmStyle

myDWConfig = def { inactiveBorderColor = "red"
                 --, fontName = "xft:foo"
                 , inactiveTextColor   = "red"}

myL = dwmStyle shrinkText myDWConfig (layoutHook def)
main = xmonad def { layoutHook = myL }
```

### Checklist

[...]

PR #349, opened 2020-05-31
Permissive font pattern by default (fix broken layout)

sergeykish:
Layout broken if font is not found with `user error (createFontSet)`

Pango 1.44 dropped support for FreeType in favor of HarfBuzz thus losing support for traditional BDF/PCF bitmap fonts. There would be no `xorg-fonts-misc` on future installations.

Fixes https://github.com/xmonad/xmonad-contrib/issues/348

### Description

Change default `fontName` to accept any font.  Previous default expects XOrg misc font (PCF) which is not supported in pango 1.44 anymore and would not be present in future installations.

Font not found causes silent `user error (createFontSet)` and breaks Layout.

### Checklist

I've tested with latest `git` versions of `xmonad` and `xmonad-contrib`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,16 @@
 
 ### Breaking Changes
 
+  * All modules that export bitmap fonts as their default
+
+    - If xmonad is compiled with XFT support (the default), use an XFT
+      font instead.  The previous default expected an X11 misc font
+      (PCF), which is not supported in pango 1.44 anymore and thus some
+      distributions have stopped shipping these.
+
+      This fixes the silent `user error (createFontSet)`; this would
+      break the respective modules.
+
   * `XMonad.Prompt`
 
     - Now `mkComplFunFromList` and `mkComplFunFromList'` take an
--- a/XMonad/Actions/ShowText.hs
+++ b/XMonad/Actions/ShowText.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE CPP                #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Actions.ShowText
@@ -74,7 +75,11 @@ data ShowTextConfig =
 
 instance Default ShowTextConfig where
   def =
+#ifdef XFT
+    STC { st_font = "xft:monospace-20"
+#else
     STC { st_font = "-misc-fixed-*-*-*-*-20-*-*-*-*-*-*-*"
+#endif
         , st_bg   = "black"
         , st_fg   = "white"
     }
--- a/XMonad/Layout/Decoration.hs
+++ b/XMonad/Layout/Decoration.hs
@@ -1,4 +1,10 @@
-{-# LANGUAGE DeriveDataTypeable, FlexibleContexts, FlexibleInstances, MultiParamTypeClasses, PatternGuards, TypeSynonymInstances #-}
+{-# LANGUAGE DeriveDataTypeable    #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE PatternGuards         #-}
+{-# LANGUAGE TypeSynonymInstances  #-}
+{-# LANGUAGE CPP                   #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Layout.Decoration
@@ -104,7 +110,11 @@ instance Default Theme where
           , activeTextColor     = "#FFFFFF"
           , inactiveTextColor   = "#BFBFBF"
           , urgentTextColor     = "#FF0000"
+#ifdef XFT
+          , fontName            = "xft:monospace"
+#else
           , fontName            = "-misc-fixed-*-*-*-*-10-*-*-*-*-*-*-*"
+#endif
           , decoWidth           = 200
           , decoHeight          = 20
           , windowTitleAddons   = []
--- a/XMonad/Layout/ShowWName.hs
+++ b/XMonad/Layout/ShowWName.hs
@@ -1,4 +1,7 @@
-{-# LANGUAGE PatternGuards, FlexibleInstances, MultiParamTypeClasses #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE PatternGuards         #-}
+{-# LANGUAGE CPP                   #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Layout.ShowWName
@@ -62,7 +65,11 @@ data SWNConfig =
 
 instance Default SWNConfig where
   def =
+#ifdef XFT
+    SWNC { swn_font    = "xft:monospace-20"
+#else
     SWNC { swn_font    = "-misc-fixed-*-*-*-*-20-*-*-*-*-*-*-*"
+#endif
          , swn_bgcolor = "black"
          , swn_color   = "white"
          , swn_fade    = 1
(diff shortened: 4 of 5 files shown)


liskin:
So if future distros are going to not ship misc fixed PCF fonts, why do we expect them to ship any PCF fonts at all? Wouldn't a better fix be an `#ifdef` that uses something like `xft:monospace` when xft is enabled?

geekosaur:
That's what I suggested in the original bug report.

sergeykish:
Sorry, I'm not qualified to answer. I do not yet understand how various font systems work. What does switch to `harfbuzz` mean, does it support [XLFD](https://wiki.archlinux.org/index.php/X_Logical_Font_Description)?

There was no difference when I removed `fuse_xft`. [It should be](https://github.com/xmonad/xmonad-contrib/blob/c3bb1cb2e76de8036e77926eb03ec1b5472e6dd4/XMonad/Util/Font.hs#L115)

I've found [another fallback](https://github.com/xmonad/xmonad-contrib/blob/c3bb1cb2e76de8036e77926eb03ec1b5472e6dd4/XMonad/Util/Font.hs#L103)

I've spent a day trying to understand what is going on. Current results - X Server contains built in font (kind of) which is selected by Xmonad config. This font is not present in the system. Everything works once I install package containing font. So it should be dependency of X Server on Arch Linux.

But this font behaves strange - with `xset fp /usr/share/font/TTF` X Server produces much better error - both with `xfontsel` and `xmonad`:

```
X Error of failed request:  BadName (named color or font does not exist)
  Major opcode of failed request:  45 (X_OpenFont)
  Serial number of failed request:  10
  Current serial number in output stream:  11
```

And xmonad pulls this buggy font even after I've installed terminus font (does not match foundry).

I believe each project may be improved a bit. Filling issues in XOrg and Arch Linux trackers.

[my test case in shell](https://gist.github.com/sergeykish/744300bbf3befa8bbe43759f44223e7f)

liskin:
> Sorry, I'm not qualified to answer. I do not yet understand how various font systems work. What does switch to `harfbuzz` mean, does it support [XLFD](https://wiki.archlinux.org/index.php/X_Logical_Font_Description)?

I don't think it does.

That being said, it's irrelevant, anyway. xmonad didn't switch to harfbuzz, pango did, and xmonad doesn't use pango. xmonad uses XLFD and optionally Xft. Both support bitmap fonts and that support isn't going away. So the only relevant question is still this: are current/future distros shipping bitmap fonts or not? If not, then defaulting to an XLFD font descriptor, no matter how permissive (wildcard-y), doesn't make sense, as going through XLFD means no antialiasing, and antialiasing is necessary for vector fonts.

If not depending on the core X11 fonts is a packaging bug in Arch that's been fixed since, I think we can close this bug. If, on the other hand, modern systems only install TTF fonts, then we need to default to something like `xft:monospace` as `-*-*-*-*-*-*-*-*-*-*-*-*-*-*` will either result in ugly rendering or fail to select any fonts at all (`xlsfonts` on my system doesn't list any TTF fonts at all).

slotThe:
> If not depending on the core X11 fonts is a packaging bug in Arch that's been fixed since, I think we can close this bug. If, on the other hand, modern systems only install TTF fonts, then we need to default to something like `xft:monospace` as `-*-*-*-*-*-*-*-*-*-*-*-*-*-*` will either result in ugly rendering or fail to select any fonts at all (`xlsfonts` on my system doesn't list any TTF fonts at all).

We have had a few people on arch reporting that the prompt doesn't come up.  This seems to always invariably be a font issue (the default font for the prompt is also set to `-misc-fixed-*-*-*-*-12-*-*-*-*-*-*-*`).  It seems like a pretty safe bet that bitmap font support will wane in other distros as well sooner or later.  I would be in favour of switching every occurance of this kind of default font to something like `xft:monospace`

geekosaur:
You'll want to make sure the `use_xft` flag defaults to enabled.

slotThe:
Good point.

@sergeykish are you still interested in this?  The `xft:monospace` solution seems to be the most future proof at this point; you'd also need to fix this in all other modules that use some form of default font (like `XMonad.Prompt`)

geekosaur:
I would actually recommend that the Cabal flag define a CPP macro that can be used to select an appropriate font expression depending on whether `use_xft` is enabled or not.

I would also beware of the all-`*` XLFD spec, because X11 will use the first font it happens to find in its enumeration which may well be unreadable. A more appropriate default would be something like `-*-*-medium-r-normal-*-*-*-*-*-*-*-iso8859-*` which is imperfect (it may select a very small or very large font) but better than e.g. dingbats or the CJK-encoded font I got with all-`*` in local `xfontsel`. (Sadly, there's no way to say the second last field should be either `iso8859` or `iso10646`.)

Even better would be a way to do multiple fallbacks, but that again requires `use_xft` and I'm not sure how well the fallback syntax works when mixing XLFD and XFT fonts. And requires other support in the code IIRC, and only some contribs support it.

slotThe:
@liskin @geekosaur I pushed a few changes; what do you people think?  This is something I'd like to get done before 0.17

geekosaur (approved):
Looks good to me.

liskin:
@slotThe Why are we changing the non-XFT default, and why to `-*-*-medium-r-normal-*-*-*-*-*-*-*-iso8859-*` in particular? Especially the iso8859 bit seems a bit weird these days. :-)

geekosaur:
See the earlier discussion. 10646 would also work, the real question is which set might be installed.

slotThe:
On Fri, Mar 19 2021 14:36, Tomáš Janoušek wrote:
> @slotThe Why are we changing the XLFD default, and why to
> `-*-*-medium-r-normal-*-*-*-*-*-*-*-iso8859-*` in particular?
> Especially the iso8859 bit seems a bit weird these days. :-)

As geekosaur already said, I just went with what was said above.  I
don't really have a strong opinion on this, and I don't suspect it
matters very much, because we compile with XFT support by default and
hence I don't think these fonts will see much use anyways.

liskin:
> As geekosaur already said, I just went with what was said above. I don't really have a strong opinion on this, and I don't suspect it matters very much, because we compile with XFT support by default and hence I don't think these fonts will see much use anyways.

Well, it'd break my setup if I hadn't enlarged the fonts last week, so I wouldn't be that sure about it not mattering. :-)

The thing is, @geekosaur's message doesn't really explain the change. It explains why `-*-*-medium-r-normal-*-*-*-*-*-*-*-iso8859-*` is better than `-*-…-*-`, which I'm in no way trying to dispute, but it doesn't explain why any change to the XLFD default is necessary. To answer that question, we'd need to know if there are distros that don't ship misc fixed fonts but still ship other _useful_ bitmap fonts addressable using XLFD, and I believe/hope the answer to that is no, so as long as we default to xft when available, the sensible `-misc-fixed-…` XLFD default can stay and we gain absolutely nothing by changing it to something that allows more fonts in weird broken setups but selects wrong fonts in the traditional non-broken setups.

geekosaur:
I only came into it after the change to `-*-(...)-*` had been made, so I can't speak to that one. I would suspect however that there willl be two groups: those that move completely to XFT, and those who keep the older setup including the (formerly?) required `-misc-fixed` font.

slotThe:
> I would suspect however that there willl be two groups: those that move completely to XFT, and those who keep the older setup including the (formerly?) required `-misc-fixed` font.

That is a fair point.  I reverted back to `misc-fixed`-style fonts for the non-XFT users and just changed the XFT fonts.

liskin (approved):
Perfect.

(We use `#ifdef` in the rest of the codebase, but I'm not sure if that's worth following or whether `#if` is fine. Feel free to merge if you know what you're doing. :-))

slotThe:
Ah, you're right of course.  In this case it doesn't really matter because if `XFT` is defined then it's also `1` and so `#if` and `#ifdef` do the same, but I think it's worth following what we do elsewhere.

PR #350, opened 2020-06-01
Add Ebay Function to X.A.Search

davama:
Add ebay function

Was looking over my xmonad.hs and found this. Been using for quite some time. 
Thought it could be useful to others.

Tested on my own environment.

If ok, ill update CHANGES file

Thank you for the continued support!

Thanks,
Dave


byorgey:
Thanks!

PR #351, opened 2020-06-03
Make XMonad.Util.NamedScratchpad more flexible

elkowar:
### Description

This PR adds more customizability to how scratchpads are initially started in `XMonad.Util.NamedScratchpad`, by adding two new exported functions:
- `customRunNamedScratchpadAction`
  (provides the option to customize the `X ()` action the scratchpad is launched by) 
- `spawnHereNamedScratchpadAction`
  (uses `XMonad.Actions.SpawnOn.spawnHere` to initially start the scratchpad on the workspace it was launched on)

### Checklist

### DIFF
--- a/XMonad/Util/NamedScratchpad.hs
+++ b/XMonad/Util/NamedScratchpad.hs
@@ -36,7 +36,9 @@ import XMonad.Actions.DynamicWorkspaces (addHiddenWorkspace)
 import XMonad.Hooks.DynamicLog (PP, ppSort)
 import XMonad.Actions.SpawnOn (spawnHere)
 
-import Control.Monad (filterM)
+import qualified Data.List.NonEmpty as NE
+
+import Control.Monad (filterM, unless)
 import Data.Maybe (listToMaybe)
 
 import qualified XMonad.StackSet as W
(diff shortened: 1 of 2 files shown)


XMonad/Util/NamedScratchpad.hs
```diff
@@ -115,30 +118,48 @@ findByName c s = listToMaybe $ filter ((s==) . name) c
...
+                               -> String                     -- ^ Scratchpad name
+                               -> X ()
+spawnHereNamedScratchpadAction = customRunNamedScratchpadAction runApplicationHere
+
+-- | Action to pop up specified named scratchpad, given a custom way to initially start the application.
+customRunNamedScratchpadAction :: (NamedScratchpad -> X ())  -- ^ Function initially running the application, given the configured @scratchpad@ cmd
+                               -> NamedScratchpads           -- ^ Named scratchpads configuration
+                               -> String                     -- ^ Scratchpad name
+                               -> X ()
+customRunNamedScratchpadAction = someNamedScratchpadAction (\f ws -> f $ head ws)
```
byorgey, line 142: How do we know this use of `head` is safe?  Can this code be rewritten without using it, in a way that is more "obviously" safe?

byorgey:
This looks good to me, other than one dodgy-looking use of `head` that I'm wondering about.

elkowar:
I've refactored the module to now use `Data.NonEmpty`, which makes it obvious that this is now, in fact, safe. I've also refactored the `someNamedScratchpadAction` a bit in the process, to (in my opinion) improve the readability quite noticably. 
If you disagree with the code-style choices I've made here, i could obviously revert this to be closer to the original style.

I think making variable names a bit less abbreviated is a good idea for modules like these, because these are the modules that people like me start looking in to as examples on how to implement their own XMonad modules and features, so having the names be more obvious to people not that deep into the XMonad data structures makes it easier to start implementing custom functionality.

byorgey:
Looks great to me, thanks!  Sorry it's taken me a while to get around to it again.

PR #357, opened 2020-06-20
Add XMonad.Hooks.WindowSwallowing

elkowar:
### Description

This PR adds a handleEventHook that implements DWM style window swallowing.
It allows you to configure certain applications like terminals, file explorers, etc, to be "swallowed" when opening another window from within them.
Once the child window closes, the parent is restored again.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,7 +22,14 @@
 
 ### New Modules
 
+  * `XMonad.Hooks.WindowSwallowing`
+
+    A handleEventHook that implements window swallowing:
+    Hide parent windows like terminals when opening other programs (like image viewers) from within them,
+    restoring them once the child application closes.
+
   * `XMonad.Actions.TiledWindowDragging`
+
     An action that allows you to change the position of windows by dragging them around.
 
   * `XMonad.Layout.ResizableThreeColumns`
--- a/XMonad/Doc/Extending.hs
+++ b/XMonad/Doc/Extending.hs
@@ -572,6 +572,11 @@ Here is a list of the modules found in @XMonad.Hooks@:
 * "XMonad.Hooks.WorkspaceHistory":
     Keeps track of workspace viewing order.
 
+* "XMonad.Hooks.WindowSwallowing"
+    A handleEventHook that implements window swallowing:
+    Hide parent windows like terminals when opening other programs (like image viewers) from within them,
+    restoring them once the child application closes.
+
 * "XMonad.Hooks.XPropManage":
     A ManageHook matching on XProperties.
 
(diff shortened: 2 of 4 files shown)


elkowar:
there are definitely some parts in this that i'm not 100% happy with, where i had to choose the least bad of the available options. One example is the dependance on `pstree`. This should not be a problem in practice, as `pstree` should be available everywhere xmonad can run, but it's still not beautiful. If you have any ideas on how i could avoid that, i'd be more than happy to adjust the code.

Also, the amount of times where this does _not_ work for reasons such as applications not providing `_NET_WM_PID` or sharing process instances is a bit sad, but not really something that we can avoid. As far as i'm able to tell, there is no other way to get the PID of a window, and there is no way to check process hierarchy without knowing the PID.

Checking the process hierarchy is pretty much a must-have, because otherwise just opening any application using a keyboard shortcut or an application launcher would cause the window focused before to be swallowed.

slotThe:
> One example is the dependance on `pstree`. This should not be a problem in practice, as `pstree` should be available everywhere xmonad can run, but it's still not beautiful. If you have any ideas on how i could avoid that, i'd be more than happy to adjust the code.

Maybe I have not understood this well enough, but it seems to me that you should be able to query `/proc/PID/stat` for this information, no?

geekosaur:
pstree is available on more platforms than Linux /proc. Do remember
xmonad originated on netbsd, not Linux.

-- 
brandon s allbery kf8nh
allbery.b@gmail.com

elkowar:
should i try to add parsing `/proc/PID/stat` as a fallback maybe?

byorgey:
I don't know much about the issues involved, but I don't think there's any harm in adding this as a new standalone module as long as the caveats are well-documented (which they certainly seem to be).  If you'd like to add parsing `/proc/PID/stat` as a fallback you can, but I think we could also merge this as-is and add that later.

slotThe:
On Mon, Jun 22 2020 08:54, brandon s. allbery kf8nh wrote: 
> pstree is available on more platforms than Linux /proc. Do 
> remember xmonad originated on netbsd, not Linux. 
> 
 
Oh very good point, I did not think about this. 

On Mon, Jun 22 2020 11:39, ElKowar wrote: 
> should i try to add parsing `/proc/PID/stat` as a fallback 
> maybe? 
> 

I personally would prefer it, but then I suppose the *BSD people would
want you to support their program-agnostic solution as well :P It's your
pr of course (and I suppose it could be worse than depending on
something as innocent as `pstree`), so at the end of the day it's your
choice.  If someone really does not want to install this additional
dependency they can always hack it in themselves and submit their own
pr.

geekosaur:
The proper BSD way is via a library, but then xmonad would have to be
setuid root. The pstree route is better.

elkowar:
As seemingly everyone agrees that the dependency on `pstree` is not really an issue, I would say (if the rest is okay) we could merge this. If I (or anyone else) finds a cleaner solution, then we'll add that seperately, I would say!

Issue #359, opened 2020-06-21
Export parsers in EZConfig


Xandaros:
XMonad.Util.EZConfig exports `parseKey`, but not any of the other parsers. In particular, I am interested in `parseKeySequence`.

Some background:
I like using the string representation of a key sequence to define my mappings, but I would like to execute some actions after the first key combination is pressed. This doesn't seem to be possible with the normal submaps, so I decided to just re-implement it. Seemed simple enough.
I would like to re-use the parsers of EZConfig to, well, parse these key sequence strings, since I can't re-use the usual `mkKeymap`. (It would create normal submaps, which I do not want)

The whole point of this is to show possible options for subsequent key-presses in the status bar, which I am used to from i3 and would like to mimic.
If there is an easier way to achieve this, I wouldn't mind hearing about it, but I still think the parsers should be exported.

byorgey:
Makes perfect sense to me.  If you submit a pull request exporting the parsers I will merge it!

PR #360, opened 2020-06-22
Export EZConfig parsers

Xandaros:
Exports parsers in `XMonad.Util.EZConfig`
See #359 

I also included `readKeySequence`, so you don't have to bother with `ReadP` if you need the top-level.


byorgey:
Thanks!

PR #364, opened 2020-07-21
Add XMonad.Hooks.ClickableWorkspaces module

geoffder:
### Description

This module enables clickable workspace tags in **XMobar** in a relatively unobtrusive fashion. The snippet I originally came across ( https://arch-ed.dk/xmobar-clickable-workspaces ) involved wrapping the workspace tags defined in `myWorkspaces` with action tags. This works of course, but now the bulky tags will get in the way of anything that expects the tags to be in a readable format (such as `XMonad.Layout.ShowWName`, which flashes them on the screen), or sorting them alphabetically (for arranging on **XMobar**). 

The pair of hooks included in ClickableWorkspaces (`doClickableWsHook` and `undoClickableWsHook`) can be placed around the dynamicLog hook (e.g `dynamicLogWithPP`), so that the workspace tags fed in to **XMobar** are wrapped for clicking, but are returned to normal immediately after. In addition to these two hooks, `X WorkspaceSort` returning functions are provided to replace those found in `XMonad.Util.WorkspaceCompare`, as the clickable wrappers will interfere with their expected behaviour.

### Checklist

  - [ x ] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ x ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

  - [ x ] I updated the `CHANGES.md` file

  - [ x ] I updated the `XMonad.Doc.Extending` file (if appropriate)


liskin:
Two comments:

1. I think this can be done without modifying the workspace names (and then reverting those modifications, and therefore also without requiring custom comparators), see https://github.com/xmonad/xmonad-contrib/blob/a58c1a6071ceaf83f284b5f641c949818c216318/XMonad/Actions/WorkspaceNames.hs#L138, which also changes the formatting of workspaces for DynamicLog only.

2. A `$usage` documentation is missing which makes it difficult for users unfamiliar with xmonad internals to use this as they'll need to read the source code (or at least the function types) to try to guess what the intended usage is.

geoffder:
@liskin Thank you for mentioning the pretty printer route, I don't know why that hadn't occurred to me. I've re-written the module as a Util, rather than a hook, and added a $usage tag to the new version as per your suggestions.

XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -0,0 +1,71 @@
...
+import XMonad
+import XMonad.Util.WorkspaceCompare (getWsIndex)
+import XMonad.Hooks.DynamicLog (PP(..))
+
+-- $usage
+-- However you have set up your PP, apply @clickablePP@ to it, and bind the result
+-- to "XMonad.Hooks.DynamicLog"\'s dynamicLogWithPP like so:
+-- 
+-- > myPP xmprocs = xmobarPP { ... }
+-- > myLogHook xmprocs = (clickablePP . myPP) xmprocs >>= dynamicLogWithPP
```
liskin, line 32: What is `xmprocs`? My loghook is without arguments, and so is the one in [DynamicLog usage help](https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/XMonad-Hooks-DynamicLog.html).

XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -0,0 +1,71 @@
...
+--
+-- * Requirements:
+--   * xdotool on system (in path)
+--   * use of UnsafeStdinReader in xmobarrc (rather than StdinReader)
+--   * Workspace switch bindings are expected to be "super + i",
+--     where i is the workspace index (beginning with 1), as in default bindings.
```
liskin, line 38: Hardcoding this seems quite unfortunate to me. Personally I'd use `wmctrl -s`. That requires EWMH support in xmonad, but that's probably a safer assumption than "super + i" switching desktops?

liskin:
@geoffder That's indeed much simpler. :-)
Added two additional comments.

XMonad/Util/ClickableWorkspaces.hs
geoffder: Ah sorry I guess I so things a bit weird, I can remove it from the example to avoid confusion. It is a list of xmobar processes that I give to ppOutput in my case.

liskin: Well a more ordinary usage that's compatible with what DynamicLog's usage suggest would definitely be welcome. :-)

geoffder:
Agreed, `wmctrl -s` is definitely a more general solution. `xdotool` has been replaced, and the *Requirements* section has been updated accordingly.

I think the logHook example is clearer now as well, with peculiarities of my config scrubbed out.

XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -47,16 +46,15 @@ xmobarEscape = concatMap doubleLts
 
 clickableWrap :: Integer -> String -> String
 clickableWrap num ws =
-  "<action=xdotool key super+" ++ show num ++ ">" ++ xmobarEscape ws ++ "</action>"
+  "<action=wmctrl -s " ++ show num ++ ">" ++ xmobarEscape ws ++ "</action>"
```
liskin, line 48: Consider using [`xmobarRaw`](https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/XMonad-Hooks-DynamicLog.html#v:xmobarRaw) instead of your own `xmobarEscape`. I think xmobar doesn't unescape the double `<<` back to a single one, does it?

liskin: Oh and you might want to use https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/XMonad-Hooks-DynamicLog.html#v:xmobarAction as well :-)

XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -47,16 +46,15 @@ xmobarEscape = concatMap doubleLts
...
 
--- Use index of workspace in users config to assign workspace switch binding.
+-- Use index of workspace in users config to target workspace with wmctrl switch.
 getClickable :: X (WorkspaceId -> String)
 getClickable = do
   wsIndex <- getWsIndex
   return $ \ws -> case wsIndex ws of
-                    Just idx -> clickableWrap (indexToKey idx) ws
+                    Just idx -> clickableWrap (toInteger idx) ws
```
liskin, line 55: This conversion to `Integer` is unnecessary I think, just change `clickableWrap` to `:: Int -> String -> String`.

liskin:
Awesome!

I found two additional nitpicks, but overall it's looking almost great. :+1: 
Sorry I'm doing the review in these increments, that's just how I am: every time I look at it I see something to improve. :-)

XMonad/Util/ClickableWorkspaces.hs
geoffder: This actually reminded me that I stupidly never got around to testing whether the escape was relevant! I just wrongly assumed that it was relevant (maybe it used to be, when the snippet I saw it in was written?), but from testing it just now, it simply makes double the number of `<`s show up in the bar. I never have had them in my workspace names and forgot to check what would happen if I did Thank you for pointing it out.

New commit is using both of those utility functions, much cleaner (and more correct to boot).

XMonad/Util/ClickableWorkspaces.hs
geoffder: You're right, I think that was a bit of "warning driven development" on my part.

geoffder:
> Awesome!
> 
> I found two additional nitpicks, but overall it's looking almost great. +1
> Sorry I'm doing the review in these increments, that's just how I am: every time I look at it I see something to improve. :-)

No problem, it has definitely made a positive impact on the module.

XMonad/Util/ClickableWorkspaces.hs
liskin: Well, it's unlikely that someone would have a malicious action in their workspace name, but it's easier to just use the correct escape and not worry about it. :-)

Way more important is correct escaping of window titles, which is already handled in the default pp: https://github.com/xmonad/xmonad-contrib/blob/3dc49721b69f5c69c5d5e1ca21083892de72715d/XMonad/Hooks/DynamicLog.hs#L549
Not sure I like this default implementation (I think `xmobarRaw` should be the default for `xmobarPP` these days), but that's totally off-topic here.

XMonad/Doc/Extending.hs
```diff
@@ -1111,6 +1111,11 @@ external utilities.
 
 A non complete list with a brief description:
 
+* "XMonad.Util.ClickableWorkspaces":
+    Provides clickablePP, which when applied to the PP pretty-printer used by
+    'dynamicLogWithPP' of 'XMonad.Hooks.DynamicLog', will make the workspace tags
```
liskin, line 1116: This should be `XMonad.Hooks.DynamicLog.dynamicLogWithPP''` and `"XMonad.Hooks.DynamicLog"`
See https://www.haskell.org/haddock/doc/html/ch03s08.html#idm140354810751008

liskin:
Almost ready to merge this, just a final touchup in the docs.

Would you like to rebase this to clean up the commits a bit but still leave some of the intermediate steps in the git history, or are you happy with me merging it using Squash and merge?

geoffder:
Wow I screwed that up, sorry. Decided to give magit a whirl for the rebase. Re-commit coming up...

geoffder:
Just rebased it into one, but looks like I messed up and forgot to edit to message to say Util insead of Hook.

geoffder:
Well, apparently I'm not so smooth with magit, made some mistakes there. Everything seems to be in order now other than the ugly commits.

liskin:
So, um, I'm a bit confused here. Shall I squash this or preserve the mess? :-)

geoffder:
Sorry I left that out, go ahead and squash it. I'm not sure what I managed to do during the rebase that undid things.

liskin:
Okay. :-)

Thanks for the contribution. :beers:

geoffder:
> Okay. :-)
> 
> Thanks for the contribution. beers

Thank you for the input and the merging!

XMonad/Util/ClickableWorkspaces.hs
liskin: Turns out if someone wants to place icons in their workspace names, this escaping stands in the way. So I think I'll drop it: https://github.com/liskin/xmonad-contrib/commit/283636e29e3f0978c7418664004e27f76bbdfc6e

PR #365, opened 2020-07-22
EZConfig: Learn about XF86Bluetooth

frozencemetery:
### Description

XF86Bluetooth is the keysym sent by function+F10 on my P1.  xev generates this:

```
    state 0x0, keycode 245 (keysym 0x1008ff94, XF86Bluetooth), same_screen YES,
```

Unfortunately XMonad still builds even when the keysyms don't exist, so this was nontrivial to diagnose for an outsider :)

### Checklist

         That repo no longer works; see https://github.com/xmonad/xmonad-testing/issues/15

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -130,6 +130,10 @@
     - Export `HiddenWindows` type constructor.
     - Export `popHiddenWindow` function restoring a specific window.
 
+  * `XMonad.Util.EZConfig`
+
+    - Added support for XF86Bluetooth.
+
 ## 0.16
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


byorgey:
@liskin do we really need a review from @twifkak (not sure if he's still involved)?  Adding a new key seems harmless enough...

liskin:
@byorgey We don't, I just thought I'd be easier for us if the maintainer got a notification and had a look. :-)

If I was the maintainer of this particular piece, I'd probably want the things to be sorted and I'd consider if perhaps the list of keys might be (semi-)automatically retrieved from somewhere, and an n-th pull request adding a single key would maybe trigger me to do something like that. But I'm not even using EZConfig so I have no interest or say in this. Just wanted to do the minimal thing to help get the PR looked at.

byorgey:
OK, that's fair.  But at this point I think I'm going to go ahead and just merge it.

PR #367, opened 2020-07-27
Fix rare issues with windows wrongly dissapearing after windowswallowing

elkowar:
### Description

In rare cases, there was an issue where the WindowSwallowing module failed to store the window stack correctly. Because of this it could happen that after closing a window that swallowed another, a unrelated window was unmapped. This PR fixes that issue by checking if the stored stack is in the expected state, and falling back to a simpler way of restoring the window if not. In these cases, the window is not guaranteed to be positioned exactly where the old child-window was, but that issue seems unavoidable if we failed to store the position of said child window previously. In most cases, this PR does not affect the behaviour.

### Checklist

  - [ ] I updated the `CHANGES.md` file (<- bugfix in module that was not yet part of a release, CHANES entry unnecessary)

### DIFF
--- a/XMonad/Hooks/WindowSwallowing.hs
+++ b/XMonad/Hooks/WindowSwallowing.hs
@@ -44,8 +44,8 @@ where
 import           XMonad
 import qualified XMonad.StackSet               as W
 import qualified XMonad.Util.ExtensibleState   as XS
-import           XMonad.Util.Run                ( runProcessWithInput )
 import           XMonad.Util.WindowProperties
+import           XMonad.Util.Run                ( runProcessWithInput )
 import           Data.Semigroup                 ( All(..) )
 import qualified Data.Map.Strict               as M
 import           Data.List                      ( isInfixOf )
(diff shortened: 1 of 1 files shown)


byorgey:
This seems fine, though it does immediately make me ask: why is the stack state sometimes stored incorrectly?  Can we fix that too?

geekosaur:
One that comes immediately to mind is swallowing in the manageHook, since
the window won't be added to the StackSet in that case unless the hook
passes it on as well as swallowing it.

elkowar:
the stack is stored "incorrectly" if the `configureEvent` is not fired when something there changes. idk _why_ that sometimes happens, but i have been able to reproduce cases where opening some windows just didn't fire that event (nor any other even we could realistically use for that). in that case, the difference between the stored stack and the actual stack is more than just that one window, which causes the issue

geekosaur:
No configure event means an unmanaged window, which won't be in the
StackSet anyway.

PR #375, opened 2020-08-30
Minimize: export type constructor

ivanbrennan:
### Description

Some users like to include type signatures in their configuration. (I am one such user.)

The `Minimize` type constructor was not exported, making it impossible to write a type signature when using `minimize`.

With this change, a user can write an xmonad.hs like:
```haskell
import XMonad
import XMonad.Layout.Minimize (Minimize, minimize)
import XMonad.Layout.LayoutModifier

myLayout :: ModifiedLayout
            Minimize
            (Choose Tall (Choose (Mirror Tall) Full))
            Window
myLayout = minimize $ layoutHook def

main :: IO ()
main = xmonad def { layoutHook = myLayout }
```

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -148,6 +148,10 @@
   * `XMonad.Hooks.DynamicLog`
     - Added `xmobarBorder` function to create borders around strings.
 
+  * `XMonad.Layout.Minimize`
+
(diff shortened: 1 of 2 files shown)


byorgey:
Thanks!

PR #376, opened 2020-09-02
NamedScratchpads + RefocusLast

ivanbrennan:
### Description

Dismissing a named scratchpad currently moves focus to the bottom of the stack. It would be nice if it refocused the most recently focused window. The `RefocusLast` module provides a way to do this.

Since `RefocusLast` relies on either a layout hook or log hook in the user's config, and because there may be use-cases where refocusing doesn't make sense, I left the behavior of existing exports unchanged and added a new `namedScratchpadActionRLWhen` function.

I'm not very satisfied with the shape of the code, so I'm making this a draft PR and hoping to get some suggestions about how I might clean it up. I have not yet tested with xmonad-testing, but I am currently using this in my local configuration and it's working as expected.

### Checklist

### DIFF
--- a/XMonad/Util/NamedScratchpad.hs
+++ b/XMonad/Util/NamedScratchpad.hs
@@ -36,7 +36,7 @@ import XMonad.Hooks.ManageHelpers (doRectFloat)
 import XMonad.Actions.DynamicWorkspaces (addHiddenWorkspace)
 import XMonad.Hooks.DynamicLog (PP, ppSort)
 import XMonad.Actions.SpawnOn (spawnHere)
-import XMonad.Hooks.RefocusLast (refocusWhen, shiftRLWhen)
+import XMonad.Hooks.RefocusLast (refocusWhen)
 
 import qualified Data.List.NonEmpty as NE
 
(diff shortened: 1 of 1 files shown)


slotThe:
On Wed, Sep 02 2020 15:17, Ivan wrote:
> Dismissing a named scratchpad currently moves focus to the bottom of
> the stack. It would be nice if it refocused the most recently focused
> window.

This sounds more like a bug than anything else to me (or possibly
something else in your config interfering with the scratchpads?).  The
current implementation inserts the scratchpad in the position of the
currently focused window `w`, pushing it down in the stack.  After one
dismisses the scratchpad, xmonad should refocus `w` again.  Indeed, I
just tested this and that's exactly what's happening for me.

ivanbrennan:
@slotThe You're right. I just put a minimal test config together and it refocuses correctly. I have several layout modifiers in my config so something is probably interfering with this. I'll keep this PR open a little while longer while I try to pinpoint the issue.

ivanbrennan:
The interference seems to be coming from `XMonad.Layout.SubLayouts`. I was able to repro with this minimal config:
```haskell
import XMonad
import XMonad.Layout.SubLayouts (subLayout)
import qualified XMonad.Util.NamedScratchpad as NS

keys' = const . M.fromList $
    ((mod4Mask, xK_Shift_R), NS.namedScratchpadAction scratchpads "scratchpad")
  ]

scratchpads = [NS.NS name command (appName =? name) NS.defaultFloating]
  where
    name = "scratchpad"
    command = terminal def ++ " -name " ++ name

layoutHook' = subLayout [] tiled tiled
  where
    tiled = Tall 1 (3/100) (1/2)

main = xmonad def
  { keys = keys',
    manageHook = NS.namedScratchpadManageHook scratchpads,
    layoutHook = layoutHook'
  }
```
(Test details in [this commit](https://github.com/ivanbrennan/xmonad-testing/commit/a64ddc7a2b642f4b6662cd5fb102940bcedd6eef).)

ivanbrennan:
Closing this. The buggy behavior only occurs when `Sublayout` gets involved. Complicating the `NamedScratchpads` code to work around that doesn't seem right. On top of everything else, I found a layout configuration that doesn't involve using sublayouts, and which I like a lot more. In simplified form, it boils down to:
```haskell
ToggleLayouts
  Full
  ( Choose
      Tall
      (ModifiedLayout Selection Tall)
  )
```
along with some help from `BoringWindows` and `RotSlaves`.

PR #378, opened 2020-09-10
BoringWindows: swapUp, swapDown

### LINKED ISSUE #374: BoringWindows lacks boring-aware swapUp/swapDown
### Problem Description

When using Minimize, swapUp and swapDown don't behave correctly if one or more windows on the current workspace are minimized.

For example, on a workspace occupied by three windows _A_, _B_, and _C_, do:
- minimize _B_
- focus _A_
- swapDown

Rather than swapping _A_ and _C_, nothing happens. The same is true if _C_ is focused and you try swapUp.

It seems as if the minimized windows are not totally ignored by the swap behavior, and attempted swaps appear to do nothing until you repeat enough swaps to traverse any minimized windows that were "in the way".

### Configuration File
I put together a demonstration in this [xmonad-testing commit](https://github.com/ivanbrennan/xmonad-testing/commit/cfddc696f1381d64cf59b77bbf4854d50186e7cf), using the following config:
```haskell
import XMonad
import XMonad.Actions.Minimize
import XMonad.Layout.Minimize

main :: IO ()
main = xmonad def
  { layoutHook = minimize (layoutHook def),
    keys = const . M.fromList $
        ((mod4Mask, xK_minus), withFocused minimizeWindow),
        ((mod4Mask, xK_equal), withLastMinimized maximizeWindowAndFocus),
        ((mod4Mask .|. shiftMask, xK_Up), windows W.swapUp),
        ((mod4Mask .|. shiftMask, xK_Down), windows W.swapDown)
      ]
  }
```

### Checklist

  - [x] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

ivanbrennan:
### Description

Add boring-aware `swapUp` and `swapDown` functions. Fixes #374.

A motivating use-case is when using Minimize. Consider a workspace with three windows, _A_, _B_, and _C_, where we:
- minimize _B_
- focus _A_
- swapDown (using the standard `swapDown` provided by XMonad.StackSet)

Rather than swapping _A_ and _C_, nothing appears to happen. A swap did occur, but its effects were not visible, since it swapped with a minimized window. If our intent was to swap _A_ with _C_, we need to perform a second swap.

The boring-aware versions of `swapUp` and `swapDown` automatically swap past boring windows, restoring intuitive behavior.

On a stack containing _n_ windows, of which _b_ are boring, the standard swap functions have rank _n_ (i.e. _n_ swaps take you full circle), while the boring-aware versions have rank _n_ - _b_.

### Checklist

  - [x] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing) (see [commit](https://github.com/ivanbrennan/xmonad-testing/commit/cb8e7d630a8227a4aa74fc325f43318080f3e306))

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -70,6 +70,10 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Layout.BoringWindows`
+
+     Added boring-aware `swapUp` and `swapDown` functions.
+
   * `XMonad.Util.NamedScratchpad`
 
      Added two new exported functions to the module:


ivanbrennan:
It's worth noting that since this is implemented by repeatedly applying swaps until we've traversed any boring windows that were "in the way", it produces results like:
```
      swapUp
      ------
 A          *B*
 boring  →   A
*B*          boring
```
i.e. non-boring windows are swapped relative to each other, but boring windows may also be moved within the stack. That may or may not be desirable depending on how boring windows are being used.

For example, in the context of the `Minimize` layout, it yields intuitive results, with minimized windows remaining invisible, but in the context of a `LimitWindows` layout using `boringAuto`, a swap can affect which windows are hidden, as the layout dictates visibility based on stack position.

An alternative interpretation would be to avoid changing the stack positions of any boring windows. That runs contrary to the way swaps are defined in StackSet, especially with respect to the wrapping case (see https://github.com/xmonad/xmonad/issues/234), so I chose not to go that route here.

I'd like to submit a separate PR adding a swap-like operation that handles the wrapping case by swapping the windows at either end of the stack (rather than rotating the entire stack by one position). I've been calling it a "sift". I think keeping boring windows in their original stack positions would be reasonable for a boring-aware interpretation of that operation.

byorgey:
Thanks!  Looks good to me.

PR #384, opened 2020-09-18
RotateSome

ivanbrennan:
### Description

When using a layout that determines window visibility based on stack position (e.g. `XMonad.Layout.LimitWindows`), it's nice to be able to replace a visible window with one that was hidden, without affecting other visible windows.

Add `surfaceNext` and `surfacePrev`, which treat the focused window and any hidden windows as a ring that can be rotated through the focused position.

More generally, expose a pure function, `rotateSome`, for rotating some elements around a stack while keeping others anchored in place.

### Checklist

  - [x] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing) (see [commit](https://github.com/xmonad/xmonad-testing/commit/3ebc9ba9e9048898e82cd23d0d4436f7c274262a))

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,18 @@
 
 ### New Modules
 
+  * `XMonad.Actions.RotateSome`
+
+    Functions for rotating some elements around the stack while keeping others
+    anchored in place. Useful in combination with layouts that dictate window
+    visibility based on stack position, such as "XMonad.Layout.LimitWindows".
+
+    Export 'surfaceNext' and 'surfacePrev' actions, which treat the focused window
+    and any hidden windows as a ring that can be rotated through the focused position.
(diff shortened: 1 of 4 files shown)


byorgey:
I confess that after reading the documentation and the code, I'm still not 100% sure what this does or when/why I might use it.  Do you think you might be able to add an example or two to the documentation ("Suppose we have windows A, B, C such that...")?  I think some examples would go a long way.

ivanbrennan:
@byorgey I added an example, some comments and a small refactor that should clarify what it does.
c5ff88b

I think you're right that an example goes a long way. But let me know if there are parts that need more clarification.

byorgey:
Ah, that helps a lot, thanks!

PR #386, opened 2020-09-24
Allow case-insensitive shell prompt completion

ivanbrennan:
### Description

Directory/file prompt completions are currently case-sensitive. Provide a way to configure case-insensitive completion.

This is achieved by setting the readline `completion-ignore-case` variable prior to generating completions, e.g.
```sh
bind 'set completion-ignore-case on'
compgen -A directory -- foobar
```

### Checklist

  - [x] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing) (see [commits](https://github.com/xmonad/xmonad-testing/compare/e55b0999801aeddbb66be9cb95f7e6c0f514626f...ivanbrennan:compl-case-sensitivity))

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@
     additional `XPConfig` argument, so that they can take into
     account the given `searchPredicate`.
 
+    A `complCaseSensitivity` field has been added to `XPConfig`, indicating
+    whether case-sensitivity is desired when performing completion.
+
   * `XMonad.Hooks.EwmhDesktops`
 
     It is no longer recommended to use `fullscreenEventHook` directly.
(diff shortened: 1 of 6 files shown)


byorgey:
I haven't read the patch in detail.  But adding all these primed function variants with an extra argument seems like a code smell to me.  Can't we include the information about case sensitivity in some existing configuration record, or something like that?

ivanbrennan:
> Can't we include the information about case sensitivity in some existing configuration record, or something like that?

@byorgey Good point. I could add a field to `XPConfig` and eliminate the need for all of the primed functions other than `getShellCompl'`.

The exception, `getShellCompl`, doesn't take an `XPConfig` argument, though all of its call sites (in this repo) have one on hand. I could updated them to pass it through if we're comfortable making a breaking signature change to `getShellCompl`.

I'll start by adding the `XPConfig` field, defaulting to case-sensitive, and eliminate the other primed functions. I'm happy to handle the `getShellCompl` case whichever way you think makes more sense.

byorgey:
Personally I think we should just leave the primed variant of `getShellCompl`.  It seems a little odd for it to depend on an `XPConfig` , since it's not inherently specific to prompts.

ivanbrennan:
> Personally I think we should just leave the primed variant of `getShellCompl`. It seems a little odd for it to depend on an `XPConfig` , since it's not inherently specific to prompts.

@byorgey Sounds good, thanks.

Also, I kept the primed variant of `directoryMultipleModes`, since that one would've required another breaking change.
https://github.com/xmonad/xmonad-contrib/pull/386/files#diff-5483aab3c43d6b1a84d7ffcfd7b230f6R50-R55

byorgey:
> Also, I kept the primed variant of directoryMultipleModes, since that one would've required another breaking change.

Yeah, I saw that, makes sense.

PR #389, opened 2020-10-13
add XMonad.Layout.VoidBorders

skewerr:
### Description

This module contains a modifier that simply sets X11 window border width
to 0 for every window in the layout it modifies. No efforts are made to
bring the border back, which can be annoying if windows are moved to a
different workspace, but it prevents the "border flash" you get with
XMonad.Layout.NoBorders.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -94,6 +94,11 @@
     `XMonad.Hooks.DynamicLog.dynamicLogWithPP`, will make the workspace tags
     clickable in XMobar (for switching focus).
 
+  * `XMonad.Layout.VoidBorders`
+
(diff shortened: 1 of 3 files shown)


byorgey:
Looks good!  However, in order for this to be usable, we also need to add an entry for it in the `exposed-modules` field of `xmonad-contrib.cabal`.  If you could push up a new commit adding that, I will merge.

skewerr:
> we also need to add an entry for it in the `exposed-modules` field

My bad, I completely forgot to add that. Should be fine now! :)

byorgey:
No worries, thanks!

PR #391, opened 2020-10-14
Fixes for XMonad.Prompt.Unicode

NickHu:
### Description

`typeUnicodePrompt` hasn't worked for some time now, and even before it was kind of buggy. I determined this was caused by `runProcessWithInput` blocking on the prompt, so `xdotool` targets that instead of the active window. This is fixed by using `spawnPipe` instead.

It was also doing its own thing for searching for Unicode characters rather than respecting `XPConfig`, so I've fixed that too. I noticed that I was getting crashes when using `fuzzySort` from XMonad.Prompt.FuzzyMatch, and determined that the cause of the issue is:

```
Prelude.minimum: empty list
```

caused by XMonad.Prompt.Unicode generating a list of completions for a given input which contains some strings which are not necessarily supersequences of the input.

For instance, there is a character that is described by `DIGIT ONE`, but the search function in XMonad.Prompt.Unicode is designed disregard word order so giving `one digit` as input should yield this as a valid completion. With the fuzzy sorter, previously when you reached `one ` it would crash in the manner above because `ONE ` is not a subsequence of `DIGIT ONE`.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -111,6 +111,11 @@
     Added `directoryMultipleModes'`, like `directoryMultipleModes` with an additional
     `ComplCaseSensitivity` argument.
 
+  * `XMonad.Prompt.FuzzyMatch`
+
+    `fuzzySort` will now accept cases where the input is not a subsequence of
+    every completion.
+
   * `XMonad.Prompt.Shell`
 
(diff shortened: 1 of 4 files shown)


byorgey:
Thanks so much!

PR #395, opened 2020-10-27
added toggle fullscreen feature

R4stafa:
### Description

I wanted to be able to toggle the fullscreen state of a window (e.g. with a key press) so I implemented a new FullscreenMessage inside XMonad.Layout.Fullscreen: ToggleFullscreen

### Checklist

        I tested them by building and installing xmonad from my feature branch

### DIFF
--- a/XMonad/Layout/Fullscreen.hs
+++ b/XMonad/Layout/Fullscreen.hs
@@ -89,6 +89,7 @@ fullscreenSupport c = c {
 -- of the above have been sent.
 data FullscreenMessage = AddFullscreen Window
                        | RemoveFullscreen Window
+                       | ToggleFullscreen Window
                        | FullscreenChanged
      deriving (Typeable)
 
@@ -107,6 +108,7 @@ instance LayoutModifier FullscreenFull Window where
   pureMess ff@(FullscreenFull frect fulls) m = case fromMessage m of
     Just (AddFullscreen win) -> Just $ FullscreenFull frect $ nub $ win:fulls
     Just (RemoveFullscreen win) -> Just $ FullscreenFull frect $ delete win $ fulls
+    Just (ToggleFullscreen win) -> Just $ FullscreenFull frect $ (if elem win fulls then delete win $ fulls else nub $ win:fulls)
     Just FullscreenChanged -> Just ff
     _ -> Nothing
 
@@ -123,6 +125,7 @@ instance LayoutModifier FullscreenFocus Window where
   pureMess ff@(FullscreenFocus frect fulls) m = case fromMessage m of
     Just (AddFullscreen win) -> Just $ FullscreenFocus frect $ nub $ win:fulls
     Just (RemoveFullscreen win) -> Just $ FullscreenFocus frect $ delete win $ fulls
+    Just (ToggleFullscreen win) -> Just $ FullscreenFocus frect $ (if elem win fulls then delete win $ fulls else nub $ win:fulls)
     Just FullscreenChanged -> Just ff
     _ -> Nothing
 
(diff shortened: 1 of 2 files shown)


XMonad/Layout/Fullscreen.hs
```diff
@@ -89,6 +89,7 @@ fullscreenSupport c = c {
 -- of the above have been sent.
 data FullscreenMessage = AddFullscreen Window
                        | RemoveFullscreen Window
+                       | ToggleFullscreen Window
```
liskin, line 92: I feel like the above documentation should be updated somehow. It seems that the existing messages are meant to be sent by the manage and event hooks, while the new `ToggleFullscreen` message is intended to be sent manually, so this may need a bit more explanation than usual. Also, is one supposed to send `FullscreenChanged` after sending `ToggleFullscreen`?

(Disclaimer: I'm not a user of X.L.Fullscreen, I only use `fullscreenEventHook` from `X.H.EwmhDesktops`.)

R4stafa: Thx for your feedback! Honestly I am very new to XMonad so I'm not yet that familiar with how things are supposed to be used but I think you are right.

> Also, is one supposed to send FullscreenChanged after sending ToggleFullscreen?

I haven't thought of that but probably yes i guess? If so then maybe it would be better to expose a function like toggleFullscreen that sends both the ToggleFullscreen and FullscreenChanged messages and just for completeness also an addFullscreen and removeFullscreen function.
Then only these functions are supposed to be used manually

R4stafa:
> You might also check under XMonad.Actions for a fullscreen action.

I greped the entire XMonad/Actions directory for "[Ff]ullscreen" but the only occurrences have nothing to do with changing the fullscreen state of a window.
Maybe it would be best to add a module `XMonad.Actions.Fullscreen` that  exposes functions like `addFullscreen`, `removeFullscreen` and `toggleFullscreen` that send the respective `FullscreenMessages`?

liskin:
@R4stafa What exactly are you trying to achieve with all of this, btw? I've been fairly happy with just having `isFullscreen --> doFullFloat` in my managehook, `ewmhFullscreen`, and the following keybindings for fullscreening and tiling again:

```
    , ((modMask,               xK_w     ), withFocused $ \w -> windows $ W.float w (W.RationalRect 0 0 1 1))
    , ((modMask,               xK_t     ), withFocused (windows . W.sink) >> up)
```

so I'm wondering what extra functionality `X.L.Fullscreen` provides. Without understanding that, I'm unfortunately unable to decide whether this addition to `X.L.Fullscreen` makes sense, and the original maintainer isn't even present on github. :-/

R4stafa:
> What exactly are you trying to achieve with all of this, btw?

I want to be able to toggle the fullscreen state of a window with **one** keybinding (like in for example i3wm). In your example you have to press 2 different keys for enabling and disabling fullscreen. I want to do both with the same keybinding depending on what the state currently is so a "fullscreened" window gets "unfullscreened" and vice versa.

Maybe there is a better way of achieving this but I didn't find one.

liskin:
@R4stafa How about this: https://github.com/liskin/dotfiles/commit/659af2ec68c26044f9e6ddf11655856613285685#diff-f3bd9f70ef878f30362ff11bbea7fd1d0d6abde1b4befa44b18cce5a27456204R190

I'm not really sure if it's "better", but it's at least reasonably self-contained and short, and we don't need to think about whether X.L.Fullscreen messages can be safely sent by users, and that's a nice improvement I think. :-)

liskin:
Oh and just to clarify my stance on this PR: I'm not saying we can't merge it. I'm just saying that I don't understand the intricacies of this module and it doesn't seem like anyone who does is around to comment.

If you find that using `ewmhFullscreen` and my `toggleFullscreen` is unsatisfactory and you like `X.L.Fullscreen` better, then we'll need to figure this thing out. Or just merge it and see what happens. :-)

R4stafa:
@liskin I switched to using `ewmhFullscreen` and it works even better in some circumstances. Your solution for toggling fullscreen works for me too, do you think we should merge it or it's better if xmonad users just copy your solution into their config if they need this functionality?

liskin:
@R4stafa Well I'm not sure where we'd put such a thing. It's a bit of a hack, really. :-)

slotThe:
@liskin I think we now have just the place to put little hacks in!

liskin:
@slotThe Oh, indeed we have. I'll try to reduce the number of my open PRs and then submit this and my rescreen hook as well.

liskin:
@R4stafa I've put this in my todo list so I will close this now and let you know once it's merged to https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Util/Hacks.hs

R4stafa:
@liskin Oh nice! Thanks!

liskin:
A little change of plans: to deal with #456, I'll add (un)fullscreen hooks to EwmhConfig (#399, https://github.com/xmonad/xmonad-contrib/issues/456#issuecomment-770193120) and provide an additional module for float-based fullscreen handling which will provide a function for toggling, and will additionaly correctly handle unfulscreening of windows that were floated previously.

(And the rescreen hook turned out to be worth a module on its own as well and has been submitted already as #460.)

PR #398, opened 2020-11-04
Add Typeable constraint to a few LayoutClass instances

liskin:
### Description

This enables adding the Typeable constraint to LayoutClass itself
(https://github.com/xmonad/xmonad/pull/242)
which in turn enables querying the current layout state. That might be
useful to e.g. show the current X.L.WorkspaceDir in xmobar.

Example usage: https://github.com/liskin/dotfiles/commit/6ea6c52aac4d39fa5a788372e005954715cb6d9a

### Checklist

**before merging, I need to finish the above checklist tasks, but I'd like to gather feedback first**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -251,6 +251,11 @@
 
     - Export `Minimize` type constructor.
 
+  * Several `LayoutClass` instances now have an additional `Typeable`
+    constraint which may break some advanced configs. The upside is that we
+    can now add `Typeable` to `LayoutClass` in `XMonad.Core` and make it
+    possible to introspect the current layout and its modifiers.
+
 ## 0.16
 
 ### Breaking Changes
--- a/XMonad/Layout/LayoutBuilder.hs
+++ b/XMonad/Layout/LayoutBuilder.hs
@@ -259,7 +259,7 @@ relBox x y w h = SubBox (Rel x) (Rel y) (Rel w) (Rel h)
 
 --------------------------------------------------------------------------------
 instance ( LayoutClass l1 a, LayoutClass l2 a
-         , Read a, Show a, Show p, Eq a, Typeable a, Predicate p a
+         , Read a, Show a, Show p, Typeable p, Eq a, Typeable a, Predicate p a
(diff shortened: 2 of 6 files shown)


byorgey:
Seems reasonable to me.

liskin:
I thought some more about this and decided that it's best if we merge this as is. This PR doesn't really change or enable  anything, so it doesn't make sense to add documentation and changelog entry for it.

I changed the commit message to say:

> This is a preparation commit that fixes the compile failures that would result in merging that change to xmonad. For this to be generally useful we first need to merge (and ideally also release) that xmonad change,
and then we'll need some documentation and perhaps a type class to help find the right LayoutModifier in the tree of ModifiedLayouts and Choices. That will come later.

So that's my plan now. Get these two preparatory changes merged and released in the right order, and then add a `InspectLayoutModifier` class, something along these lines:

```
class InspectLayoutModifier l a where
    inspectModifier :: (LayoutModifier m a, Typeable m) => l a -> Maybe (m a)

instance (LayoutClass l a, LayoutModifier m a, Typeable m, Typeable a) => InspectLayoutModifier (ModifiedLayout m l) a where
    inspectModifier l = case cast l of
        Just (ModifiedLayout m (_ :: l a)) -> Just m
        Nothing -> Nothing
```

If you still find my plan reasonable, I think this is now ready to be merged and https://github.com/xmonad/xmonad/pull/242 can then be merged as well.

byorgey:
I think it's still worth a brief note in the CHANGELOG, since in theory anyone using these instances may have things break if they don't already have a `Typeable` constraint.

byorgey:
However, overall I think the plan sounds reasonable.

liskin:
You've got a point. :-)

Here:
```
  * Several `LayoutClass` instances now have an additional `Typeable`
￼    constraint which may break some advanced configs. The upside is that we
￼    can now add `Typeable` to `LayoutClass` in `XMonad.Core` and make it
￼    possible to introspect the current layout and its modifiers.
```

force-pushed.

Issue #396, opened 2020-10-30
Breaking change in EWMH window activation


liskin:
#192 introduced a breaking change:

```
  * `XMonad.Hooks.EwmhDesktops`
￼
￼    `ewmh` function will use `logHook` for handling activated window. And now
￼    by default window activation will do nothing.
```

I [believe](https://github.com/xmonad/xmonad-contrib/pull/192/) this breaking change can be avoided if we designed that a bit differently.

#192 changed `ewmhDesktopsEventHook` to invoke `logHook` instead of focusing the window that requested activation through `_NET_ACTIVE_WINDOW`, and now `logHook` is supposed to invoke a `ManageHook` through `activateLogHook` which consults a global `NetActivated` in extensible state to tell if it's being invoked from `ewmhDesktopsEventHook`. This seems convoluted to me.

A much better and simpler design, in my opinion, is to invoke the `ManageHook` directly from `ewmhDesktopsEventHook`, and we just need to figure out how to configure it. I propose adding `ewmh'` which would take an additional `EwmhConfig` record with a `focusHook` field. Additionally, this record can be extended with a `handleFullscreen :: Boolean` to make `ewmhFullscreen` unnecessary (and thus impossible to get the order between `ewmh` and `ewmhFullscreen` wrong, improving the user experience). We should also add a `pagerFocusHook` because while one might want to ignore browsers activating themselves, one usually doesn't want to ignore pagers (rofi, alttab, etc.) asking to activate said browser windows.

I'm marking this as blocker for release because #192 had been reverted before for similar reasons (introducing a major behaviour change, #161), and I believe it's worthwhile to try to avoid breaking changes if possible. Especially if said breaking change comes with a convoluted design. :-)

sgf-dma:
> #192 changed ewmhDesktopsEventHook to invoke logHook instead of focusing the
> window that requested activation through _NET_ACTIVE_WINDOW, and now logHook
> is supposed to invoke a ManageHook through activateLogHook which consults a
> global NetActivated in extensible state to tell if it's being invoked from
> ewmhDesktopsEventHook. This seems convoluted to me.

Well.. it invokes `logHook` because i was told to rewrite it that way in #162.
I still think, that my original design of invoking 'manageHook' is the better
one, but i certainly doesn't know enough to argue that this is correct from
the xmonad point of view. And i was told, that indeed it is wrong, and changed
as was suggested (after all, it does not make any big difference for the
module).

[...]

PR #399, opened 2020-11-05
(wip) Refactor EWMH window activation

liskin:
### Description

Work in progress on #396.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,7 @@
       additional combinators.
 
       TODO: refactor and update and make it not a breaking change
+      TODO: mention doAskUrgent
 
   * `XMonad.Prompt.Directory`
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,8 @@
       `activateLogHook` to your `logHook`. Also, module `X.H.Focus` provides
       additional combinators.
 
+      TODO: refactor and update and make it not a breaking change
+
   * `XMonad.Prompt.Directory`
 
     The `Dir` constructor now takes an additional `ComplCaseSensitivity`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,9 +16,9 @@
   * `XMonad.Hooks.EwmhDesktops`
 
     - It is no longer recommended to use `fullscreenEventHook` directly.
-      Instead, use `ewmhFullscreen` which additionally advertises fullscreen
-      support in `_NET_SUPPORTED` and fixes fullscreening of applications that
-      explicitly check it, e.g. mupdf-gl, sxiv, …
+      Instead, use `ewmh' def{fullscreen = True}` which additionally advertises
+      fullscreen support in `_NET_SUPPORTED` and fixes fullscreening of
+      applications that explicitly check it, e.g. mupdf-gl, sxiv, …
 
       `XMonad.Layout.Fullscreen.fullscreenSupport` now advertises it as well,
       and no configuration changes are required in this case.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,24 +15,21 @@
 
   * `XMonad.Hooks.EwmhDesktops`
 
-    It is no longer recommended to use `fullscreenEventHook` directly.
-    Instead, use `ewmhFullscreen` which additionally advertises fullscreen
-    support in `_NET_SUPPORTED` and fixes fullscreening of applications that
-    explicitly check it, e.g. mupdf-gl, sxiv, …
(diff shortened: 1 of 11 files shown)


liskin:
(test failures expected, we need to merge https://github.com/xmonad/X11/pull/71 and release X11)

TheMC47:
I played around with this a bit, and it's definitely an improvement to the UX in my opinion. Awesome work :tada: I'm not familiar with `X.H.Focus` or `X.A.WorkspaceNames` so I didn't get to try them fully (my config is still pretty basic), but I have some suggestions for the documentation of `XMonad.Hooks.EwmhDesktops` module:

- The examples in your personal dotfiles are nice, especially the one with `activateHook`.  I'm only a bit confused on why one would need `workspaceListSort`
- Regarding fullscreen support:
  -  It might be nice to mention why you *wouldn't* want fullscreen support. I personally don't use it, because I want picture-in-picture support sometimes. So it's maybe a good idea to mention that and refer to `windowedFullscreenFixEventHook`
  -  Since we're making the window float, if the user focuses another window in the same workspace by mistake, they'll have no indication that they're focused on the wrong window. Maybe it makes sense to have a combinator that we can use on `windows W.focusUp` for example, so it only triggers if no window in the current workspace is in fullscreen mode. It doesn't have to be here, I'm just thinking out loud

liskin:
> * The examples in your personal dotfiles are nice, especially the one with `activateHook`.  I'm only a bit confused on why one would need `workspaceListSort`

Yeah, I'll definitely add examples. Basically the current state of the PR is that I haven't even tried to change the docs. :-(

Good point about `workspaceListSort` — I don't actually know why it's there. I think one use might be to hide scratchpad workspaces, but there may be other use cases that would be nice to mention, but I have absolutely no idea about any of these. May need to ask around…

> * Regarding fullscreen support:
>   
>   * It might be nice to mention why you _wouldn't_ want fullscreen support. I personally don't use it, because I want picture-in-picture support sometimes. So it's maybe a good idea to mention that and refer to `windowedFullscreenFixEventHook`

Will do.

>   * Since we're making the window float, if the user focuses another window in the same workspace by mistake, they'll have no indication that they're focused on the wrong window. Maybe it makes sense to have a combinator that we can use on `windows W.focusUp` for example, so it only triggers if no window in the current workspace is in fullscreen mode. It doesn't have to be here, I'm just thinking out loud

There's https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/XMonad-Layout-Fullscreen.html#v:fullscreenFocus which probably deals with this problem. I've never used X.L.Fullscreen personally, but it's an interesting approach if one needs more control than just floats. Oh, and X.L.Fullscreen is another usecase of disabling fullscreen support in ewmh, as the event hooks would conflict otherwise.

Anyway, I intend to add (un)fullscreening hooks to EwmhConfig to make X.L.Fullscreen integrate better, to improve unfullscreening of floated windows, and possibly other fullscreen improvements: I wrote more about this in https://github.com/xmonad/xmonad-contrib/pull/395#issuecomment-773232202
But I get the feeling that fullscreening hooks should probably be in a separate PR, and I should try to reduce the scope of this one (#399): refactor the activation and X.H.Focus stuff, finish docs, review, merge, and then deal with the rest.

TheMC47:
> I should try to reduce the scope of this one (#399): refactor the activation and X.H.Focus stuff, finish docs, review, merge, and then deal with the rest.

that sounds reasonable :+1: tell me if you need help with/feedback on the docs (not sure whether I'll have useful things to say about the implementation)

liskin:
I cherry-picked the non-controversial bits into master over the last few days. I intend to reimplement the rest on top of the extensible config stuff (#547) later.

sgf-dma:
In fact, i don't really understand what review you want from me. I have nothing to say about code, because i already forget (again) how all this work. Or you just want, that i try my config on current master?

liskin:
@sgf-dma Once this is in a usable state (right now it's a bit behind current developments), I'd welcome if you tried using it in your config. I remember you had some concerns about this being unwieldy for certain use cases. If you don't remember or don't care, though, I'm fine with that, and I'll just try my best not to break legitimate use cases.

tl;dr: I'll let you know once it's ready.

sgf-dma:
@liskin Ok, thanks. Well, use cases, which i thought essential (when wrote X.H.Focus) are provided in examples in comments. So, it'd be good, if all examples will work.

liskin:
(just a rebase so that I can my daily driver, nothing exciting yet)

PR #404, opened 2020-11-16
Remove Data.Default Deprecations

slotThe:
### Description

The commit that marked these things as deprecated (0f1b6fb7726fc54b0ef318143ab2ffc43faed8fa) was merged on May 28, 2013; this seems like plenty of time for people to have removed these from their configs.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,11 @@
     The `Dir` constructor now takes an additional `ComplCaseSensitivity`
     argument to indicate whether directory completion is case sensitive.
 
+  * All modules still exporting a `defaultFoo` constructor
+
+    All of these were now removed. You can use the re-exported `def` from
+    `Data.Default` instead.
+
 ### New Modules
 
   * `XMonad.Actions.RotateSome`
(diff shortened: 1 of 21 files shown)


liskin (approved):
💖

psibi:
I will merge this by in a couple of days if there is no objections.

liskin:
@psibi: Not an objection, but an additional info that you might've missed as it was discussed on IRC:

@slotThe asked whether we have any rules when to remove deprecations, and we agreed it **would be nice** to keep things around for at least one full Debian stable cycle, so that people who use the distro provided xmonad-contrib don't have their configs broken without any warnings. So that would mean we shouldn't remove stuff deprecated after 0.14.

**But** then we realized that xmonad doesn't show warnings, and it doesn't even log them into .xsession-errors (stderr is redirected so that it can be shown in an xmessage when the compilation fails, but when it doesn't fail, everything that went to stderr is dicarded entirely). Therefore "don't have their configs broken without any warnings" now means "never remove or change anything, ever". :-(

Probably not feasible to really postpone all changes/deprecations for another few years, but we should make those warnings more visible at least so that in the future we can be more certain that people will have seen the warnings and adapted their configs.

psibi:
@liskin Thanks for writing it down and informing about it! (I'm in the IRC channel, but I don't follow the channel much).

PR #405, opened 2020-11-16
Update XMonad.Actions.TopicSpace

slotThe:
### Description

The `XMonad.Actions.TopicSpace` module is extremely useful, but some
small idiosyncrasies make it harder to use than necessary.  This pr
tries to fix that.  I'll provide a short description of the commits and
what they aim to achieve:

  1. Makes `switchTopic` update the last focused topic properly.
     Before, this functionality was solely used for pretty printing the
     topics, making the documentation of e.g. `maxTopicHistory` very
     confusing.
  2. Adds a new function, `switchNthLastFocusedExclude`, which makes it
     possible to exclude certain topics when querying the history.
  3. Corrects several spelling mistakes and tries to make the overall
     module documentation a little more palatable.  The whole config is
     divided up into smaller chunks and some parts that have nothing to
     do with the module as such (like a lot of `myConfig`) were removed.
     Some more example configuration was also added in places where this
     made sense.
  4. Indents everything to 2 spaces (the majority of the module was
     already indented as such, but there were some outliers).
  5. Properly fixes `maxTopicHistory` by making `setLastFocusedTopic`
     aware of the length (again, so far `maxTopicHistory` was only used
     for pretty-printing).
  6. Applies some hlint hint
  7. Changes the order of filtering the topic history list and consing
     the given topic onto it (because we respect `maxTopicHistory` now
     this is much more convenient when e.g. preventing certain topics
     from ever entering the history).
  8. Adds `switchTopicWith`, which allows one to pass a custom 
     filtering function to `setLastFocusedTopic`
  9. Adds a new function to `XMonad.Hooks.WorkspaceHistory`, in order to easily
     modify the workspace history with a pure function, because
  10. Does away with the hand-rolled history that `TopicSpace` was using and uses
      the one provided by `XMonad.Hooks.WorkspaceHistory` instead.  This is not a
      user-facing change, as something like `getLastFocusedTopics` has just been
      aliased.
  11. Adds new, screen-aware, functions `getLastFocusedTopicsByScreen` and
      `switchNthLastFocusedByScreen` (that do what they say).
  12. Updates the changelog

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -177,9 +177,9 @@
   * `XMonad.Util.NamedScratchpad`
 
      Added two new exported functions to the module:
-     - `customRunNamedScratchpadAction` 
-         (provides the option to customize the `X ()` action the scratchpad is launched by) 
-     - `spawnHereNamedScratchpadAction` 
+     - `customRunNamedScratchpadAction`
+         (provides the option to customize the `X ()` action the scratchpad is launched by)
+     - `spawnHereNamedScratchpadAction`
          (uses `XMonad.Actions.SpawnOn.spawnHere` to initially start the scratchpad on the workspace it was launched on)
 
   * `XMonad.Util.Run`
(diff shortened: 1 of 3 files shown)


slotThe:
@byorgey I saw that you're using this module, would you care to review this pr?

byorgey:
@slotThe will do, thanks for the nudge!

byorgey:
Looks great to me!  @slotThe , thanks so much for all the improvements!

PR #407, opened 2020-11-17
X.L.SubLayouts: Avoid moving floats to end of window stack

liskin:
### Description

This makes the following sequence of operations idempotent, as it should be:

    windows $ W.float w $ W.RationalRect 0 0 1 1
    windows $ W.sink w

Previously, any window not visible to the Sublayout modifier (xmonad
invokes runLayout with tiled windows only) would be reordered to the end
of stack. This commit changes the reordering logic to only reorder
windows in Groups and keep all other windows where they were.

### Checklist

  - (n/a) I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

  - (n/a) I updated the `XMonad.Doc.Extending` file (if appropriate)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -368,6 +368,10 @@
     - Made the history respect words that were "completed" by `alwaysHighlight`
       upon confirmation of the selection by the user.
 
+  * `XMonad.Layout.SubLayouts`
+
+    - Floating windows are no longer moved to the end of the window stack.
+
 ## 0.16
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


slotThe (approved):
I don't use sublayout, so I'm not sure how much my "looks good to me" is actually worth but... looks good to me :)

> This makes the following sequence of operations idempotent, as it should be:
> 
> ```
> windows $ W.float w $ W.RationalRect 0 0 1 1
> windows $ W.sink w
> ```

Since our testing now actually works, this may become one now?

liskin:
> Since our testing actually works now, this could perhaps become one?

Adding some tests for SubLayouts is probably a good idea but I'm not sure we can actually test that specific property, as that involves an actual `windows` invocation which interacts with X (the server, not the monad).

I thought I'd just wait until after the release and then merge this and #430 and see if anyone complains, but maybe that's unnecessarily catious. Adding a test for the affected parts may be better.

liskin:

PR #408, opened 2020-11-18
X.H.DynamicLog: Start recommending poperty-based logging

slotThe:
### Description

Stop recommending pipes as the best/first thing to do and instead recommend
property based logging for all bars that support it.  Also strongly recommend
users to use the already pre-made functions that do all of the plumbing.

Using pipes is very error-prone (see e.g. #400), so this seems like a logical
step.  

Unfortunately, how we go about spawning the status bar in this case is not quite
clear.  We can't use `spawn`, nor `safeSpawn`, because that would spawn the
status bar again (without killing the other one first) when restarting xmonad
via `M-q`.  `spawnOnce` is also unsuited (and only used here as a placeholder)
because users expect changes to their bar configurations to take effect when
they `M-q` restart xmonad.  On IRC, we've discussed saving the PIDs somewhere
and killing them when executing `statusBarProp`, but there was a very good
objection to that:

> <fizzie> I'd be a little wary about something that just blindly sends
> signals. What if the process got killed in the meanwhile (crash or whatever),
> and the PID reused for something else? (Not that there's probably that great
> alternatives, and I'm sure it works in practice.)

I'm putting this up as a draft because, frankly, I've run out of ideas.

TODO: 

  - [x] Update CHANGELOG.md

  - [x] Find a good solution for spawning status bars  

### Checklist

### DIFF
--- a/XMonad/Hooks/DynamicLog.hs
+++ b/XMonad/Hooks/DynamicLog.hs
@@ -25,7 +25,10 @@ module XMonad.Hooks.DynamicLog (
     -- * Drop-in loggers
     dzen,
     dzenWithFlags,
+    xmobarProp,
     xmobar,
+    statusBarProp,
+    statusBarPropTo,
     statusBar,
     statusBar',
     dynamicLog,
@@ -62,20 +65,23 @@ module XMonad.Hooks.DynamicLog (
 import Codec.Binary.UTF8.String (encodeString)
 import Control.Applicative (liftA2)
 import Control.Monad (msum)
-import Data.Char ( isSpace, ord )
-import Data.List (intersperse, stripPrefix, isPrefixOf, sortBy)
-import Data.Maybe ( isJust, catMaybes, mapMaybe, fromMaybe )
-import Data.Ord ( comparing )
-import qualified Data.Map as M
+import Data.Char (isSpace, ord)
+import Data.List (intersperse, isPrefixOf, sortBy, stripPrefix)
+import Data.Map (Map)
+import Data.Maybe (catMaybes, fromMaybe, isJust, mapMaybe)
+import Data.Ord (comparing)
+
+import qualified Data.Map        as M
 import qualified XMonad.StackSet as S
 
 import Foreign.C (CChar)
 
 import XMonad
 
-import XMonad.Util.WorkspaceCompare
 import XMonad.Util.NamedWindows
 import XMonad.Util.Run
+import XMonad.Util.SpawnOnce (spawnOnce)
+import XMonad.Util.WorkspaceCompare
 
 import XMonad.Layout.LayoutModifier
 import XMonad.Hooks.UrgencyHook
@@ -201,12 +207,19 @@ dzen conf = dzenWithFlags flags conf
     bg      = "'#3f3c6d'"
     flags   = "-e 'onstart=lower' -dock -w 400 -ta l -fg " ++ fg ++ " -bg " ++ bg
 
-
--- | Run xmonad with a xmobar status bar set to some nice defaults.
+-- | Run xmonad with a property-based xmobar status bar set to some nice
+-- defaults.
 --
 -- > main = xmonad =<< xmobar myConfig
 -- >
 -- > myConfig = def { ... }
+--
+xmobarProp :: LayoutClass l Window
+           => XConfig l -> IO (XConfig (ModifiedLayout AvoidStruts l))
+xmobarProp conf = statusBarProp "xmobar" xmobarPP toggleStrutsKey conf
+
+-- | This function works like 'xmobarProp', but uses pipes instead of property
+-- logs.
 --
 -- This works pretty much the same as 'dzen' function above.
 --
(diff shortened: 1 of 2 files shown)


liskin:
@slotThe see https://github.com/xmonad/xmonad-contrib/issues/433

TheMC47:
I opened #434 and I rebased it on top of this branch. I'm confused with what's the best thing to do git-wise: should we merge this, then I rebase the other one on top of master?

slotThe:
On Fri, Dec 18 2020 04:35, Yecine Megdiche wrote:
> I opened #434 and I rebased it on top of this branch.

Great!  I just pushed some final changes, as well as doc updates; I'll
convert this from its draft state now

> I'm confused with what's the best thing to do git-wise: should we
> merge this, then I rebase the other one on top of master?

That sounds like the way to go yes

TheMC47:
Great, I've done that

slotThe:
@liskin @psibi This should now be ready for review, if you're inclined

XMonad/Hooks/DynamicLog.hs
```diff
@@ -87,62 +93,155 @@ import XMonad.Hooks.ManageDocks
 -- >    import XMonad
 -- >    import XMonad.Hooks.DynamicLog
 --
--- If you just want a quick-and-dirty status bar with zero effort, try
--- the 'xmobar' or 'dzen' functions:
+-- The recommended way to use this module with xmobar, as well as any other
+-- status bar that supports property logging, is to use on of the following two
```
psibi, line 97: /s/on/one ?

psibi (changes requested):
This is new to me, so I don't have much idea about what's going on. That being said, it LGTM. When you mention property, I'm assuming you are referring to this: https://en.wikipedia.org/wiki/X_Window_System_core_protocol#Properties

If yes, can you add that link as part of haddock ? If not, I would like to learn more about it. :-)

slotThe:
> That being said, it LGTM. When you mention property, I'm assuming you are referring to this: https://en.wikipedia.org/wiki/X_Window_System_core_protocol#Properties
> 
> If yes, can you add that link as part of haddock ? If not, I would like to learn more about it. :-)

Yes indeed!  It literally just dumps it out to such a property, for example in my case: 

``` shell
$ xprop -root
...
_XMONAD_LOG(UTF8_STRING) = "<fc=#f8f8f2> 1:WEB</fc> <fc=#f8f8f2> 2</fc>  <box type=Top width=2 color=#8be9fd>3</box> <fc=#f8f8f2> 6:VPN</fc><fc=#ff79c6> • </fc>Full<fc=#ff79c6> • </fc><fc=#f8f8f2>[</fc><fc=#ff79c6>slot: xprop -root</fc><fc=#f8f8f2>]</fc> <fc=#bbbbbb>[</fc><fc=#bd93f9>DynamicLog.hs</fc><fc=#bbbbbb>]</fc>"
...
```

I've updated the docs with a few links (and have hopefully made it clear enough that these are supplemental, to not scare away potential users :)

psibi:
Thank you!

psibi:
@slotThe I migrated my xmonad + xmobar configuration to use this and it works perfectly. Thanks for the feature!

liskin:
I pushed https://github.com/xmonad/xmonad-contrib/commit/555cd7c59695ac5b53426b7070725ca1a69612d8 with a few corrections. Sorry I didn't look at this earlier. On the other hand, by just pushing it, I saved us both a back-and-forth. :-)

Anyway, a couple remarks:

https://github.com/xmonad/xmonad-contrib/blob/555cd7c59695ac5b53426b7070725ca1a69612d8/XMonad/Hooks/DynamicLog.hs#L143
^^ there's xmonadpropread in xmonad-contrib/scripts, we may clean it up and document it and link to it

https://github.com/xmonad/xmonad-contrib/commit/555cd7c59695ac5b53426b7070725ca1a69612d8#diff-2107a045a6f2cac0a766c17d80d8b0ffb94ebe4954952ccd2d9b4147da80c44cL202
I dropped the text about statusBarProp handling restarts correctly as that isn't yet true (will be fixed by @TheMC47 in #434, so this will need to be added back again).

I added another note about configuring XMonadLog and opened https://github.com/jaor/xmobar/issues/517 because it's silly that users need to do that.

slotThe:
On Sat, Dec 19 2020 10:32, Tomáš Janoušek wrote:
> I pushed
> https://github.com/xmonad/xmonad-contrib/commit/555cd7c59695ac5b53426b7070725ca1a69612d8
> with a few corrections. Sorry I didn't look at this earlier. On the
> other hand, by just pushing it, I saved us both a back-and-forth. :-)

These all look pretty good, thanks!

> https://github.com/xmonad/xmonad-contrib/blob/555cd7c59695ac5b53426b7070725ca1a69612d8/XMonad/Hooks/DynamicLog.hs#L143
> ^^ there's xmonadpropread in xmonad-contrib/scripts, we may clean it
> up and document it and link to it
>

Oh very interesting!  Didn't know that existed, I'll look into
prettifying it a bit and then linking to it (unless you want to do that
:>)

liskin:
I won't get to it, so go ahead. The locale/encoding handling in there is probably broken, too.

PR #410, opened 2020-11-23
Remove (more) Deprecations, Properly Deprecate Modules

slotThe:
### Description

Remove all deprecations (that are not `Data.Default` deprecations) that
were added before 0.14 (which, as it turns out, is all of them).

I already had this done, so I'll post it here as a draft (and more for
reference than anything else) because of something @liskin said in #404,
replicated below.

> @slotThe asked whether we have any rules when to remove deprecations,
> and we agreed it **would be nice** to keep things around for at least
> one full Debian stable cycle, so that people who use the distro
> provided xmonad-contrib don't have their configs broken without any
> warnings. So that would mean we shouldn't remove stuff deprecated
> after 0.14.
> 
> **But** then we realized that xmonad doesn't show warnings, and it
> doesn't even log them into .xsession-errors (stderr is redirected so
> that it can be shown in an xmessage when the compilation fails, but
> when it doesn't fail, everything that went to stderr is dicarded
> entirely). Therefore "don't have their configs broken without any
> warnings" now means "never remove or change anything, ever". :-(
> 
> Probably not feasible to really postpone all changes/deprecations for
> another few years, but we should make those warnings more visible at
> least so that in the future we can be more certain that people will
> have seen the warnings and adapted their configs.

### Checklist


liskin:
Can you also grep the source tree for the "deprecated" word (not just pragma) and progress from a documentation-only deprecation to pragma, pls? I was suprised to not see X.L.Named on the list of things being removed here. :-)

---

Also a quick summary for others of our yesterday's little deprecation warning brainstorming:

I had an idea to use freedesktop notifications to display deprecation warnings. The reasoning being:

* it's less intrusive than xmessage so people who can't deal with them right now can reasonably postpone it without it being too annoying
* unfortunately we can't rely on a notification deamon being available, so the warning may end up not being seen
* that might be a good tradeoff anyway, as people with minimal setup are probably less likely to be affected by deprecations
* maybe consider logging to stderr (.xsession-errors) in addition to xmonad.errors

We might need to somehow parse the output of ghc/stack or perhaps just check if there's any `/deprecation/` in there, roughly at [this point](https://github.com/xmonad/xmonad/blob/c3e032e08e51a41140f63fc37aa9b45d80f3d6b4/src/XMonad/Core.hs#L653).

slotThe:
On Tue, Nov 24 2020 03:12, Tomáš Janoušek wrote:
> Can you also grep the source tree for the "deprecated" word (not just
> pragma) and progress from a documentation-only deprecation to pragma,
> pls? I was suprised to not see X.L.Named on the list of things being
> removed here. :-)

Sure, it's done!  I didn't deprecate any functions inside of X.L.Named,
as deprecating the module itself should be enough.

liskin:
Yep, can confirm that my config now prodoces warnings. :+1:

slotThe:
Here are some additional thoughts that came up on IRC a while back, but I forgot to copy over:

- We don't want to use `dbus` or a similar functionality because that has really quite a lot of dependencies 

- `notify-send` is also iffy, as it doesn't seem to indicate whether it was able to send the notification or not (possibly it's becase no one checks the return value of https://sources.debian.org/src/libnotify/0.7.9-3/tools/notify-send.c/#L278 )

- I'm left to wonder whether the xmessage solution would really be that bad; how often do people who would get annoyed at these messages beause they want to fix this in 6 months time realistically recompile their xmonad configs?

slotThe:
(Rebase to account for https://github.com/xmonad/xmonad-contrib/pull/597)

slotThe:
@liskin this is finally properly rebased and can probably be merged in a few days/weeks/whenever is convenient

CHANGES.md
```diff
@@ -4,10 +4,52 @@
 
 ### Breaking Changes
 
-* `XMonad.Util.NamedScratchpad`:
+  * `XMonad.Util.NamedScratchpad`:
 
   - Deprecated the module; use `XMonad.Util.NamedScratchpad` instead.
```
liskin, line 9: What's the difference between `XMonad.Util.NamedScratchpad` and `XMonad.Util.NamedScratchpad`?

```
1. XMonad.Util.NamedScratchpad
2. XMonad.Util.NamedScratchpad
```

Don't see any :-(

geekosaur: Pretty sure that was intended to be `X.U.Scratchpad`; see https://github.com/xmonad/xmonad-contrib/issues/756.

liskin: Oh, okay

slotThe: Ah yes, silly typo there.  Thanks for pushing a fix already :)

PR #411, opened 2020-11-23
Fix transients managing in X.C.Example

PRESFIL:
### Description

I found a small mistake in the example and decided to fix it. I'm new, don't swear) Please, review my changes.
I tested this on my setup, which is now minimal, without any other manage hooks.

### Checklist

### DIFF
--- a/XMonad/Config/Example.hs
+++ b/XMonad/Config/Example.hs
@@ -72,8 +72,8 @@ myManageHook = composeOne
   [ className =? "Pidgin" -?> doFloat
   , className =? "XCalc"  -?> doFloat
   , className =? "mpv"    -?> doFloat
-  , isDialog              -?> doCenterFloat
 
-    -- Move transient windows to their parent:
-  , transience
+  -- Handle floating windows:
+  , transience            -- move transient windows to their parent
+  , isDialog              -?> doCenterFloat
   ]
--- a/XMonad/Config/Example.hs
+++ b/XMonad/Config/Example.hs
@@ -68,12 +68,12 @@ myXPConfig = def
 --
 -- Use the `xprop' tool to get the info you need for these matches.
 -- For className, use the second value that xprop gives you.
-myManageHook = composeAll
-  [ className =? "Pidgin" --> doFloat
-  , className =? "XCalc"  --> doFloat
-  , className =? "mpv"    --> doFloat
-  ] <+> composeOne
+myManageHook = composeOne
   -- Handle floating windows:
   [ transience            -- move transient windows to their parent
   , isDialog              -?> doCenterFloat
+  ] <+> composeAll
+  [ className =? "Pidgin" --> doFloat
+  , className =? "XCalc"  --> doFloat
+  , className =? "mpv"    --> doFloat
   ]
--- a/XMonad/Config/Example.hs
+++ b/XMonad/Config/Example.hs
@@ -68,10 +68,10 @@ myXPConfig = def
 --
 -- Use the `xprop' tool to get the info you need for these matches.
 -- For className, use the second value that xprop gives you.
-myManageHook = composeOne
-  [ className =? "Pidgin" -?> doFloat
-  , className =? "XCalc"  -?> doFloat
-  , className =? "mpv"    -?> doFloat
+myManageHook = composeAll
+  [ className =? "Pidgin" --> doFloat
+  , className =? "XCalc"  --> doFloat
+  , className =? "mpv"    --> doFloat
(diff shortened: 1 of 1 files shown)


PRESFIL:
Maybe I should add `isFloat` | `isFloating` at the bottom (it there are 2 identical implementations)?

byorgey:
Hi @PRESFIL , thanks so much!  I'm not super familiar with managing transients, can you explain what the mistake was?

liskin:
I'm not familiar with managing transients either, but recently I had the opportunity to learn a few things about ManageHooks and they're a confusing bunch: `composeOne` applies them top to bottom, `composeAll` applies them bottom to top, and `<+>` applies them right to left.

So when you say “handle floating windows after making windows floating” and do `composeOne [… --> doFloat] <+> composeOne [transience, …]`, the "handle" part with `transience` actually happens first.

From looking at what `transience` does, it seems it only moves the window to another workspace and doesn't touch its tiled/floating status, so I think it's meant to be combined (as in `composeAll` or `<+>`, not `composeOne`) with some other managehook that makes it floating. So I'd suggest grouping all those `… -?> do*Float` into one `composeOne` and then add `<+> transience'` on either side. (But in practice, many dialogs will have fixed size window hints and will already be floated by xmonad core itself, so the `isDialog` part isn't really needed as long as `transience'` ends up being applied.)

(And please do verify what I'm saying, I'm not really an expert on ManageHooks, don't use them very much.)

PRESFIL:
@liskin, thanks for pointing out my mistake., Thank you for that pointed to my mistake. I'm not sure, so I'll put my trust on you.

I wanted to make sure that all the actions with *changing the state* of windows were first worked out, and then actions for *applying the location* were worked on this windows. Considering what you said, it should look like this:

```haskell
myManageHook = composeOne
  -- Handle floating windows:
  [ transience            -- move transient windows to their parent
  , isDialog              -?> doCenterFloat
  ] <+> composeOne
  [ className =? "Pidgin" -?> doFloat
  , className =? "XCalc"  -?> doFloat
  , className =? "mpv"    -?> doFloat
  ] 
```

> isDialog part isn't really needed

These two branches now working similar:
- *transience branch* moves on top it transient parent (*transient chain* can be made)
- if in is not transient, *isDialog branch* moves dialog window to center of screen

All of this looks like:
изображение

p.s. MPV spawned later, but showed under old floating windows...

liskin:
Looks good to me now, but again: don't really put all that much trust in me and instead try to verify that it does what you think it should do. :-)
Also, by trying to break the original version and then confirming that the changed version works better, you'll have a much clearer idea what the problem was and how to reproduce it, and will then be able to answer @byorgey's question. ;-)

PRESFIL:
> you'll have a much clearer idea what the problem was and how to reproduce it, and will then be able to answer @byorgey's question

> Hi @PRESFIL , thanks so much! I'm not super familiar with managing transients, can you explain what the mistake was?

In original case, when transient windows appears, it will be managed by `isDialog -?> doCenterFloat`, as it is are dialog window too. And it will appear centered. `transience` branch will be skipped in any case.

Also, if Pidgin will create transient window, it will not be handled nor as transient, nor as dialog. I think default managehook is too straightforward and simple.

I'm trying to fix these points and handle `trancient` for all windows in the same way. Clever examples are better than narrow :smile: (thanks additional info about manageHook. I saw it in the wiki but I didn't think much of it). And it'is not too complicated.

PRESFIL:
@byorgey , @liskin
Have I convinced you?
Have you seen any other problems?
Do I need to fill in CHANGES.md for such little things?

liskin:
I have no problem merging this if you've observed that the new version behaves better than the old one. I still think that managing transients and floating dialogs shouldn't be exclusive, but I have no idea how to test that: we'd need a transient dialog that isn't already floated by xmonad core, and I don't even know how to get a transient window. :-)

And no, there's no need to bother with CHANGES.

PRESFIL:
> ... observed that the new version behaves better than the old one.

I think this works better because previously the transience was ignored and all isDialogs were treated the same. transience everything was there just like that.

> I still think that managing transients and floating dialogs shouldn't be exclusive.
It'is OK, because it is not default Xmonad config :smiley:

> ... I have no idea how to test that ...

I added `className =? "Thunar" --> doSink` to `composeAll` section and newly created transient, dialog and _already-not-floated-by-Xmonad_ window, not managed all of this hooks. All dialogs from previous screenshot are tiled:

image

byorgey:
@PRESFIL  thanks!

Issue #381, opened 2020-09-12
How to run automated tests?


ivanbrennan:
### Problem Description

It's unclear how contributors can run any of the automated tests. This decreases their value and discourages adding new ones.

I recently wanted to add some QuickCheck property tests for a module I was adding. I eventually managed to run my tests as documented in these changes: https://github.com/ivanbrennan/xmonad-contrib/compare/master...ivanbrennan:sift-test

The approach I used is probably not very re-usable or robust, but it demonstrates that it is _possible_ to run the tests, and shows some of the challenges involved in doing so.

I'm curious whether there's an established pattern for how tests can/should be run, and I'd be happy to document it if so. It would be nice if xmonad-contrib.cabal had a test-suite target, though I suspect that would require changes in xmonad to expose the instances/utils in its [tests](https://github.com/xmonad/xmonad/tree/master/tests) directory.

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

TheMC47:
Better late than never: now you can run `stack test` to run the test-suite

ivanbrennan:
@TheMC47 That's great to hear, thanks!
Should this be mentioned in the readme?

TheMC47:
@ivanbrennan  I feel that it should be mentioned somewhere, but I'm not sure where that somewhere is... I think the best place would be https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md

ivanbrennan:
> @ivanbrennan I feel that it should be mentioned somewhere, but I'm not sure where that somewhere is... I think the best place would be https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md

That seems reasonable.

TheMC47:
@ivanbrennan  so it's now mentioned in the CONTRIBUTING.md, I guess we could close the issue

ivanbrennan:
@TheMC47 Thanks for setting up the test suite!

PR #417, opened 2020-12-08
Refactor the test-suite

TheMC47:
### Description
It wasn't clear how to run the tests for xmonad-contrib (see #381 ) and after discussing this on the IRC, refactoring seemed a good idea. The old approach was to link the Instances module from the core library, generate a custom Main file, and run the test-suite with GHC. I had little luck running it with the old approach, mainly because a lot of the tests used old versions of QuickCheck. Through this PR, I tried to:

- Copy over the Instances and Utils modules from the core library, since they can't be exposed (part of the test-suite)
- Clean up and unify the Arbitrary instances added in the different files, and move them to the new Instances module 
- Make the tests compatible with newer versions of QuickCheck
- Grab the test runner from the core library and adapt it, so all tests can be run together

Functionally, I didn't change anything in the tests. Some tests had a high discard rate by QuickCheck, but I don't know how to fix it (the test-suite takes around 10 seconds to run after being compiled, which I believe it's reasonable). 

I'm not sure where to exactly document that people could run the tests, or whether I did something wrong with cabal (This is the first time tinkering with it), so feedback is much appreciated! 

### Checklist


slotThe:
Just poppin' by to say that I did a simple `stack test` and things
worked beautifully!

On Tue, Dec 08 2020 01:15, Yecine Megdiche wrote:
> Some tests had a high discard rate by QuickCheck, but I don't know how
> to fix it (the test-suite takes around 10 seconds to run after being
> compiled, which I believe it's reasonable).

I have fixed problems like this in the past by generating less random
and more "this is the case that we actually care about"-samples.  In
this case (as SwapWorkspaces is the slow one out) this would involve
producing a StackSet that we know contains the desired tags.  However, I
also think that 10 seconds is very acceptable, no need for that much
premature optimization ;)

TheMC47:
Glad it worked! Indeed, the SwapWorkspaces tests have the "slow" ones. There was a problem with having two instances of the Arbitrary (one in RotateSome, one in SwapWorkspaces) and I opted to use the one in SwapWorkspaces, and it's still a bit slow. The tests there also rely on the instance to have certain properties (it's commented somewhere, I forgot exactly what it was)

xmonad-contrib.cabal
```diff
@@ -25,15 +25,17 @@ extra-source-files: README.md CHANGES.md scripts/generate-configs scripts/run-xm
                     scripts/window-properties.sh
                     scripts/xinitrc scripts/xmonad-acpi.c
                     scripts/xmonad-clock.c
-                    tests/genMain.hs
+                    tests/Main.hs
+                    tests/Utils.hs
+                    tests/Instances.hs
```
liskin, line 30: Do we need these now that we have a test-suite with `hs-source-dirs: tests`?

TheMC47: I think we do, or at least that's what I found in the core library https://github.com/xmonad/xmonad/blob/96d4f1fe855c07538ea5833f89a8043ac2224515/xmonad.cabal#L41

liskin: Those lines predate [adding other-modules to the test-suite](https://github.com/xmonad/xmonad/commit/0c97a8975425c525598ef0e3f3e1bfc84f779b1a#diff-5911ea1a0e2ec2cb41389429db836564c3df03359044f63df35d1b0a59ff576dR134).

I mean, it may be required, I am no expert on Cabal. It just seems somewhat unlikely. Can you perhaps try if cabal/stack sdist works properly without it?

TheMC47: Removing them doesn't break anything. I'll push the change

liskin: Thanks :-)

TheMC47: Awesome :smile:

psibi:
@TheMC47 Can you rebase this branch against master to trigger the CI ? Thanks!

TheMC47:
@psibi  done, checks have passed!

PR #421, opened 2020-12-10
Add filterOutWs[PP]

slotThe:
### Description

This generalizes the `namedScratchpadFilterOutWorkspace` and `namedScratchpadFilterOutWorkspacePP` functions from `X.U.NamedScratchpad`, by allowing the user to specify a list of workspace tags themselves.  The resulting functions are `X.U.WorkspaceCompare.filterOutWs` and `X.H.DynamicLog.filterOutWsPP`.

The respective functions in `X.U.NamedScratchpad` have been deprecated accordingly.

### Checklist

### DIFF
--- a/XMonad/Hooks/DynamicLog.hs
+++ b/XMonad/Hooks/DynamicLog.hs
@@ -46,7 +46,7 @@ module XMonad.Hooks.DynamicLog (
     wrap, pad, trim, shorten, shortenLeft,
     xmobarColor, xmobarAction, xmobarBorder,
     xmobarRaw, xmobarStrip, xmobarStripTags,
-    dzenColor, dzenEscape, dzenStrip,
+    dzenColor, dzenEscape, dzenStrip, filterOutWsPP,
 
     -- * Internal formatting functions
     pprWindowSet,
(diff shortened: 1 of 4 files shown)


XMonad/Util/NamedScratchpad.hs
```diff
@@ -190,6 +190,7 @@ namedScratchpadManageHook = composeAll . fmap (\c -> query c --> hook c)
 -- doesn't contain it. Intended for use with logHooks.
 namedScratchpadFilterOutWorkspace :: [WindowSpace] -> [WindowSpace]
 namedScratchpadFilterOutWorkspace = filter (\(W.Workspace tag _ _) -> tag /= scratchpadWorkspaceTag)
+{-# DEPRECATED namedScratchpadFilterOutWorkspace "Use XMonad.Util.WorkspaceCompare.filterOutWs instead" #-}
```
liskin, line 193: Would be more helpful to suggest `XMonad.Util.WorkspaceCompare.filterOutWs scratchpadWorkspaceTag`, but that's not exported. It's probably fine to say it's "NSP" and it'll always be "NSP", but maybe we can just redefine this using filterOutWs(PP) and leave these in?

slotThe: Ah good point, thanks!  I agree that "NSP" changing is extremely unlikely, though it's probably still better to get this right.  The constructor not being exported strikes me as odd, you'd think it would be useful to not have to care about what that specific workspace is called ;)

I think I would be vaguely in favour of exporting the constructor and adjusting the deprecation warning, but if you still think rewriting these in terms of the new functions is the better approach then I don't have any qualms about that either

liskin: I don't have any strong opinion either way. (We don't even know if we'll ever live to see the removing of deprecations added this year. ;-))

slotThe: Okay, I've exported `scratchpadWorkspaceTag` and updated the deprecations; just because we may not live to see the day doesn't mean we can't try ;)

liskin: Another thing that occurred to me now: right now the only examples of usage of the new functions are in the deprecation warning, and once (we may not live to see this, but in theory) the deprecated functions are removed, their documentation will be lost. It would therefore be nice to add documentation about usage of the new functions to `$usage` of NamedScratchpad.

slotThe: Ah very good point, I added a short notice about this to the `$usage` section of the module.  I linked to the functions (yay Haddock!) and they themselves contain the examples of their deprecated counterparts; do you think this is enough?

XMonad/Hooks/DynamicLog.hs
```diff
@@ -508,6 +508,23 @@ xmobarStripTags tags = strip [] where
     openTag str = "<" ++ str ++ "="
     closeTag str = "</" ++ str ++ ">"
 
+-- | Transforms a pretty-printer into one not displaying the given workspaces.
+--
+-- For example, filtering out the @NSP@ workspace before giving the 'PP' to
+-- 'dynamicLogWithPP':
+--
+-- > logHook = dynamicLogWithPP . filterOutWsPP ["NSP"] $ def
```
liskin, line 516: Should this be `[ scratchpadWorkspaceTag ]` instead?

PR #426, opened 2020-12-11
Use the default XFCE terminal with `XMonad.Config.Xfce`

pshevtsov:
### Description

When using xmonad with XFCE the “open terminal” key binding doesn't work, because it tries to start the app which doesn't exist in the system. This PR changes the `terminal` setting to use the default XFCE terminal — `xfce4-terminal`.

### Checklist


liskin:
Interesting, the rename seems to have happened in 2012: https://github.com/xfce-mirror/xfce4-terminal/releases/tag/xfce4-terminal-0.6.0 and nobody complained since than.

Thank you! :-)

PR #427, opened 2020-12-12
Fix Documentation in X.H.FadeWindows

### LINKED ISSUE #419: Use modern window opacity atom in X.H.FadeWindows and X.H.FadeInactive
These modules don't work with picom and probably other modern compositors, because they use the pre-standard window opacity atom instead of `_NET_WM_WINDOW_OPACITY`.

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

slotThe:
### Description

(Probably) fixes #419 (still waiting for confirmation of the person on
IRC for that)

So far, the documentation for X.H.FadeWindows had, as an example, the
following `logHook`:

``` haskell
  logHook = fadeWindowsLogHook myFadeHook

  myFadeHook = composeAll [ isUnfocused --> transparency 0.2
                          ,                 opaque
                          ]
```

The problem is that `composeAll = mconcat` and `mappend` for `FadeHook`
simply throws away the left argument if there's a right one, hence
everything gets marked opaque and `_NET_WM_WINDOW_OPACITY` is never set.

Switching things around solves this problem:

``` haskell
  myFadeHook = composeAll [                 opaque
                          , isUnfocused --> transparency 0.2
                          ]

```

### Checklist

### DIFF
--- a/XMonad/Hooks/FadeWindows.hs
+++ b/XMonad/Hooks/FadeWindows.hs
@@ -205,7 +205,7 @@ fadeWindowsLogHook h =  withWindowSet $ \s -> do
   forM_ visibleWins $ \w -> do
     o <- userCodeDef (Opacity 1) (runQuery h w)
     setOpacity w $ case o of
-                     OEmpty    -> 0.93
+                     OEmpty    -> 1
                      Opacity r -> r
 
 -- | A 'handleEventHook' to handle fading and unfading of newly mapped
--- a/XMonad/Hooks/FadeWindows.hs
+++ b/XMonad/Hooks/FadeWindows.hs
@@ -76,17 +76,26 @@ import           Graphics.X11.Xlib.Extras                (Event(..))
 -- >     , handleEventHook = fadeWindowsEventHook
 -- >     {- ... -}
 -- >
--- > myFadeHook = composeAll [isUnfocused --> transparency 0.2
--- >                         ,                opaque
+-- > myFadeHook = composeAll [                 opaque
+-- >                         , isUnfocused --> transparency 0.2
 -- >                         ]
 --
 -- The above is like FadeInactive with a fade value of 0.2.
 --
--- FadeHooks do not accumulate; instead, they compose from right to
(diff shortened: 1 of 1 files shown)


XMonad/Hooks/FadeWindows.hs
```diff
@@ -76,17 +76,20 @@ import           Graphics.X11.Xlib.Extras                (Event(..))
...
--- as opaque.  The 'opaque' hook above is optional, by the way, as any
--- unmatched window will be opaque by default.
+-- 'FadeHook's do not accumulate; instead, they compose from right to
+-- left like 'ManageHook's, so in the above example @myFadeHook@ will
+-- render unfocused windows at 4/5 opacity and the focused window as
+-- opaque.  This means that, in particular, the order in the above
+-- example is important.
+--
+-- The 'opaque' hook above is optional, by the way, as any unmatched
+-- window will be opaque by default.
```
liskin, line 92: I don't think this one is true, either:
https://github.com/xmonad/xmonad-contrib/blob/f9c3f4ecf62eab2747bb537e693f155a615eb8eb/XMonad/Hooks/FadeWindows.hs#L208-L212
That seems like without `opaque`, unmatched windows get 0.93.

Am I understanding this correctly, @geekosaur?

liskin (approved):
The PR is a clear improvement, but it'd be nice to clarify that one additional thing I just found. :-)

XMonad/Hooks/FadeWindows.hs
slotThe: I just tested this and you seem to be right; can't believe I've not noticed that during testing!

geekosaur:
Yes, that's correct. I think I wanted some kind of "cookie" that left things usable but warned that there might be a missing case; it's arguable that the default should be 1.0, though.

slotThe:
@liskin @geekosaur Okay, I've changed the default to 1 and updated the documentation accordingly

liskin (approved):

PR #429, opened 2020-12-13
Add CI workflow against xmonad latest commit

psibi:
### Description

Add CI against xmonad latest commit

### Checklist


.github/workflows/tests.yml
```diff
@@ -14,18 +14,20 @@ jobs:
...
 
       - name: Cache dependencies
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
           path: ~/.stack
-          key: ${{ runner.os }}-${{ matrix.resolver }}-${{ hashFiles('stack.yaml') }}
+          key: ${{ runner.os }}-${{ matrix.resolver }}-${{ hashFiles(matrix.yaml) }}
           restore-keys: |
+            ${{ runner.os }}-${{ matrix.resolver }}-${{ hashFiles(matrix.yaml) }}
```
liskin, line 30: This should be a no-op. Is it not? If no, it looks like a hack and probably deserves a comment.

.github/workflows/tests.yml
```diff
@@ -14,18 +14,20 @@ jobs:
...
       - name: Clone project
         uses: actions/checkout@v2
 
       - name: Cache dependencies
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
           path: ~/.stack
-          key: ${{ runner.os }}-${{ matrix.resolver }}-${{ hashFiles('stack.yaml') }}
+          key: ${{ runner.os }}-${{ matrix.resolver }}-${{ hashFiles(matrix.yaml) }}
```
liskin, line 28: I still believe `'*.cabal'` should be added here, as otherwise stale caches with missing deps will be used whenever an additional dependency is added to the cabal file.

liskin (changes requested):
Just recording the things I mentioned on IRC so it's not forgotten. :-)

.github/workflows/tests.yml
psibi: @liskin Ah, I see what you mean by no-op here. I have pushed a commit removing this.

.github/workflows/tests.yml
psibi: @liskin I have added `xmonad-contrib.cabal` as an additional hash file.

psibi:
@liskin I have updated the MR and the cache seems to be restored properly (that was the only concern I had with introducing the cabal hashes). If you can give another quick review, I would be happy to get this merged.

.github/workflows/tests.yml
```diff
@@ -14,18 +14,20 @@ jobs:
...
 
       - name: Cache dependencies
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
           path: ~/.stack
-          key: ${{ runner.os }}-${{ matrix.resolver }}-${{ hashFiles('stack.yaml') }}
+          key: ${{ runner.os }}-${{ matrix.resolver }}-${{ hashFiles(matrix.yaml) }}-${{ hashFiles('xmonad-contrib.cabal') }}
           restore-keys: |
+            ${{ runner.os }}-${{ matrix.resolver }}-${{ hashFiles(matrix.yaml) }}
```
liskin, line 30: It's probably not important but can we be consistent here and add a trailing `-` to the first `restore-keys:` entry just like below? :-)

liskin (approved):
One last minor nit but I like it now. Thanks!

.github/workflows/tests.yml
psibi: Done.

psibi:
Merging this PR. The CI failure is expected and is supposed to be fixed in the subsequent PR's.

liskin:
Awesome!

Can you please merge https://github.com/xmonad/xmonad/pull/254 to fix the failures?

psibi:
@liskin Done. BTW, don't you have access to merge it yourself ? If not, I think we should fix that.

liskin:
@psibi No, I'm only in the "Contributors" team, not in the core.

psibi:
@byorgey Can we add @liskin as part of the core team ?

geekosaur:
I've sent the invite.

liskin:
Thanks @geekosaur!

I'll try to use my best judgement but I'd still appreciate if the expectations were said explicitly, like where's the boundary between "just merge it", "seek review from anyone" and "seek review from more senior members". ;-)

geekosaur:
Given how people keep disappearing, I'm not sure we can have boundaries there. One may find oneself the most senior member available.

PR #430, opened 2020-12-13
X.L.SubLayouts: Rewrite updateGroup, updateWs' using stack of stacks

liskin:
### Description

A stack of stacks is a more natural representation of what SubLayouts does: it packs information about the global focus as well as focus in individual groups (sublayouts).

It doesn't carry information about the sublayouts themselves (but a similar structure in X.L.Groups does), so we still use Groups and fromGroups in some places, but future refactor can simplify that as well, I'm sure.

My main motivation for this is that I need to expose the window groups to the user config, and a stack of stacks seems to be a nice data structure for that. The motivation for exposing the groups is that I want to manipulate focus a way that takes groups into account. As an example, I want the following:

 * mod-1, mod-2 to mod-0 switches to n-th group if not already focused, and if focused, focus next in the group

 * show these numbers and window titles in xmobar (like tmux/screen/vim status line), like so:

    ```
    1a weechat  1b browser  2 vim  3 mutt
    ```

Achieving this just using BoringWindows is quite tricky, but with the ability to somehow (InspectLayout, which is work-in-progress, or message with IORef) get the stack of stacks out of SubLayouts, this becomes easy.

---

- [ ] merge https://github.com/xmonad/xmonad-contrib/pull/407 first
- [ ] maybe try to refactor more of SubLayouts to use GroupStack?

### Checklist

### DIFF
--- a/XMonad/Layout/SubLayouts.hs
+++ b/XMonad/Layout/SubLayouts.hs
@@ -62,7 +62,6 @@ import qualified XMonad.Layout.BoringWindows as B
 import qualified XMonad.StackSet as W
 import qualified Data.Map as M
 import Data.Map(Map)
-import qualified Data.Set as S
 
 -- $screenshots
 --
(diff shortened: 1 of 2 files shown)


XMonad/Layout/SubLayouts.hs
```diff
@@ -442,20 +429,48 @@ updateWs = windowsMaybe . updateWs'
...
+    wset = S.fromList (W.integrate st)
+    dead = W.filter (`S.member` wset) -- drop dead windows or entire groups
+    refocus s = focusWindow' f s <|> pure s -- sync focus into groups
+    gs' = mapGroups (refocus <=< dead) gs
+    gset = S.fromList . concatMap W.integrate . M.elems $ gs'
+    -- after refocus, f is either the focused window of some group, or not in
+    -- gs' at all, so `lu f` is never Nothing
+    lu w = if w `S.member` gset
+        then w `M.lookup` gs'
+        else Just (W.Stack w [] [])
```
slotThe, line 463: Wouldn't it be better to just `lu w = fromJust $ ...` here?  If this invariant is ever not satisfied, the current code will also blow up when doing `let Just f' = lu f in f'` above, so it wouldn't be any less safe and it would simplify the construction of the new stack above

liskin: No, the invariant only holds for `f` (the focused window), not for all windows.
I don't like the `let Just f' = …` either but I don't think the invariant can be captured by Haskell's type system. :-/

liskin: But I can surely improve the comment. :-)

liskin:

liskin:
(sorry for the noise, I overestimated GitHub's branch renaming capabilities)

PR #431, opened 2020-12-14
Faster builds

vmchale:
### Description

This speeds up builds, as `xmonad-contrib` is a "terminal" library it is often being compiled when nothing else is, so that it is not concurrent. 

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -127,6 +127,8 @@
 
 ### Bug Fixes and Minor Changes
 
+  * Add `-j +RTS -A32m -RTS` to `ghc-options`
+
   * `XMonad.Actions.DynamicProjects`
 
     The `changeProjectDirPrompt` function respects the `complCaseSensitivity` field
--- a/xmonad-contrib.cabal
+++ b/xmonad-contrib.cabal
@@ -70,7 +70,7 @@ library
         cpp-options: -DXFT
 
     if true
-        ghc-options:    -fwarn-tabs -Wall
+        ghc-options:    -fwarn-tabs -Wall -j +RTS -A32m -RTS
 
     if flag(testing)
         ghc-options:    -fwarn-tabs -Werror
@@ -387,5 +387,5 @@ test-suite properties
                , unix
                , utf8-string
                , xmonad >= 0.15 && < 0.16
-  ghc-options: -i.
+  ghc-options: -i. -j +RTS -A32m -RTS
   cpp-options: -DTESTING


peti (commented):
I don't think it's wise to hard-code machine-specific run-time options in the cabal file. :-1:

liskin:
I agree with @peti that it seems really weird to hardcode such stuff, and I am a living example of someone not compiling xmonad-contrib as the only "terminal" library — in my setup it's built concurrently together with xmobar, ghcid, arbtt and possibly other "terminal" tools.

That being said, it does make the build cca 2.5× faster whenever I need to rebuild just xmonad-contrib (e.g. when rebasing patches on top of master before submitting a PR), so I'd really love to use this. Is there perhaps a way I can explicitly tell stack to use these flags only sometimes? Can we, instead of hardcoding these, just add a few tips somewhere to documentation?

liskin:
Well, adding `-j` to `ghc-options:` in stack.yaml kind of works, but `+RTS …` doesn't as the Cabal wrapper interprets it and fails:

```
2020-12-15 01:44:15.792470: [debug] Run process within /home/tomi/src-haskell/xmonad-contrib/: /home/tomi/.stack/setup-exe-cache/x86_64-linux/Cabal-simple_mPHDZzAJ_3.0.1.0_ghc-8.8.4 … --ghc-options -j --ghc-options +RTS --ghc-options -A32m --ghc-options -RTS --exact-configuration --ghc-option=-fhide-source-paths
2020-12-15 01:44:15.793848: [warn] Cabal-simple_mPHDZzAJ_3.0.1.0_ghc-8.8.4: unknown RTS option: --ghc-options
```

I should probably report to stack that they should add `--RTS` before passing `--ghc-options` :-)

vmchale:
@peti  is it machine-specific? 

My hope was `-A32m` would make it better on parallel machines...

vmchale:
@liskin I think you can do

```
package xmonad-contrib
  ghc-options: +RTS -A32m -RTS
```

should work. Stack surely has such a feature!

liskin:
@vmchale I think I tried that: https://github.com/liskin/dotfiles/commit/2feb09921a16c9d03933542c34e864c5923f3851, but failed: https://github.com/xmonad/xmonad-contrib/pull/431#issuecomment-744988895

liskin:
Probably not the outcome you originally wanted, @vmchale, but I think it's a reasonable compromise:

* I added `ghc-options: -j` to `cabal.haskell-ci` and `stack.yaml`, because we are certain xmonad-contrib is a leaf package there, and it speeds up our CI builds considerably (thank you! :tada:)
* xmonad-contrib.cabal doesn't override this as we shouldn't force this upon users
* I didn't find a way to set RTS flags in stack.yaml, and I don't think it's wise to spend much time on this, as it's just marginal gains, so the added complexity and maintenance burden isn't worth it

Still, thanks a lot for pointing out that `-j` is very noticeable in xmonad-contrib.

Issue #433, opened 2020-12-14
Restart xmobar/dzen without relying on pipes


TheMC47:
### Problem Description

As mentioned in #408 , property-based logging should be encouraged. However, there still isn't a way to restart processes without relying on pipes or PIDs.
This was suggested by @liskin in the IRC:
> <Liskni_si> anyway xmobar sets _NET_WM_PID on its window
> <Liskni_si> so we can just queryTree root windows, look for xmobars, read _NET_WM_PID, check if it matches persistent state, kill
> <Liskni_si> dzen does that as well

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

TheMC47:
PS: I'm trying to familiarize myself with the code-base, and this seemed like a manageable task. Tips are highly appreciated :smile:

liskin:
Tips: there's saving and killing of xmobars in my dotfiles, as well as some queryTrees looking for xmobars to lowerWindow them below trayer. Feel free to steal all of that and just add the pid check.

TheMC47:
Closed by #434

PR #434, opened 2020-12-18
Restarting status bars on restart 

TheMC47:
### Description

After discussing in the IRC channel and in #433 on how we should go about restarting status bars, I tried to add the proposed solution to `XMonad.Hooks.DynamicLog`. This needs to be merged after #408, since I rebased my branch on top of it (this is why the diff looks like a mess and why there are a lot of commits)

Big thanks to @liskin  and @slotThe for the tips and the discussion!

### Checklist

### DIFF
--- a/XMonad/Hooks/DynamicLog.hs
+++ b/XMonad/Hooks/DynamicLog.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE FlexibleContexts, PatternGuards #-}
+{-# LANGUAGE FlexibleContexts, PatternGuards, TypeApplications #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -37,6 +37,10 @@ module XMonad.Hooks.DynamicLog (
     xmonadPropLog,
     xmonadPropLog',
 
+    -- * Specialized spawning and cleaning
+    spawnStatusBarAndRemember,
+    cleanupStatusBars,
+
     -- * Build your own formatter
     dynamicLogWithPP,
     dynamicLogString,
@@ -64,12 +68,15 @@ module XMonad.Hooks.DynamicLog (
 
 import Codec.Binary.UTF8.String (encodeString)
 import Control.Applicative (liftA2)
-import Control.Monad (msum)
+import Control.Exception (try, SomeException)
+import Control.Monad (msum, void)
 import Data.Char (isSpace, ord)
 import Data.List (intersperse, isPrefixOf, sortBy, stripPrefix)
 import Data.Map (Map)
 import Data.Maybe (catMaybes, fromMaybe, isJust, mapMaybe)
 import Data.Ord (comparing)
+import System.Posix.Types (ProcessID)
+import System.Posix.Signals (sigTERM, signalProcessGroup)
 
 import qualified Data.Map        as M
 import qualified XMonad.StackSet as S
@@ -80,8 +87,8 @@ import XMonad
 
 import XMonad.Util.NamedWindows
 import XMonad.Util.Run
-import XMonad.Util.SpawnOnce (spawnOnce)
 import XMonad.Util.WorkspaceCompare
+import qualified XMonad.Util.ExtensibleState as XS
 
 import XMonad.Layout.LayoutModifier
 import XMonad.Hooks.UrgencyHook
@@ -200,7 +207,11 @@ import XMonad.Hooks.ManageDocks
 --
 -- The above has the problem that xmobar will not get restarted whenever you
 -- restart xmonad ('XMonad.Util.SpawnOnce.spawnOnce' will simply prevent your
--- chosen status bar from spawning again).
+-- chosen status bar from spawning again). Using 'statusBarProp', however, takes
+-- care of the necessary plumbing /and/ keeps track of the started status bars, so
+-- they can be correctly restarted with xmonad. This is achieved using
+-- 'spawnStatusBarAndRemember' to start them and 'cleanupStatusBars' to kill
+-- previously started ones.
 --
 -- Even if you don't use a statusbar, you can still use 'dynamicLogString' to
 -- show on-screen notifications in response to some events. For example, to show
@@ -349,7 +360,52 @@ statusBarPropTo :: LayoutClass l Window
 statusBarPropTo prop cmd pp =
     makeStatusBar
         (xmonadPropLog' prop =<< dynamicLogString pp)
-        (spawnOnce cmd)
+        (spawnStatusBarAndRemember cmd)
+
+-- | Spawns a status bar and saves its PID. This is useful when the status bars
+-- should be restarted with xmonad. Use this in combination with 'cleanupStatusBars'.
+--
+-- Note: in some systems, multiple processes might start, even though one command is
+-- provided. This means the first PID, of the group leader, is saved.
+--
+spawnStatusBarAndRemember :: String -- ^ The command used to spawn the status bar
+                          -> X()
+spawnStatusBarAndRemember cmd = do
+  newPid <- spawnPID cmd
+  XS.modify (StatusBarPIDs . (newPid :) . getPIDs)
+
+-- This newtype wrapper, together with the ExtensionClass instance makes use of
+-- the extensible state to save the PIDs bewteen xmonad restarts.
+newtype StatusBarPIDs = StatusBarPIDs { getPIDs :: [ProcessID] }
+  deriving (Show, Read)
+
+instance ExtensionClass StatusBarPIDs where
+  initialValue = StatusBarPIDs []
+  extensionType = PersistentExtension
+
+
+-- | Kills the status bars started with 'spawnStatusBarAndRemember', and resets
+-- the state. This could go for example at the beginning of the startupHook.
+--
+-- Concretely, this function sends a 'sigTERM' to the saved PIDs using
+-- 'signalProcessGroup' to effectively terminate all processes, regardless
+-- of how many were started by using  'spawnStatusBarAndRemember'.
+--
+-- There is one caveat to keep in mind: to keep the implementation simple;
+-- no checks are executed before terminating the processes. This means: if the
+-- started process dies for some reason, and enough time passes for the PIDs
+-- to wrap around, this function might terminate another process that happens
+-- to have the same PID. However, this isn't a typical usage scenario.
+--
+cleanupStatusBars :: X ()
+cleanupStatusBars =
+    getPIDs <$> XS.get
+    >>= (io . mapM_ killPid)
+    >> XS.put (StatusBarPIDs [])
+   where
+    killPid :: ProcessID -> IO ()
+    killPid pidToKill = void $ try @SomeException (signalProcessGroup sigTERM pidToKill)
+
 
 -- | Like 'statusBarProp', but uses pipes instead of property-based logging.
 -- Only use this function if your status bar does not support reading from a
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -196,6 +196,9 @@
       instead of pipe-based logging, due to the various issues associated with
       the latter.
 
+    - Added `spawnStatusBarAndRemember` and `cleanupStatusBars` to provide 
+      a way to safely restart status bars without relying on pipes.
+
   * `XMonad.Layout.BoringWindows`
 
      Added boring-aware `swapUp`, `swapDown`, `siftUp`, and `siftDown` functions.


slotThe (commented):
This looks nice!  I'll approve after having tested it tomorrow.

A small style-thing I've noticed: Since the original module author decided to use four spaces for indentation, having just two functions that use two spaces in there looks a bit weird to me

P.S.: looking at the commits right now it seems like you'll need to rebase again when #408 is merged—sorry about that

XMonad/Hooks/DynamicLog.hs
```diff
@@ -185,74 +286,175 @@ dzenWithFlags flags conf = statusBar ("dzen2 " ++ flags) dzenPP toggleStrutsKey
...
+  pidsToKill <- catMaybes <$> (filterM deleteFilter ws >>= mapM (runQuery pid))
+  io $ traverse_ killPid pidsToKill
+  XS.put (StatusBarPIDs [])
+ where
+  savedPIDs :: X [ProcessID]
+  savedPIDs = getPIDs <$> (XS.get :: X StatusBarPIDs)
+
+  isSaved :: Window -> X Bool
+  isDock :: Window -> X Bool
+  deleteFilter :: Window -> X Bool
```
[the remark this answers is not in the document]
TheMC47: Indeed, thanks for the tip

TheMC47:
@slotThe  Thanks for the review! I'll rebase and squash the commit when this is ready to merge

slotThe:
@TheMC47 I'm having some issues with this.  

Before a restart I see this:
 
``` shell
$ ps axu | rg xmobar
slot     14276  0.0  0.0   1004    68 ?        Ss   15:02   0:00 /bin/sh -c xmobar
slot     14277  1.7  0.5 1073782980 41396 ?    Sl   15:02   0:04 xmobar
slot     14384  0.0  0.0  16944  5912 pts/12   S+   15:06   0:00 rg xmobar
```

And after a restart:

``` shell
$ ps axu | rg xmobar
slot     14276  0.0  0.0   1004    68 ?        Ss   15:02   0:00 /bin/sh -c xmobar
slot     14277  1.8  0.5 1073782980 41856 ?    Sl   15:02   0:05 xmobar
slot     14503  0.0  0.0   1004    68 ?        Ss   15:08   0:00 /bin/sh -c xmobar
slot     14504 20.0  0.4 1073782552 39160 ?    Rl   15:08   0:00 xmobar
slot     14532  0.0  0.0  16944  5856 pts/12   S+   15:08   0:00 rg xmobar
```

So the old bar did not get killed correctly.  I think the problem is that `spawnPID` saves the PID for `/bin/sh -c xmobar`, when we actually want the successor of that (at this point I'm reminded why I have a `succ` in my config at exactly this spot).

I wonder if this is just a quirk with my particular xmobar setup, as I don't remember seeing a workaround like that in other peoples configs, who were killing their xmobars in this or similar ways.

TheMC47:
@slotThe I also wondered why you had the `succ` there. I can't reproduce the problem though

liskin:
> I wonder if this is just a quirk with my particular xmobar setup, as I don't remember seeing a workaround like that in other peoples configs, who were killing their xmobars in this or similar ways.

bash has an undocumented feature that if there's just one command after `-c`, it does an automatic `exec`. Other shells don't necessarily have it, specifically `dash` that is used as the default /bin/sh on Debian definitely does not.

The fix is to always use `exec …`.

TheMC47:
So I just cleaned the history and removed the draft.

> bash has an undocumented feature that if there's just one command after `-c`, it does an automatic `exec`. Other shells don't necessarily have it, specifically `dash` that is used as the default /bin/sh on Debian definitely does not.
> 
> The fix is to always use `exec …`.

@liskin  should we mention that somewhere? Or should I somehow include it in `spawnStatusBarAndRemember`?

liskin:
Tough question. The only way to make sure we don't get a steady stream of user complaints is to make it "just work", and but that probably means implementing that bash trick on xmonad side, because just hardcoding `exec` into `spawnStatusBarAndRemember` breaks stuff like `cd somewhere && xmobar` or `VAR=something xmobar`. We could do this in `spawn` but I don't think we want to. :-/

But we should definitely at least make the defaults work right, so `xmobarProp` (and preferably also `xmobar` and `dzenWithFlags` and any other) should be changed to `exec xmobar` and `exec dzen` and so on. It should also be documented, both here and in `spawn`.

Maybe there is a better way, but I can't think of anything right now.

liskin:
Oh, maybe... `spawn` could just check whether there are _any_ shell metacharacters at all, and if there are none, either exec it directly or prepend `exec`. That's what perl's [exec](https://perldoc.perl.org/functions/exec) does.

We'd obviously still need to document this, but it will improve many uses of `spawn` right away.

liskin:
There's just this one tiny problem: it breaks existing `spawn "exec something"` invocations and changes behaviour of any `spawn "<shell builtin> …"` invocation. :-(

TheMC47:
Not sure what's the right thing to do regarding `spawn`, but I definitely agree that the defaults should work. Will adding `exec` to the commands launching xmobar/dzen break anything?

liskin:
Adding exec to the commands in DynamicLog should be perfectly safe, go ahead.

geekosaur:
Might I suggest looking at process groups instead of trying to get `exec` right?

liskin:
@geekosaur It's good that at least someone here still remembers the Unix basics. Apparently I already forgot this stuff. :-(

So it seems that `spawn` calls [createSession](https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-Process.html#v:createSession) which creates a new session and a new process group, and all we need to do is to negate the pid and kill that process group! @slotThe can you try that?

TheMC47:
@geekosaur  That does sound indeed like a good alternative! If it works for @slotThe, I'll adapt the functions accordingly

slotThe:
On Sat, Dec 19 2020 10:51, Tomáš Janoušek wrote:
> So it seems that `spawn` calls
> [createSession](https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-Process.html#v:createSession)
> which creates a new session and a new process group, and all we need
> to do is to negate the pid and kill that process group! @slotThe can
> you try that?

Killing the process group definitely works, the problem is that we are
also checking whether the PID we saved sets _NET_WM_WINDOW_TYPE_DOCK.
Of course the `/bin/sh -c xmobar` process doesn't even have a window
associated with it and so nothing ever gets killed.

liskin:
Oh, bummer. So, back to "check for metachars and add exec if not already there" I guess? :-/

TheMC47:
@liskin  I had the idea of tagging the status bars with a custom property: Instead of storing the PIDs, we kill all windows that have, for example `_XMONAD_RESTART` set.

geekosaur:
The process group should already handle all of this; we don't need to care
about properties, at least for this part. In particular, the bash that
doesn't spawn a window is still part of the process group.

TheMC47:
@geekosaur I'm not sure I understand how that solves the problem. I'm a bit of a UNIX noob, but what if we save the Group ID instead of the PID?

geekosaur:
The group ID of a process group is the negated process ID of the group
leader. That's why you were told to kill() that.

slotThe:
On Sun, Dec 20 2020 04:26, brandon s allbery kf8nh wrote:
> The process group should already handle all of this; we don't need to
> care about properties, at least for this part. In particular, the bash
> that doesn't spawn a window is still part of the process group.

I'm not sure I follow.  Both processes are part of the same process
group yes, but the original intent with checking for the
_NET_WM_WINDOW_TYPE_DOCK property was that the status bar died for
various reasons and the PID that we have saved was repurposed for a
different process.  We then obviously don't want to kill it and so we
have to at least check if it's a dock before sending a SIGTERM.

geekosaur:
You don't need to check that either: a reused process ID would be in a
different process group.

liskin:
On Sun, Dec 20, 2020 at 05:17:26AM -0800, brandon s allbery kf8nh wrote:
>You don't need to check that either: a reused process ID would be in a
>different process group.

Not necessarily, a reused PID is just as likely to be a process group 
leader.

geekosaur:
If the leader goes away for any reason, so does the rest of the group, so
this still can't happen.

liskin:
This is a massive rabbit hole it seems. Let me recap what this is about: we're basically trying to manage a child process, restarting it when it's alive and not attempting to restart it when it died on its own, and unlike a proper process manager like systemd or runit or upstart that waits on its children, maybe does something on a SIGCHLD and possibly uses pidfds to avoid the pid reuse race condition, we've decided we'll manage without all that because we don't want to change the xmonad main loop. This exercise is futile I'm afraid. :-)

We need to step back and decide what we really want to achieve and what we're willing to sacrifice.

liskin:
> If the leader goes away for any reason, so does the rest of the group, so this still can't happen.

I don't think that's relevant. We're talking about an entirely unrelated process that happens to have the same PID as our xmobar used to have. All bets are off.

TheMC47:
@liskin  why won't something like this work?
 
> @liskin I had the idea of tagging the status bars with a custom property: Instead of storing the PIDs, we kill all windows that have for example `_XMONAD_RESTART` set.

geekosaur:
Hm, yes you would need to check if the leader exited. There's no good way
to manage this while ignoring children. Which, given it already causes
problems anyway (there is special casing for the BSDs already), may mean it
is time for that to change including sending child exit events for contribs
to wait on.

geekosaur:
You don't know how many status bars a spawn started, unless you just want
to set this property on all dock windows in which case just use the dock
property.

TheMC47:
Not really, I have docks that I don't want to be restarted with this (my system-tray, for example). If there is a good way to follow the chain of children and add the Property, then I think this might work

geekosaur:
There isn't one; that's the root of the problem.

liskin:
> @liskin why won't something like this work?
> 
> > @liskin I had the idea of tagging the status bars with a custom property: Instead of storing the PIDs, we kill all windows that have for example `_XMONAD_RESTART` set.

The problem with that is that you don't know which windows to tag in the first place. You'd need xmobar (and any other bar) to do the tagging.

So now that you need the status bar to cooperate anyway, why don't we just add a way to tell the bar to restart, or  something like that? :-)

liskin:
I like what @geekosaur suggests. Add SIGCHLD processing to the core, to make it less likely we'll kill the wrong process. In the meantime we can merge something without safety checks, it's fairly unlikely anyway: xmobars don't usually die, and when they do, it's something you deal with right away rather than wait for the pids to wrap around. :-)

mgsloan:
I have not read the comment thread here, but the comment from @liskin caught my eye.  My fork of XMonad [disables ignoring SIGCHLD](https://github.com/mgsloan/xmonad/commit/8b9aa8a28216051976bc5191ae44b0f8250e3f61) so that my config can do proper process management: https://github.com/mgsloan/mgsloan-dotfiles/blob/master/env/src/Process.hs

TheMC47:
If I'm understanding correctly, we could still do the SIGCHLD without modifying the default signal handlers in the core: calling `installHandler` will internally make the `sigaction` syscall, which means the last set handler is the one that'll be executed. 
This means we could actually manage spawned process by xmonad, without changing the core or breaking configs, and entirely in xmonad-contrib!

liskin:
That's a bit of hack but it should work, yeah. :-)

Oh and there's one additional interesting bit in @mgsloan's config: walking parent pids upwards. That's another way to solve https://github.com/xmonad/xmonad-contrib/pull/434#issuecomment-748479909, although still quite hacky.

(Anyway I haven't given it much thought. I'm quite busy with life these days.)

geekosaur:
Walking parent pids upward works as long as all the parents still exist.

liskin:
On Mon, Dec 21, 2020 at 02:49:03PM -0800, brandon s allbery kf8nh wrote:
>Walking parent pids upward works as long as all the parents still exist.

Which they should in this simple `sh -c xmobar` case. Are we trying to fix anything else as well? :-)

TheMC47:
So I thought about this more and briefly talked with @liskin  on the IRC, and I think even though we don't before killing PIDs, it won't be a problem.

It would only cause a problem when the bar is killed, given enough time for the PIDs to wrap around, and then gets recompiled... which seems not a typical use-case.

However, if you think this is absolutely necessary, this summarizes the "definitely 100% safe" solutions:

- Handle SIGCHLD. This should mean that we write a signal handler, and we remove the saved PIDs when they're killed. However, the signal handler must be of type `IO`, which means we can't really change the xmonad state from inside it, and we probably need to try something with `IORef` to save and modify the PIDs 
-  Walk up the parent PIDs. This seems more doable, as we'd only need to keep track of the xmonad PID (is this saved somewhere?), get the highest parent in the tree, and see if we're saving that

slotThe:
> So I thought about this more and briefly talked with @liskin on the IRC, and I think even though we don't before killing PIDs, it won't be a problem.
> 
> It would only cause a problem when the bar is killed, given enough time for the PIDs to wrap around, and then gets recompiled... which seems not a typical use-case.

I would agree with this overall; this is really quite the edge case and probably not worth the complications it causes.

XMonad/Hooks/DynamicLog.hs
```diff
@@ -351,7 +358,36 @@ statusBarPropTo :: LayoutClass l Window
...
+-- | Kills the status bars started with 'spawnStatusBarAndRemember', and resets
+-- the state. This could go for example at the beginning of the startupHook.
+cleanupStatusBars :: X ()
+cleanupStatusBars =
+    getPIDs <$> (XS.get :: X StatusBarPIDs)
+    >>= (io . mapM_ killPid)
+    >> XS.put (StatusBarPIDs [])
+   where
+    killPid :: ProcessID -> IO ()
+    -- Negate the PID to kill the whole group
```
slotThe, line 388: This can be removed now that we use `signalProcessGroup`; though perhaps it would be good to add a little bit of documentation to `cleanupStatusBars` on /why/ we kill the process group, which edge cases will cause problems etc.

TheMC47: @slotThe  is this better?

XMonad/Hooks/DynamicLog.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE FlexibleContexts, PatternGuards #-}
+{-# LANGUAGE FlexibleContexts, PatternGuards, ScopedTypeVariables, TypeApplications #-}
```
slotThe, line 1: `ScopedTypeVariables` is not needed anymore

XMonad/Hooks/DynamicLog.hs
```diff
@@ -349,7 +356,50 @@ statusBarPropTo :: LayoutClass l Window
...
+--
+-- There is one caveat to keep in mind: to keep the implementation simple;
+-- no checks are executed before terminating the processes. This means: if the
+-- started process dies for some reason, and enough time passes for the PIDs
+-- to wrap around, this function might terminate another process that happens
+-- to have the same PID. However, this isn't a typical usage scenario.
+--
+cleanupStatusBars :: X ()
+cleanupStatusBars =
+    getPIDs <$> (XS.get :: X StatusBarPIDs)
```
slotThe, line 396: Do we actually need this type annotation here?  Seems like it may be inferred via `getPIDs` (ignore this if you think this helps with reading the function).

slotThe:
Looks good to me! 

Just one more thing: in 555cd7c59695ac5b53426b7070725ca1a69612d8 the (not then correct) [docs](https://github.com/xmonad/xmonad-contrib/commit/555cd7c59695ac5b53426b7070725ca1a69612d8#diff-2107a045a6f2cac0a766c17d80d8b0ffb94ebe4954952ccd2d9b4147da80c44cL200-L205) where removed about `statusBarProp` cleaning up after itself; this is now the case and may be added back to the module documentation

slotThe (approved):

TheMC47:
I updated the documentation, if everything is fine as it is I'll squash the commits and push again

TheMC47:
Thank you all for the conversation and for the reviews!

I updated the documentation, if everything is fine as it is I'll squash the commits and push again

geekosaur:
For the record, handling `SIGCHLD` is off the table because Haskell signal handlers are unsafe and essentially can't do anything except throw exceptions to the main thread, which isn't enough given it'll be blocked in `XNextEvent()`.

TheMC47:
Thank you for the information @geekosaur ! I still have a lot to learn about X Programming.

Anyway, I rebased and squashed the commits, so if there are no other remarks about the documentation, this is ready for merge

slotThe:
@liskin any last remarks?  If not then we can merge this I think

liskin:
Looks good! :-)

PR #437, opened 2020-12-30
X.H.DynamicLog: Add ppTitleUnfocused

slotThe:
### Description

Add the `ppTitleUnfocused` field to the `PP` record, in order to show
unfocused windows (on the current workspace) in a status bar.

It's sort of unfortunate we have to do this, instead of having a more
strongly-typed `ppTitle` that, instead of a `String`, takes something
like

``` haskell
  data WindowPP = WindowPP
      { win       :: !Window
      , name      :: !String
      , isUrgent  :: !Bool
      , isFocused :: !Bool
      }
```

But that would require changing the signature of `ppTitle` and thus
probably breaking every config in existence that uses this module...

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -201,8 +201,11 @@
       instead of pipe-based logging, due to the various issues associated with
       the latter.
 
-    - Added `spawnStatusBarAndRemember` and `cleanupStatusBars` to provide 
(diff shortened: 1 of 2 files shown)


liskin (approved):

slotThe:
I will merge this in a couple of days if there are not objections

PR #439, opened 2021-01-01
Added cleanupStatusBars to the startup-hook

TheMC47:
### Description

In #434, `cleanupStatusBars` isn't added to the `startupHook`. Thanks @slotThe for finding the issue  

### Checklist

### DIFF
--- a/XMonad/Hooks/DynamicLog.hs
+++ b/XMonad/Hooks/DynamicLog.hs
@@ -360,7 +360,7 @@ statusBarPropTo :: LayoutClass l Window
 statusBarPropTo prop cmd pp =
     makeStatusBar
         (xmonadPropLog' prop =<< dynamicLogString pp)
-        (spawnStatusBarAndRemember cmd)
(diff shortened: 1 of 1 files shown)


XMonad/Hooks/DynamicLog.hs
```diff
@@ -452,7 +452,7 @@ makeStatusBar lh sh k conf = pure $ docks $ conf
     { layoutHook  = avoidStruts (layoutHook conf)
     , logHook     = logHook conf *> lh
     , keys        = (<>) <$> keys' <*> keys conf
-    , startupHook = startupHook conf *> sh
+    , startupHook = cleanupStatusBars *> startupHook conf *> sh
```
slotThe, line 455: You probably want to add this to `statusBarPropTo` instead, because functions like `statusBar` do their own cleanup via the pipes and hence don't care about `cleanupStatusBars`

TheMC47: Right, yes!

Issue #280, opened 2018-09-13
Layout.NoBorders doesn't remove some borders in multihead setup


volfyd:
### Problem Description

Layout.NoBorders fails to remove borders on inactive screens in the following scenario:

Use smartBorders
On one monitor, fullscreen chrome. The borders disappear.
Then press mod-e to move to another monitor. The borders on chrome reappear.

Reverting https://github.com/xmonad/xmonad-contrib/commit/c6cdb77e3b3d6a3842b914a0cf64357f03fcf0c1 fixes the issue. The code related to the regression is probably under the section 

```
 It seems more efficient to have each layout manage
            -- its own floating windows - and equally valid though untested
            -- against a multihead setup.

```

liskin:
Can you please try if https://github.com/xmonad/xmonad-contrib/pull/440 fixes this?

PR #440, opened 2021-01-01
X.L.NoBorders: Fix float handling on multihead setups

liskin:
### Description

Before c6cdb77e3b3d6a3842b914a0cf64357f03fcf0c1, handling of floats was trivial (too trivial, and thus inefficient and flickery): any float whose rectangle covered (RationalRect 0 0 1 1), regardless of what screen or workspace it belonged to, got its borders hidden.

Then c6cdb77e3b3d6a3842b914a0cf64357f03fcf0c1 added some nuance (OnlyLayoutFloatBelow, OnlyLayoutFloat, OnlyScreenFloat) which requires looking at floats in the context of individual screens. Unfortunately the implementation only worked in singlehead setups, as it always used only the active screen/workspace regardless of which screen/workspace was currently being layouted. This resulted in the new functionality not working correctly on inactive screens and also regressed the handling of fullscreen floats on inactive screens.

The fix is conceptually easy: use the screen that is being layouted.

There is a catch, unfortunately: this code runs in redoLayout, but the information we need is only available in modifyLayout and modifyLayoutWithUpdate. I could store a screen id in the ConfigurableBorder record in modifyLayoutWithUpdate and use it later in redoLayout, but that'd be a massive hack, so what I do instead is to find the relevant screen by searching for one that covers the `lr` (layout/screen rectangle). In most setups, this should be reliable. If it turns out it is not, we can fix this later.

Fixes: c6cdb77e3b3d ("'XMonad.Layout.NoBorders': various improvements:") 
Fixes: https://github.com/xmonad/xmonad-contrib/issues/280

#### X.L.NoBorders: Fix docs of data Ambiguity

In c6cdb77e3b3d6a3842b914a0cf64357f03fcf0c1, several new constructors were added but not "In order of increasing ambiguity (less borders more frequently)" as the docs for data Ambiguity say. Correct the order and clarify that Never only considers ambiguity of tiled windows (as in the doc for smartBorders).

### Checklist

  - (n/a) I updated the `XMonad.Doc.Extending` file (if appropriate)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -378,6 +378,11 @@
       window was up by allowing pointer events to pass through the
       custom tree select event loop.
 
+  * `XMonad.Layout.NoBorders`
+
+    - Fixed handling of floating window borders in multihead setups that was
+      broken since 0.14.
+
 ## 0.16
 
 ### Breaking Changes
--- a/XMonad/Layout/NoBorders.hs
+++ b/XMonad/Layout/NoBorders.hs
@@ -246,23 +246,31 @@ instance SetsAmbiguous Ambiguity where
                                 $ W.screenDetail scr
                             ]
 
+            -- Find the screen containing the workspace being layouted.
+            -- (This is a list only to avoid the need to specialcase when it
+            -- can't be found or when several contain @lr@. When that happens,
+            -- the result will probably be incorrect.)
+            thisScreen = [ scr | scr <- W.screens wset
+                               , screenRect (W.screenDetail scr) `R.supersetOf` lr ]
+
             -- This originally considered all floating windows across all
-            -- workspaces. It seems more efficient to have each layout manage
-            -- its own floating windows - and equally valid though untested
-            -- against a multihead setup. In some cases the previous code would
-            -- redundantly add then remove borders from already-borderless
-            -- windows.
+            -- workspaces. It seems more efficient to have each screen manage
+            -- its own floating windows - and necessary to support the
+            -- additional OnlyLayoutFloat* variants correctly in multihead
+            -- setups. In some cases the previous code would redundantly add
+            -- then remove borders from already-borderless windows.
             floating = do
+                scr <- thisScreen
                 let wz :: Integer -> (Window,Rectangle)
                        -> (Integer,Window,Rectangle)
                     wz i (w,wr) = (i,w,wr)
                     -- For the following: in stacking order lowest -> highest.
                     ts = reverse . zipWith wz [-1,-2..] $ wrs
                     fs = zipWith wz [0..] $ do
-                        w       <- reverse . W.index $ wset
+                        w       <- reverse . W.integrate' . W.stack . W.workspace $ scr
                         Just wr <- [M.lookup w (W.floating wset)]
                         return (w,scaleRationalRect sr wr)
-                    sr = screenRect . W.screenDetail . W.current $ wset
+                    sr = screenRect . W.screenDetail $ scr
                 (i1,w1,wr1) <- fs
                 guard $ case amb of
                     OnlyLayoutFloatBelow ->
--- a/XMonad/Layout/NoBorders.hs
+++ b/XMonad/Layout/NoBorders.hs
@@ -301,9 +301,6 @@ data Ambiguity
         -- ^ This constructor is used to combine the borderless windows
         -- provided by the SetsAmbiguous instances from two other 'Ambiguity'
         -- data types.
-    | OnlyScreenFloat
-        -- ^ Only remove borders on floating windows that cover the whole
-        -- screen.
     | OnlyLayoutFloatBelow
         -- ^ Like 'OnlyLayoutFloat', but only removes borders if no window
         -- stacked below remains visible. Considers all floating windows on the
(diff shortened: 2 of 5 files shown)


slotThe:
I can confirm that this works for me!

One thing I've noticed (which I think is too small to make a separate pr
for): The documentation for `Screen` is wrong as well.  It reads

> -- ^ Borders are never drawn on singleton screens.  With this one you
> -- really need another way such as a statusbar to detect focus.

but that's not at all what's actually happening.  It does draw borders
on singleton screens, e.g. when more than one window is visible.  What
it actually does is never draw borders on singleton _windows_
(i.e. windows that occupy the whole screen), not even when several
screens are connected.

liskin:
> but that's not at all what's actually happening. It does draw borders
> on singleton screens

I think the "singleton" in "singleton screens" means "a screen is singleton iff there's only one window on that screen". "Singleton screen_s_" would be an oxymoron if singleton related to the number of screens, I think.

Disclaimer: English is not my first language.

(Also, the help for OtherIndicated is also a bit confusing: what is "full"? The Full layout?)

slotThe:
On Sun, Jan 03 2021 02:52, Tomáš Janoušek wrote:
> I think the "singleton" in "singleton screens" means "a screen is
> singleton iff there's only one window on that screen". "Singleton
> screen_s_" would be an oxymoron if singleton related to the number of
> screens, I think.
>
> Disclaimer: English is not my first language.

Oh right, that makes more sense.  I think it's still worth mentioning
somewhere in the docs what "singleton screen" means in this context;
having less ambiguity fits with this module, after all :)

> (Also, the help for OtherIndicated is also a bit confusing: what is
> "full"? The Full layout?)

I agree that this seems a bit weird.  This may or may not have
originally referred to the `Full' layout, but really what's important is
that it's a layout that does not display *all* windows (at least that's
how I read [1]).  The `Full' layout of course does this very prominently
by hiding all but one window, but with modifiers as provided in
`X.L.LimitWindows' this is possible in all layouts.

[1] https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Layout/NoBorders.hs#L289

liskin:
> I agree that this seems a bit weird. This may or may not have originally referred to the `Full' layout, but really what's important is that it's a layout that does not display *all* windows (at least that's how I read [1]).

Not displaying all windows should already be covered by the `[w]` pattern match in `tiled`:

https://github.com/xmonad/xmonad-contrib/blob/f9a226e75afe0ecb308a450eb6143981d2e953c5/XMonad/Layout/NoBorders.hs#L238
https://github.com/xmonad/xmonad-contrib/blob/f9a226e75afe0ecb308a450eb6143981d2e953c5/XMonad/Layout/NoBorders.hs#L281-L282

That `length (concat nonF) > length wrs` condition is truly weird. It compares the number of (not necessarily visible) windows on all visible workspaces with the number of visible windows on this workspace. Seems like a bit of an apples/oranges comparison to me.

Also, `singleton $ filter (1==) $ map length nonF` checks all windows, not just visible ones.

https://github.com/xmonad/xmonad-contrib/blob/f9a226e75afe0ecb308a450eb6143981d2e953c5/XMonad/Layout/NoBorders.hs#L287-L290

So with Full layout on both screens in a dualhead setup, it behaves as follows (I actually tested it):

* one window on each screen: borders everywhere, because `filter (1==) …` returns several
* two windows on one, one window on the other: no borders anywhere, because `filter (1==) …` returns `[_]` and that is `singleton`
* ten windows on one, ten windows on the other: no borders anywhere, because `filter (1==) …` returns `[]` and that is `singleton` as well

It's just broken, there's no point in trying to fix the doc. :-(

liskin:
I decided I'm not going to worry about `OtherIndicated` as that's clearly ouf of scope here.
I'll try to improve the doc for `Screen` and add a property or unit test for the issue being fixed, and then I'd probably merge it, as I feel it's clearly a step forward (bug being fixed, doc improved a bit).

liskin:

liskin:
I merged #452 here and added a unit test, and checked that it fails without the fix. It's a bit longer than I hoped, but that length would be amortized if someone added more tests for other NoBorders usecases. :-)

XMonad/Layout/LimitWindows.hs
```diff
@@ -1,4 +1,7 @@
 {-# LANGUAGE CPP, FlexibleInstances, MultiParamTypeClasses, DeriveDataTypeable, PatternGuards #-}
+#ifdef TESTING
+{-# OPTIONS_GHC -Wno-duplicate-exports #-}
+#endif
```
slotThe, line 4: Is there any reason we shouldn't just expose those internals (maybe under a literal `*internals` heading)?  This way we'd get rid of the CPP

liskin: Many people (including me) import xmonad-contrib modules unqualified so exposing stuff named `select` and `update` might lead to extra duplication warnings.

If we really wanted to get rid of CPP it would probably be better to move that stuff to a separate X.L.LimitWindows.Internal module.

slotThe: Ah good point.  We have no precedence of an `Internal` style module while `CPP` is (sadly, at least imo) used all over the place so it's probably fine as it is now

liskin: I am not opposed to going with Internal later (possibly as part of a greater effort to silence warnings), but right now I wanted a quick fix to make it bearable to iterate on the tests in ghcid.

PR #442, opened 2021-01-08
X.Prompt: Correctly Update History with alwaysHighlight

slotThe:
### Description

So far, when `alwaysHighlight` was enabled and the user selected an item
while not having completely written it out in the prompt, the raw input
string and not the eventual completion string would be entered into the
prompt history, which is obviously not the desired behaviour.  This can
cause the history to clutter up with all these abbreviations, making
subsequent invocations of the prompt tedious to work with.

For example, an input of "xm" would narrow to both "xmonad" and
"xmobar", but thanks to `alwaysHighlight` "xmobar" was selected.  If the
user now pressed enter, the prompt would correctly return "xmobar" as
the string to act upon, but "xm" would be entered into the prompt
history!

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -356,10 +356,13 @@
 
     - Added `filterOutWs` for workspace filtering.
 
-  * `XMonad.Prompt
+  * `XMonad.Prompt`
 
     - Accommodate completion of multiple words even when `alwaysHighlight` is
       enabled.
+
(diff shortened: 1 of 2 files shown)


liskin (approved):
I'm not using the prompt very much (didn't even know it has a history and I don't use alwaysHighlight) so I have no idea if there may be any hidden pitfalls, but it looks reasonable. Feel free to merge whenever you see fit.

PR #443, opened 2021-01-11
Refactor X.H.DynamicLog and multiple loggers support

TheMC47:
### Description

This was kinda a big one: 
- I added the datatype `StatusBarConfig` which serves as an abstraction over status bar: 
  - how to start a status bar, 
  - how to log to the status bar,
  - what to do with the status bar on restart,
  - which key to toggle the struts
  - how to toggle the struts
  
   With that, users could have more flexibility on the status bars without having to do the plumbing from scratch. I also made sure the public API does not change. It may be interesting to refactor `X.H.DynamicBars` to also use that abstraction since it offers a nice set of features but I don't recall seeing it used anywhere.

- Added support for multiple status bars, which can be implemented quite naturally given the abstraction provided.

Let me know what you think or if anything needs a second thought! I tried to write examples in the documentation, but tell me if anything needs a rewrite or more clarification
 
### Checklist

### DIFF
--- a/XMonad/Hooks/DynamicLog.hs
+++ b/XMonad/Hooks/DynamicLog.hs
@@ -37,11 +37,7 @@ module XMonad.Hooks.DynamicLog (
 
     xmonadPropLog,
     xmonadPropLog',
-
-    -- * Specialized spawning and cleaning
-    spawnStatusBarAndRemember,
-    cleanupStatusBars,
-
+    xmonadDefProp,
     -- * Build your own formatter
     dynamicLogWithPP,
     dynamicLogString,
@@ -58,7 +54,24 @@ module XMonad.Hooks.DynamicLog (
 
     -- * Internal formatting functions
     pprWindowSet,
-    pprWindowSetXinerama
+    pprWindowSetXinerama,
+
+    -- * Specialized spawning and cleaning
+    spawnStatusBarAndRemember,
+    cleanupStatusBars,
+
+    -- * Further customization
+    -- $furthercustomization
+    statusBarPropConfig,
+    statusBarPropToConfig,
+    statusBarHandleConfig,
+    statusBarHandleConfig',
+    makeStatusBar,
+    makeStatusBar',
+    StatusBarConfig(..),
+
+    -- * Multiple loggers
+    -- $multiple
 
     -- * To Do
     -- $todo
@@ -74,7 +87,6 @@ import Control.Monad (msum, void)
 import Data.Bool (bool)
 import Data.Char (isSpace, ord)
 import Data.List (intercalate, isPrefixOf, sortOn, stripPrefix)
-import Data.Map (Map)
 import Data.Maybe (catMaybes, fromMaybe, isJust, mapMaybe)
 import System.Posix.Signals (sigTERM, signalProcessGroup)
 import System.Posix.Types (ProcessID)
@@ -346,7 +358,7 @@ statusBarProp :: LayoutClass l Window
                            -- ^ The desired key binding to toggle bar visibility
               -> XConfig l -- ^ The base config
               -> IO (XConfig (ModifiedLayout AvoidStruts l))
-statusBarProp = statusBarPropTo "_XMONAD_LOG"
+statusBarProp = statusBarPropTo xmonadDefProp
 
 -- | Like 'statusBarProp', but one is able to specify the property to be written
 -- to.
@@ -358,10 +370,8 @@ statusBarPropTo :: LayoutClass l Window
                              -- ^ The desired key binding to toggle bar visibility
                 -> XConfig l -- ^ The base config
                 -> IO (XConfig (ModifiedLayout AvoidStruts l))
-statusBarPropTo prop cmd pp =
-    makeStatusBar
-        (xmonadPropLog' prop =<< dynamicLogString pp)
-        (cleanupStatusBars *> spawnStatusBarAndRemember cmd)
+statusBarPropTo prop a p k c = do {sb <- statusBarPropToConfig prop a p; makeStatusBar' sb k c}
+
 
 -- | Spawns a status bar and saves its PID. This is useful when the status bars
 -- should be restarted with xmonad. Use this in combination with 'cleanupStatusBars'.
@@ -370,7 +380,7 @@ statusBarPropTo prop cmd pp =
 -- provided. This means the first PID, of the group leader, is saved.
 --
 spawnStatusBarAndRemember :: String -- ^ The command used to spawn the status bar
-                          -> X()
+                          -> X ()
 spawnStatusBarAndRemember cmd = do
   newPid <- spawnPID cmd
   XS.modify (StatusBarPIDs . (newPid :) . getPIDs)
@@ -418,9 +428,8 @@ statusBar :: LayoutClass l Window
                        -- ^ The desired key binding to toggle bar visibility
           -> XConfig l -- ^ The base config
           -> IO (XConfig (ModifiedLayout AvoidStruts l))
-statusBar cmd pp k conf = do
-    h <- spawnPipe cmd
-    makeStatusBar (dynamicLogWithPP pp{ ppOutput = hPutStrLn h }) mempty k conf
+statusBar = makeSB .: statusBarHandleConfig
+
 
 -- | Like 'statusBar' with the pretty printing options embedded in the
 -- 'X' monad. The @X PP@ value is re-executed every time the logHook runs.
@@ -432,32 +441,54 @@ statusBar' :: LayoutClass l Window
                            -- ^ The desired key binding to toggle bar visibility
            -> XConfig l    -- ^ The base config
            -> IO (XConfig (ModifiedLayout AvoidStruts l))
-statusBar' cmd xpp k conf = do
-    h <- spawnPipe cmd
-    makeStatusBar
-        (xpp >>= \pp -> dynamicLogWithPP pp{ ppOutput = hPutStrLn h })
-        mempty
-        k
-        conf
+statusBar' = makeSB .: statusBarHandleConfig'
 
--- | Helper function to make status bars.  This should not be used on its own;
--- use 'statusBarProp' or 'statusBar' (or their respective variants) instead.
+-- | Incorporates a 'StatusBarConfig' into an 'XConfig' by taking care of the
+-- necessary plumbing (starting, restarting and logging to it).
+--
+-- Using this function multiple times to combine status bars may result in
+-- only one status bar working properly. See the section on using multiple
+-- status bars for more details.
 makeStatusBar :: LayoutClass l Window
-              => X ()      -- ^ 'logHook' to execute
-              -> X ()      -- ^ 'startupHook' to execute
-              -> (XConfig Layout -> (KeyMask, KeySym))
-                           -- ^ The desired key binding to toggle bar visibility
-              -> XConfig l -- ^ The base config
-              -> IO (XConfig (ModifiedLayout AvoidStruts l))
-makeStatusBar lh sh k conf = pure $ docks $ conf
-    { layoutHook  = avoidStruts (layoutHook conf)
-    , logHook     = logHook conf *> lh
-    , keys        = (<>) <$> keys' <*> keys conf
-    , startupHook = startupHook conf *> sh
+              => StatusBarConfig -- ^ The status bar config
+              -> XConfig l          -- ^ The base config
+              -> IO (XConfig l)
+makeStatusBar (StatusBarConfig lh sh ch) conf =
+  return $ conf
+    { logHook     = logHook conf *> lh
+    , startupHook = startupHook conf *> ch *> sh
     }
-  where
-    keys' :: XConfig Layout -> Map (KeyMask, KeySym) (X ())
-    keys' = (`M.singleton` sendMessage ToggleStruts) . k
+
+-- | Like 'makeStatusBar', but takes an extra key to toggle struts. It also
+-- applies the 'avoidStruts' layout modifier and the 'docks' combinator.
+--
+-- Using this function multiple times to combine status bars may result in
+-- only one status bar working properly. See the section on using multiple
+-- status bars for more details.
+makeStatusBar' :: LayoutClass l Window
+               => StatusBarConfig -- ^ The status bar config
+               -> (XConfig Layout -> (KeyMask, KeySym))
+                                    -- ^ The key binding
+               -> XConfig l          -- ^ The base config
+               -> IO (XConfig (ModifiedLayout AvoidStruts l))
+makeStatusBar' sb k conf = do
+  conf' <- makeStatusBar sb conf
+  return $ docks $ conf' { layoutHook = avoidStruts (layoutHook conf')
+                         , keys       = (<>) <$> keys' <*> keys conf'
+                         }
+  where keys' = (`M.singleton` sendMessage ToggleStruts) . k
+
+-- | A helper function to be used with 'statusBar', 'statusBarProp'
+-- and their variants.
+makeSB :: LayoutClass l Window
+       => IO StatusBarConfig -- ^ The status bar config
+       -> (XConfig Layout -> (KeyMask, KeySym))
+                            -- ^ The key binding
+       -> XConfig l          -- ^ The base config
+       -> IO (XConfig (ModifiedLayout AvoidStruts l))
+makeSB sb k conf = do
+  sb' <- sb
+  makeStatusBar' sb' k conf
 
 -- | Write a string to a property on the root window.  This property is of type
 -- @UTF8_STRING@. The string must have been processed by 'encodeString'
@@ -477,13 +508,17 @@ xmonadPropLog' prop msg = do
 
 -- | Write a string to the @_XMONAD_LOG@ property on the root window.
 xmonadPropLog :: String -> X ()
-xmonadPropLog = xmonadPropLog' "_XMONAD_LOG"
+xmonadPropLog = xmonadPropLog' xmonadDefProp
 
 -- |
 -- Helper function which provides ToggleStruts keybinding
 --
 toggleStrutsKey :: XConfig t -> (KeyMask, KeySym)
 toggleStrutsKey XConfig{modMask = modm} = (modm, xK_b )
+
+-- | The default property xmonad writes to. (@_XMONAD_LOG@).
+xmonadDefProp :: String
+xmonadDefProp = "_XMONAD_LOG"
 
 ------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


slotThe (commented):
LGTM!  Just a few (mostly doc) nits I've noticed

I have to say I'm still not quite sure all of this extra complexity (by which I mean _another_ set of functions with exactly the same name that do things a little differently) is worth having the (admittedly nicer) record interface.  If we decide to go with this though, I think this is the right approach

geekosaur:
Two bars can read the same property provided they expect the same content.
Maybe that's what they meant.

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +766,192 @@ xmobarStripTags tags = strip [] where
...
+--
+-- > ...
+-- > mySBCombinator = (withKey myToggleStrutsKey . withAction myAction . combineSBConfs <$> sequence myBars >>=)
+-- >     . flip makeStatusBar
+-- > ...
+--
+-- The above examples also work if the different status bars support different
+-- logging methods: you could do mix property logging and logging via standard input.
+-- One thing to keep in mind is that if you have multiple status bars with
+-- property logging, they must read from different properties.
```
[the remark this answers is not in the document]
TheMC47: If they read from the same property, they will get the same content. It's not intuitive, but that's how `xmonadPropLog` works. I'll update this with @geekosaur 's comment

XMonad/Hooks/DynamicLog.hs
```diff
@@ -359,10 +374,11 @@ statusBarPropTo :: LayoutClass l Window
...
                 -> IO (XConfig (ModifiedLayout AvoidStruts l))
-statusBarPropTo prop cmd pp =
-    makeStatusBar
-        (xmonadPropLog' prop =<< dynamicLogString pp)
-        (cleanupStatusBars *> spawnStatusBarAndRemember cmd)
+statusBarPropTo prop = makeSBWithKey .: statusBarPropToConfig prop
+
+-- >>> :t withKey toggleStrutsKey <$> statusBarPropToConfig "" "" def
+-- withKey toggleStrutsKey <$> statusBarPropToConfig "" "" def
+--   :: IO StatusBarConfig
```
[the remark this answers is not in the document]
TheMC47: That's leftover from testing (it's called `REPLoid` from `dante.el`, which basically lets me run code in comments directly)

XMonad/Hooks/DynamicLog.hs
```diff
@@ -433,32 +446,35 @@ statusBar' :: LayoutClass l Window
...
-              => X ()      -- ^ 'logHook' to execute
-              -> X ()      -- ^ 'startupHook' to execute
-              -> (XConfig Layout -> (KeyMask, KeySym))
-                           -- ^ The desired key binding to toggle bar visibility
-              -> XConfig l -- ^ The base config
+              => StatusBarConfig -- ^ The status bar config
+              -> XConfig l       -- ^ The base config
               -> IO (XConfig (ModifiedLayout AvoidStruts l))
-makeStatusBar lh sh k conf = pure $ docks $ conf
+makeStatusBar (StatusBarConfig lh sh ch k action) conf = pure $ docks $ conf
```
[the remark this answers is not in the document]
TheMC47: I thought about doing that, but isn't normally the `XConfig l` the last parameter? Or it doesn't matter?

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +766,192 @@ xmobarStripTags tags = strip [] where
 filterOutWsPP :: [WorkspaceId] -> PP -> PP
 filterOutWsPP ws pp = pp { ppSort = (. filterOutWs ws) <$> ppSort pp }
 
+
+-- $internals
+-- Internally, the status bars are managed through the 'StatusBarConfig' datatype that provides a
+-- convenient abstraction over what a status bar is and how to manage it; and the 'makeStatusBar'
+-- which and binds it to an 'XConfig'.
+-- Instead of directly modifying the config, the functions 'statusBarProp' and 'statusBar', create an
+-- appropriate 'StatusBarConfig' and call 'makeStatusBar' on it. This enables the users to have more
```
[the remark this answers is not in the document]
TheMC47: maybe -the?

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +766,192 @@ xmobarStripTags tags = strip [] where
 filterOutWsPP :: [WorkspaceId] -> PP -> PP
 filterOutWsPP ws pp = pp { ppSort = (. filterOutWs ws) <$> ppSort pp }
 
+
+-- $internals
```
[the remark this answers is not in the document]
TheMC47: Hmm, I don't like it either, let's wait if someone has a better idea

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +766,192 @@ xmobarStripTags tags = strip [] where
...
+-- Instead of directly modifying the config, the functions 'statusBarProp' and 'statusBar', create an
+-- appropriate 'StatusBarConfig' and call 'makeStatusBar' on it. This enables the users to have more
+-- flexibility and control over the status bars, without having do the plumbing themselves.
+--
+-- For example, if you want to use property logging as provided by 'statusBarProp', but you want to
+-- customize the action executed to toggle the struts, you could do the following:
+--
+-- > basePropSB = statusBarPropConfig "xmobar" myPP -- default key: mod-b
+-- > mySB = withAction (sendMessage ToggleGaps *> sendMessage ToggleStruts) <$> basePropSB
+-- > main = xmonad =<< (mySB >>= flip makeStatusBar) myConf
```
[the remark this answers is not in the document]
TheMC47: You're right

XMonad/Hooks/DynamicLog.hs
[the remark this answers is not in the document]
slotThe: That would also work yes

XMonad/Hooks/DynamicLog.hs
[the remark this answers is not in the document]
slotThe: It's normally the last parameter for ergonomics reason; you can simply compose things nicer that way.  This breaks down here because `makeStatusBar` is used in a sort of "unintended" way (where we already have the config already, but we want to change something with the status bar options), so having the `XConfig` argument first here would make more sense

XMonad/Hooks/DynamicLog.hs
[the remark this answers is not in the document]
slotThe: Ah I see.  I do think it's intuitive that bars will display the same result if they read the same input, but your explanation made it sounds like something bad would happen if one tried that kind of setup :)  Updating this sounds like a good idea (especially since some people may want identical bars that are just on different monitors)

TheMC47:
Thank you for the review @slotThe ! I'll be pushing again soon

> I have to say I'm still not quite sure all of this extra complexity (by which I mean _another_ set of functions with exactly the same name that do things a little differently) is worth having the (admittedly nicer) record interface. If we decide to go with this though, I think this is the right approach

I'm a bit biased (duh) but I think the added complexity does come with some benefits. Now that we have a first idea how a record based solution might look like, we can better discuss what's the best thing to do here (if we decide we want this): maybe the functions or the record need some more work, or maybe it makes sense to make a separate module :man_shrugging: 

> Two bars can read the same property provided they expect the same content.
@geekosaur  thanks for the comment, that's what I meant

XMonad/Hooks/DynamicLog.hs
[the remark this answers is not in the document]
TheMC47: Maybe my explanation was *slightly* influenced by me wondering for an hour why my bars were displaying the same thing

XMonad/Hooks/DynamicLog.hs
```diff
@@ -433,32 +446,35 @@ statusBar' :: LayoutClass l Window
...
+    , startupHook = startupHook conf *> ch *> sh
     }
   where
     keys' :: XConfig Layout -> Map (KeyMask, KeySym) (X ())
-    keys' = (`M.singleton` sendMessage ToggleStruts) . k
+    keys' = (`M.singleton` action) . k
+
+-- | Like 'makeStatusBar', but takes an extra key to toggle struts.
+-- This is a helper function to be used with 'statusBar' and
+-- 'statusBar''.
```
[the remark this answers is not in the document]
TheMC47: Yes, I updated the comment accordingly

XMonad/Hooks/DynamicLog.hs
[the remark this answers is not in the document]
TheMC47: It should be fixed now

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +766,192 @@ xmobarStripTags tags = strip [] where
...
+--
+-- The different statusBar functions are analogous to the following status bar configs, with the
+-- difference that the configs do not require a key binding (and use @mod-b@ by default):
+--
+-- * 'statusBarPropConfig' is analogous to 'statusBarProp' (with 'statusBarPropTo' analogous to 'statusBarPropConfig')
+--
+-- * 'statusBarHandleConfig' is analogous to 'statusBar' (with 'statusBar'' analogous to 'statusBarHandleConfig'')
+--
+-- The main difference is that the functions returning a status bar config do not require a keybinding
+-- and use @mod-b@. Like the toggle action, this can be customized with 'withKey'.
```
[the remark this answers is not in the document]
TheMC47: What about now?

XMonad/Hooks/DynamicLog.hs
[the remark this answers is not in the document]
slotThe: Because the flip is gone it may now be nice to incorporate the config into the status bar function right away and write something like 

``` haskell
mySB = makeStatusBar myConf . withAction (sendMessage ToggleStruts) =<< basePropSB
-- or, if that makes things clearer
mySB = makeStatusBar myConf =<< (withAction (sendMessage ToggleStruts) <$> basePropSB)
```

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +764,197 @@ xmobarStripTags tags = strip [] where
...
+
+-- | Returns a 'StatusBarConfig' with the given 'toggleAction'. This might
+-- be useful when using a more complex layout in which @sendMessage ToggleStruts@
+-- is not enough to make the status bar disappear.
+-- For example, if you use "XMonad.Layout.Gaps", you might want to also toggle the
+-- gaps:
+--
+-- > mySB = (withAction (sendMessage ToggleGaps *> sendMessage ToggleStruts) <$> statusBarPropConfig "xmobar" myPP >>=)
+-- >     . makeStatusBar
+-- > main = xmonad =<< mySB myConf
```
slotThe, line 831: It's kinda neat to have these examples all be single lines, but this goes a bit too far I think.  Especially someone who may not be that confident in their ability to read operator soup may be pretty lost here.  It may be more worthfile to show a nicely structured `mySB`

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +764,197 @@ xmobarStripTags tags = strip [] where
...
+--
+-- > ...
+-- > mySBCombinator = (withAction myAction . combineSBConfs <$> sequence myBars >>=). makeStatusBar
+-- > main = xmonad =<< mySBCombinator myConfig
+--
+-- if you also want to change the key, you can chain the transformations like so:
+--
+-- > ...
+-- > mySBCombinator = (withKey myToggleStrutsKey . withAction myAction . combineSBConfs <$> sequence myBars >>=)
+-- >     . makeStatusBar
```
slotThe, line 925: Same "problem" here as in 829-831

XMonad/Hooks/DynamicLog.hs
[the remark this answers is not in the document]
TheMC47: Don't you think that might confuse users on how to compose it with other functions?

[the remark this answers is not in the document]
TheMC47: To elaborate: I typically want to have a set of "modifiers" to my base config, that I can simply toggle on and off, so having something that composes with `.` would be convenient.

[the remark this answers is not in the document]
TheMC47: After playing with my config a bit, I think your suggestion is also fine.  I updated the docs accordingly

XMonad/Hooks/DynamicLog.hs
TheMC47: You're right, could you check if it's better now?

XMonad/Hooks/DynamicLog.hs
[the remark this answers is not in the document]
TheMC47: I changed this to `More customization`  which sounds better (and less scary!) than `Internals`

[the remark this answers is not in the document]
slotThe: Perhaps `Further customization`?  Sounds better to my ears, at least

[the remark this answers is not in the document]
TheMC47: Sounds good :+1:

slotThe:
I was playing around with this a little while this morning, and I'm liking it more and more; I will merge this in a few days if no one objects.

TheMC47:
Glad to hear that!

XMonad/Hooks/DynamicLog.hs
```diff
@@ -433,32 +445,34 @@ statusBar' :: LayoutClass l Window
...
+-- necessary plumbing.
 makeStatusBar :: LayoutClass l Window
-              => X ()      -- ^ 'logHook' to execute
-              -> X ()      -- ^ 'startupHook' to execute
-              -> (XConfig Layout -> (KeyMask, KeySym))
-                           -- ^ The desired key binding to toggle bar visibility
-              -> XConfig l -- ^ The base config
+              =>  XConfig l       -- ^ The base config
+              -> StatusBarConfig -- ^ The status bar config
               -> IO (XConfig (ModifiedLayout AvoidStruts l))
```
liskin, line 455: The order of parameters is a bit weird here. It seems to be optimized for the following usecase:
```
main = xmonad =<< makeStatusBar myConf =<< mySB
```
but it quickly gets ugly if one doesn't have a toplevel def for myConf, like in my config: https://github.com/liskin/dotfiles/blob/95495d2966543893349c37c61361a38c1fb104d4/.xmonad/xmonad.hs#L508

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +764,196 @@ xmobarStripTags tags = strip [] where
...
+-- To directly combine confs created by 'statusBarHandleConfig', 'statusBarHandleConfig''
+-- or 'statusBarPropConfig', take a look at 'combineSBs' instead,
+-- which takes instead a list of 'StatusBarConfig's that are wrapped in 'IO'.
+combineSBConfs :: [StatusBarConfig] -- ^ The list of configs to be merged
+               -> StatusBarConfig
+combineSBConfs confs = def
+  { sbLogHook     = mapM_ sbLogHook confs
+  , sbStartupHook = mapM_ sbStartupHook confs
+  , sbCleanupHook = mapM_ sbCleanupHook confs
+  }
```
liskin, line 946: It'd be a bit nicer if StatusBarConfig was a Monoid or Semigroup. We wouldn't need this then.
But it can't be, because of keyToggle and toggleAction. :-(

Frankly, the entire key/action handling seems a bit weird to me. Like, not just the stuff that you added, even the stuff that was already there. That `XConfig → …` stuff that's there just to let the user override the mod key… Seems like a lot of complexity for little benefit. And then you have keyToggle and toggleAction in StatusBarConfig which, at first glance, appears to possibly be useful for toggling individual bars, but then a comment deep in the docs reveals that no, it doesn't do that. :-/

If we can figure out an easier interface for this, I'd say we can break compatibility. Yeah, we should try to keep the basic simple `dzen`, `xmobar` and `xmobarProp` functions as they are including picking up modkey changes, but the advanced interface doesn't need to be dragged down by this.

(We don't need to block the merge because of this, but if we managed to find some good design before the release, that would be super awesome. :-))

liskin: Oh and of course we can go the other way. Adjust the types so that StatusBarConfig can be a Semigroup and toggling of individual statusbar is possible. Not sure if that's worth it, but it's definitely possible.

But I should also say that I don't really have a horse in this race:

My setup is still so custom and complicated that it doesn't fit even this new improved interface. I have one xmobar on the primary screen which uses dynamicLogString and xmonadPropLog, but doesn't use xmobarProp/statusBarPropTo/makeStatusBar precisely because I don't want my keybindings touched, and also because I need a more robust version of cleanupStatusBars (parallel killing of multiple xmobars and a trayer, with retries when they refuse to die, implemented using async, which needs threaded runtime and therefore can't ever go to xmonad-contrib, or maybe it could with some ifdefs or runtime detection or something, dunno). Then I have one xmobar per screen (and I change the number of screens very often) that doesn't use dynamicLogString at all. And an event hook on top that keeps the trayer above xmobar.

Therefore I have no strong opinion about this. Just suggestions. ;-)

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +764,196 @@ xmobarStripTags tags = strip [] where
...
+--
+-- * 'statusBar'' relies on 'statusBarHandleConfig''
+--
+-- The main difference is that the functions returning a status bar config do not require a keybinding
+-- and use @mod-b@. Like the toggle action, this can be customized with 'withKey'.
+
+-- | This datataype abstracts a status bar to provide a common interface functions like 'statusBar'
+-- or 'statusBarProp'. Once defined, a status bar can be incorporated in 'XConfig' by using 'makeStatusBar',
+-- which takes care of all the necessary plumbing.
+data StatusBarConfig = StatusBarConfig { sbLogHook     :: X()
```
liskin, line 799:
```suggestion
data StatusBarConfig = StatusBarConfig { sbLogHook     :: X ()
```

`X()` feels weird. `()` is a type on its own that is a parameter to the `X` type constructor.

liskin (commented):
This is really nice piece of work.

I do have a few comments, though. I always do. :-)

XMonad/Hooks/DynamicLog.hs
TheMC47: It was actually different, the `StatusBarConfig` was the first parameter. @slotThe  and I discussed it a bit (https://github.com/xmonad/xmonad-contrib/pull/443#discussion_r556565592) 

The original `StatusBarConfig -> XConfig l -> IO (XConfig (ModifiedLayout AvoidStruts l))` resulted in constantly using `flip makeStatusBar conf` in the code, which was  bit weird.

Do you think we should add more documentation on how to use this with a config like yours? Or provide another more convenient function, just for a better UX? Both options seem valid to me

XMonad/Hooks/DynamicLog.hs
TheMC47: Before this PR I didn't use the `statusBar` for a similar reason; I found it weird that it was required to provide a keybinding that I don't have full control on its functionality. I also initially thought about implementing this as `Monoid` but the `keyToggle/toggleAction`  stuff was in the way.

One suggestion would be to change the interface of `StatusBarConf` so that instead of `keyToggle/toggleAction` the users would need to provide a `(XConfig Layout -> Map (ButtonMask, KeySym) (X ()))` and simply compose them when merging multiple status bars, instead of requiring a top level key/action pair. With some helper functions to make creating these maps easier, we could deprecate `statusBar` and friends in favor of the configs.

TheMC47: Regarding your config, you could create multiple `StatusBarConf`s and compose them, and override the `sbCleanupHook` to whatever you want. The problem would be that it's still not enough to react to dynamic screen changes I guess.  I'm planning to move on to `DynamicBars` after we decide on the interface here, and either rewrite it in terms of `StatusBarConf` or leave it and write a similar module. I thought about adding an event hook to the `StatusBarConf`, but I think it's gonna be left unused in `DynamicLog` and only be used in `DynamicBars`, so I didn't

XMonad/Hooks/DynamicLog.hs
liskin: Oh, I see. But I think the real problem wasn't the `flip`. The real problem was that you almost always had `IO StatusBarConfig` but `makeStatusBar` takes `StatusBarConfig`, so you had to wrestle with `>>=` or something. But this would've been fine, too:
```
do sb <- mySB
   xmonad =<< (makeStatusBar sb . ewmh . docks . … $ def{ … })
```
I mean, not totally fine, the IO in makeStatusBar is unnecessary, so it should really be:
```
do sb <- mySB
   xmonad $ makeStatusBar sb . ewmh . docks . … $ def{ … }
```

Or maybe keep the IO but use it for something: if all the functions that return StatusBarConfig return it in IO, then it's probably good UX to have makeStatusBar evaluate it. So then you have
```
xmonad =<< makeStatusBar mySB (ewmh . docks . def{ … })
```
You can do that even now with
```
xmonad =<< combineSBs [mySB] (ewmh . docks . def{ … })
```
but it discards the keybinding stuff so it's not a direct replacement. But that's another issue that I believe should be addressed. It's really be nice if these two were actually equivalent.

I haven't really thought this all through to the end, it needs some experimenting and maybe it's a dead end.

TheMC47: I wanted `makeStatusBar` to return a pure XConfig, but I'm not sure how to do that. The main problem is with logging with pipes. You would need to have access to the pipe so you can write to it, so essentially, you'd need to spawn the status bar before creating the config. 

I think changing the parameter to an `IO StatusBarConf` would provide a better interface

liskin: Yeah but you're not creating the pipes in `makeStatusBar` any more (or ever). `makeStatusBar` is literally `pure something`.

XMonad/Hooks/DynamicLog.hs
```diff
@@ -746,6 +764,196 @@ xmobarStripTags tags = strip [] where
...
+               -> StatusBarConfig
+combineSBConfs confs = def
+  { sbLogHook     = mapM_ sbLogHook confs
+  , sbStartupHook = mapM_ sbStartupHook confs
+  , sbCleanupHook = mapM_ sbCleanupHook confs
+  }
+
+-- | Like 'combineSBConfs', but expects the status bars to be wrapped in 'IO', and applies
+-- the combined status bars to the base config. This is useful to directly combine configs
+-- returned by 'statusBarHandleConfig', 'statusBarHandleConfig'' and 'statusBarPropConfig'.
```
liskin, line 950: It's also worth noting that if one chooses not to "directly combine configs" and instead uses `makeStatusBar` multiple times, the `ch1 *> sh1 *> ch2 *> sh2 *> …` sequence will result in killing of all but the last status bar.

Also, `AvoidStruts` will be applied to the layout multiple times, and eventhooks will run multiple times.

This is probably worth mentioning somewhere. But I'd postpone this until we deal with my other comments which may or may not results in changes to the API.

XMonad/Hooks/DynamicLog.hs
liskin: Yeah, `XConfig Layout -> Map (ButtonMask, KeySym) (X ())` is definitely better as it allows `const mempty`, i.e. no touching the (key)bindings at all.

What if someone wants to bind their mouse button 9 to toggle the struts? Then the best type is probably `XConfig l → XConfig l`. Or maybe at that point they should go with `const mempty` and add it to the rest of their custom keybindings.

So anyway, let's continue with that thought for a bit. All of StatusBarConfig can be just `XConfig l → XConfig l`. (Yeah, I know that's where we started, and I know I was the one who suggested records. Let's ignore that now.) The only reason we bother with the record in the first place is because we want to make it easier for people to change the keybinding and keep all the other defaults. Other customizations don't benefit from the StatusBarConfig abstraction much, I think, and it's often easier to just do all the plumbing by hand. Or… is it expected that someone would modify other fields of the StatusBarConfig record?

What I'm trying to say here is that we should explicitly think about and agree what usecases we're trying to make easier, what customizations we are trying to abstract over, etc. (And I do have to admit I got lost a bit. Maybe it was explicit somewhere and I missed it. Can you remind me if that is so?)

XMonad/Hooks/DynamicLog.hs
TheMC47: Right, yes, silly me :sweat_smile:  I think it's between
```haskell
do sb <- mySB
   xmonad $ makeStatusBar sb . ewmh . docks . … $ def{ … }
``` 
or
```haskell
xmonad =<< makeStatusBar mySB (ewmh . docks . def{ … })
```
The second one is more convenient to write, but it maybe is uglier since the `IO` is not really needed. Both seem valid to me, maybe we say what you or others prefer

XMonad/Hooks/DynamicLog.hs
TheMC47: That's a really good point, it slipped my mind to mention that in the docs

XMonad/Hooks/DynamicLog.hs
TheMC47: I think the main benefit of `StatusBarConfig` is that we can have composable bars. Whether we include the keybindings or not, as well as the layout modifier, is not really important. I only added it because that functionality was provided in `statusBar` and its variants, and I wanted to provide a replacement with the same functionality. 

I'm kinda tempted to make this another module that is a more user-friendly and more flexible version of `X.H.DynamicBars`

liskin: Keybindings and layout modifier make composability harder, so I'd say it's best to keep those somewhat separate. A version of `makeStatusBar` that only handles startup/cleanup and logging, but not keybindings/layout would probably be useful, and then maybe another that adds also the rest.

That thing with alternating startup/cleanup hooks leading to only the last status bar staying up makes me think that the composability here is not perfect yet. Right now `sbCleanupHook` is either mempty or `cleanupStatusBars`, and chaining multiple `cleanupStatusBars` is functionally equivalent to having just one `cleanupStatusBars`. Also, `cleanupStatusBars` is noop if nothing was started via `spawnStatusBarAndRemember`, so we might treat `cleanupStatusBars` the same as the layout: it should be there somewhere, but only once, doesn't need to compose with the rest. Therefore `StatusBarConfig` is now just a loghook and a startuphook, everything composes nicely, and it can be just an `XConfig → XConfig` or `XConfig → IO XConfig`. Not saying it should be, but it could be, and it's good to think about it this way, and then composability comes much more naturally. We might prefer a newtype, to prevent accidents, or keep the record so that we know what's what.

liskin: (Making this a separate module means we wouldn't need to bother with backward compat, but I wouldn't give up just yet. If the interface is good enough to eventually replace the implementation of the most simple API in DynamicLog, everyone benefits.)

TheMC47: I see your point, and I agree with separating the keybindings and the layout from the rest. 

However, while `sbCleanupHook` might not really look useful here, I was thinking about using it later to provide functionality similar to `DynamicBars`, by triggering a cleanup and a restart as a reaction to a change in the screen configuration. In that case, `sbCleanupHook` would probably be either a `cleanupStatusBars` or closing a handle, and chaining multiple cleanup hooks might be necessary since you really don't know what to do to each bar. I also think that using a record has the slight advantage that it might be easier for users to go beyond the defaults, and provide their own `sb*Hook`s

liskin: I see, that is a reasonable path forward.

Still, with bar-specific cleanups, the "alternating startup/cleanup hooks leading to only the last status bar staying up" problem doesn't exist. That also means that the ordering between startup and cleanup hooks doesn't really matter, and the startup hook can do the cleanup first. :-)

`cleanupStatusBars` as it is now doesn't fit into that unfortunately. It's nice and simple, though, so maybe what I'm saying is worse. Not sure, really. Maybe we can extend `StatusBarPIDs` with tags (e.g. using [Unique](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Unique.html#t:Unique)) so that we can optionally selectively restart individual bars.

About users going beyond the default, I am somewhat sceptical about what value they'd gain by using the record instead of just doing all the plumbing themselves.

TheMC47: Not sure if using `Unique` would work though, persisting state between the restarts requires a `Show` instance and `Unique` doesn't seem to provide that.

liskin: You're right, that wouldn't work. There's certainly some other way to make the tagging work. No need to worry about that now. :-)

TheMC47:
@liskin  I tried to add the suggested changes, it would be awesome if you could check it again!

I noticed that `statusBarProp`, `statusBarPropTo` and `statusBar'` are only in the git version, so I think we could afford to replace them with the new interface.

XMonad/Hooks/DynamicLog.hs
TheMC47: A simple workaround would be to have the cleanup only kills:

- `partialCleanup` when the screen configuration changes
- `fullCleanup` when restarting xmonad

this makes sense for both property-based and stdin-based loggers, as in the case of a partial cleanup, we can't kill stdin-based loggers with the current setup. I gave this some thought, and we probably can't get away with only one cleanup hook

liskin (commented):
I like the changes!

I added a few comments here and there and after doing all that I realized that you'd probably fix the docs later anyway and just wanted to check if the major changes you did are okay, and I think they are okay, so please do tidy it up and then come check my comments for anything you might have forgotten. :-)

XMonad/Hooks/DynamicLog.hs
```diff
@@ -745,6 +765,160 @@ xmobarStripTags tags = strip [] where
...
+-- Instead of directly modifying the config, the functions 'statusBarProp' and 'statusBar', create an
+-- appropriate 'StatusBarConfig' and call 'makeStatusBar' on it. This enables the user to have more
+-- flexibility and control over the status bars, without having do the plumbing themselves. It's worth
+-- mentioning that this approach tries to stay away from your keybindings, layout and other things,
+-- unlike using 'statusBar' and the similar functions. On the upside, it has the benefit of having a
+-- cleaner config and composable status bars.
+--
+-- For example, if you want to use property logging as provided by 'statusBarProp' you could do the following:
+--
+-- > mySB = statusBarPropConfig "xmobar" myPP
```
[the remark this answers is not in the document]
liskin: Oh nevermind, this was correct. Sorry. On the other hand maybe having `IO StatusBarConfig` in `makeStatusBar` makes using `<>` cumbersome?

XMonad/Hooks/DynamicLog.hs
```diff
@@ -745,6 +765,160 @@ xmobarStripTags tags = strip [] where
...
+-- > main = xmonad =<< makeStatusBar mySB myConf
+--
+-- This plays nicer with other combinators that you might have already in your config:
+--
+-- > mySB = statusBarPropConfig "xmobar" myPP
```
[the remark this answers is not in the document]
liskin: (nevermind this too)

XMonad/Hooks/DynamicLog.hs
TheMC47: We could drop `makeSBwithKey` if we decide to only leave `statusBar` and remove the other variants

liskin: But I think `makeSBwithKey` isn't too bad. It makes it easy for users to customize the bars a bit without touching the keys/layout plumbing. :-)

TheMC47: Since the naming is confusing and `makeSBwithKey` does more than just the key, I think it needs another name (maybe `makeStatusBar` and `makeStatusBar'`

XMonad/Hooks/DynamicLog.hs
```diff
@@ -745,6 +765,160 @@ xmobarStripTags tags = strip [] where
...
+
+-- $further customization
+-- Internally, the status bars are managed through the 'StatusBarConfig', which provides a convenient
+-- abstraction over what a status bar is and how to manage it.
+-- Instead of directly modifying the config, the functions 'statusBarProp' and 'statusBar', create an
+-- appropriate 'StatusBarConfig' and call 'makeStatusBar' on it. This enables the user to have more
+-- flexibility and control over the status bars, without having do the plumbing themselves. It's worth
+-- mentioning that this approach tries to stay away from your keybindings, layout and other things,
+-- unlike using 'statusBar' and the similar functions. On the upside, it has the benefit of having a
+-- cleaner config and composable status bars.
```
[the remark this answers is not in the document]
TheMC47: I've rewritten this, tell me if it's fine now

XMonad/Hooks/DynamicLog.hs
```diff
@@ -745,6 +765,160 @@ xmobarStripTags tags = strip [] where
...
+-- Like 'makeStatusBar', this composes nicely with other combinators:
+--
+-- > mySB = statusBarPropConfig "xmobar" myPP
+-- > main = xmonad =<< (combineSBs myBars . ewmh . docks $ def {...})
+--
+-- The above examples also work if the different status bars support different
+-- logging methods: you could do mix property logging and logging via standard input.
+-- One thing to keep in mind: if multiple bars read from the same property, their content
+-- will be the same. If you want to use property-based logging with multiple bars,
+-- they should read from different properties.
```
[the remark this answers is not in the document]
TheMC47: I really like the concept of property-based logging, but having more than one status bar with that isn't that pleasant to use

TheMC47:
For some reason, I can't comment on your last comment, but I think I resolved everything now. Thanks again for the review!

liskin:
Summary of the next steps we agreed on IRC:

* move the StatusBarConfig stuff to XMonad.Hooks.StatusBar
* move PP and dynamicLogWithPP to a separate module as well (name undecided, perhaps XMonad.Hooks.DynamicLog.PP? better ideas welcome) so that X.H.StatusBar can use it and X.H.DynamicLog can import X.H.StatusBar to provide the old wrappers like `dzen` and `statusBar`
* X.H.DynamicLog reexports this new PP module
* `git show v0.16:XMonad/Hooks/DynamicLog.hs` is the level of compatibility we wish to keep, all the new `statusBar'` and `statusBarProp` functions can be moved to X.H.StatusBar and possibly cleaned up (we don't need two variants of each where one takes `PP` and the other `X PP`, the latter is sufficient)
* the docs and CHANGES might need some updating as well
* #463 then goes into X.H.StatusBar as well, so we'll have one module with composable and possibly dynamic status bars, and that will be awesome
* #450 should then be a separate module built upon X.H.D.PP only (out of scope here, just related)

PR #446, opened 2021-01-19
Generate CI with the haskell-ci utility.

peti:
I believe that those CI build instructions can co-exist peacefully with the stack-based build that we currently have.


psibi:
@peti Can you enable LTS based builds too or should I create a PR enabling them back ?

peti:
I don't know how to re-enable those LTS builds. The current CI builds assume that one `stack.yaml` file can build all LTS versions, but that is not really true. We need distinct `stack-lts-x.yaml` files for every LTS version `x` we want to support so that we can adapt the set of required `extra-packages` etc. I am happy to provide those `stack.yaml` files, but I don't know how to setup the CI builds such that it uses them in a way that interacts correctly with the build cache.

liskin:
> The current CI builds assume that one stack.yaml file can build all LTS versions, but that is not really true. We need distinct stack-lts-x.yaml files for every LTS version x we want to support so that we can adapt the set of required extra-packages etc.

Do we really need to have this extra complexity for a tool that you said is only meant to be used by developers? Adding multiple stack.yaml files with various sets of hand-picked extra-deps is not going to help with Debian packaging, which is one of the things that started this whole mess. Also, we didn't get a reply from the Debian folks yet, so it's entirely possible that they don't need to rebuild the manpage at all.

If you insist that depending on a particular version of pandoc is necessary and worth the maintenance costs (I am still not convinced about this, but you apparently decided not to discuss this further, unfortunately), why don't you build (and run!) the generatemanpage utility in the cabal-based workflow only and revert the stack-based workflow to only build/test xmonad without the utility?

psibi:
>  I am happy to provide those stack.yaml files, but I don't know how to setup the CI builds such that it uses them in a way that interacts correctly with the build cache.

No worries, I will create a PR fixing it. As liskin has said above, we don't really need different stack.yaml files.

>  I never touched the LTS builds here in the contrib repository. I only switched CI to Nightly in xmonad proper

Yeah, I replied in the wrong repository. :-) I will fix it in the proper repo.

peti:
> As liskin has said above, we don't really need different stack.yaml files.

We cannot build `generatemanpage` without some customization. A single file for all kinds of LTS builds is not going to cut it.

peti:
> We didn't get a reply from the Debian folks yet, so it's entirely possible that they don't need to rebuild the manpage at all.

Frankly, this whole Debian packaging thing is a complete non-issue to me. I don't understand why you seem to obsess over it.

> you apparently decided not to discuss this further, unfortunately

I have written a long and detailed response to your PR about this subject and I have mentioned a number of points that suggest changing the way `generatemanpage` depends on pandoc might be a bad idea. Unfortunately, you never addressed any of those points and instead simply repeated what you had said in the original posting already. I did not observe any willingness to understand and consider my point of view, therefore I chose not to pursue the discussion since it's gonna be a waste of time.

psibi:
> We cannot build generatemanpage without some customization. A single file for all kinds of LTS builds is not going to cut it.

Then probably I can do that customisation to make the builds pass ? But my first step is to enable lts builds itself so that we test for different resolvers.

liskin:
> I did not observe any willingness to understand and consider my point of view

This feeling is mutual:

> Frankly, this whole Debian packaging thing is a complete non-issue to me. I don't understand why you seem to obsess over it.

… :-/

peti:
Dude, I *HAVE* explained my reasoning for this position in great detail. If you don't want to hear my reasons then that's fine, but claiming that I would be unwilling to engage with your point of view is bullshit.

PR #448, opened 2021-01-22
More loggers: A `maybe`-like combinator, and screen-specific loggers

TheMC47:
### Description

I added some more loggers:
- For multi-monitor setups:
  - `logWhenActive` to have loggers active only when a certain screen is active.
  - `logTitleOnScreen`, `logLayoutOnScreen` and `logCurrentOnScreen` as screen-specific versions of `logTitle`, `logLayout` and `logCurrent`
- A "combinator": `logMaybe` which works like `maybe` (and an infix version for it `(.|) `
- `logConst` to log a constant `String` (I find that it feels better than writing `return . Just` repeatedly, but I didn't change it in the file... maybe see first what people feel about it?)
- Minor cleanups (documentation fix, one `hlint` suggestion)

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -304,6 +304,12 @@
 
   * `XMonad.Util.Loggers`
     - Make `battery` and `loadAvg` distro-independent.
+    - Added `logTitleOnScreen`, `logCurrentOnScreen` and `logLayoutOnScreen`
+      as screen-specific variants of `logTitle`, `logCurrent` and `logLayout`.
+    - Added `logWhenActive` to have loggers active only when a certain
+      screen is active.
+    - Added `logConst` to log a constant `String`, and `logDefault` (infix: `.|`)
+      to combine loggers.
 
   * `XMonad.Hooks.DynamicLog`
     - Added `xmobarBorder` function to create borders around strings.
--- a/XMonad/Util/Loggers.hs
+++ b/XMonad/Util/Loggers.hs
@@ -32,7 +32,11 @@ module XMonad.Util.Loggers (
     -- * XMonad Loggers
     -- $xmonad
     , logCurrent, logLayout, logTitle
-
+    , logConst, logDefault, (.|)
+    -- * XMonad: Screen-specific Loggers
+    -- $xmonad-screen
+    , logCurrentOnScreen, logLayoutOnScreen
+    , logTitleOnScreen, logWhenActive
     -- * Formatting Utilities
     -- $format
     , onLogger
@@ -43,7 +47,7 @@ module XMonad.Util.Loggers (
 
   ) where
 
-import XMonad (liftIO)
+import XMonad (liftIO, Window)
 import XMonad.Core
 import qualified XMonad.StackSet as W
 import XMonad.Hooks.DynamicLog
@@ -51,7 +55,7 @@ import XMonad.Util.Font (Align (..))
 import XMonad.Util.NamedWindows (getName)
 
 import Control.Exception as E
-import Data.List (isPrefixOf, isSuffixOf)
+import Data.List (find, isPrefixOf, isSuffixOf)
 import Data.Maybe (fromMaybe)
 import System.Directory (getDirectoryContents)
 import System.IO
@@ -125,7 +129,7 @@ battery = logCmd "acpi | sed -r 's/.*?: (.*%).*/\\1/; s/[dD]ischarging, ([0-9]+%
 --   For more information see something like
 --   <http://www.cplusplus.com/reference/clibrary/ctime/strftime.html>.
 date :: String -> Logger
-date fmt = io $ do cal <- (getClockTime >>= toCalendarTime)
+date fmt = io $ do cal <- getClockTime >>= toCalendarTime
                    return . Just $ formatCalendarTime defaultTimeLocale fmt cal
 
 -- | Get the load average.  This assumes that you have a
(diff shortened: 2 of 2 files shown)


XMonad/Util/Loggers.hs
```diff
@@ -182,6 +186,72 @@ logLayout = withWindowSet $ return . Just . ld
 logCurrent :: Logger
 logCurrent = withWindowSet $ return . Just . W.currentTag
 
+-- | Log the given string, as is.
+logConst :: String -> Logger
+logConst = return . Just
+
+-- | Like they 'maybe' combinator, but for  loggers. If the given logger
```
slotThe, line 193: I suppose this is more like `fromMaybe` instead of `maybe` (at least signature wise)?  Also there's a double space between "for" and "loggers"

XMonad/Util/Loggers.hs
```diff
@@ -182,6 +186,72 @@ logLayout = withWindowSet $ return . Just . ld
...
+-- gives @Nothing@, the default logger is returned.
+-- For example, to display a quote when no windows are on the screen,
+-- you can do:
+--
+-- > logMaybe logTitle (logConst "Hey, you, you're finally awake.")
+logMaybe :: Logger -> Logger -> Logger
+logMaybe l d = l >>= maybe d logConst
```
slotThe, line 200: I think, since you're explicitly mentioning `fromMaybe`, these arguments should be the same as that function, i.e. the default first and then the thing you're looking at

XMonad/Util/Loggers.hs
```diff
@@ -182,6 +186,72 @@ logLayout = withWindowSet $ return . Just . ld
...
+-- | Get the title (name) of the focused window, on the given screen.
+logTitleOnScreen :: ScreenId -> Logger
+logTitleOnScreen =
+  withScreen
+    $ traverse (fmap show . getName)
+    . (W.focus <$>)
+    . W.stack
+    . W.workspace
+
+-- | Get the name of the current workspace on the given screen.
```
slotThe, line 234: "current workspace" gave me a small pause, but that might be just me; maybe "visible workspace"?

XMonad/Util/Loggers.hs
```diff
@@ -182,6 +186,72 @@ logLayout = withWindowSet $ return . Just . ld
...
+logCurrentOnScreen = withScreen $ logConst . W.tag . W.workspace
+
+-- | Get the name of the current layout on the given screen.
+logLayoutOnScreen :: ScreenId -> Logger
+logLayoutOnScreen =
+  withScreen $ logConst . description . W.layout . W.workspace
+
+-- | A screen dependent 'Logger', which is just a function that takes
+-- a Screen and returns a 'Logger'
+type ScreenLogger = W.Screen WorkspaceId (Layout Window) Window ScreenId ScreenDetail -> Logger
```
slotThe, line 245: This is a very small detail, but perhaps, to keep in line with the [rest of xmonad](https://hackage.haskell.org/package/xmonad-0.15/docs/XMonad-Core.html#t:WindowSet), we should be doing 

``` haskell
type WindowScreen = Screen WorkspaceId (Layout Window) Window ScreenId ScreenDetail
```

and then

``` haskell
withScreen :: (WindowScreen -> Logger) -> ScreenId -> Logger
```

slotThe (approved):
This looks good to me!  Just a few doc nits, as usual :>

CHANGES.md
```diff
@@ -304,6 +304,12 @@
 
   * `XMonad.Util.Loggers`
     - Make `battery` and `loadAvg` distro-independent.
+    - Added `logTitleOnScreen`, `logCurrentOnScreen` and `logLayoutOnScreen`
+      as screen-specific variants of `logTitle`, `logCurrent` and `logLayout`.
+    - Added `logWhenActive` to have loggers active only when a certain
+    screen is active.
```
liskin, line 310:
```suggestion
      screen is active.
```

XMonad/Util/Loggers.hs
liskin: Alternatively it could be named logFallback or logDefault or something and the order can stay.

XMonad/Util/Loggers.hs
```diff
@@ -182,6 +186,72 @@ logLayout = withWindowSet $ return . Just . ld
...
+-- It is also possible to bind loggers like 'logTitle' to a specific screen. For
+-- example, using @logTitleOnScreen 1@ will log the title of the focused window
+-- on screen 1, even if screen 1 is not currently active.
+
+-- | Only display the 'Logger' if the screen with the given 'ScreenId' is
+-- active.
+-- For example, this can be used to create a marker for the active
+-- screen:
+--
+-- > logWhenActive 0 (logConst "*")
```
liskin, line 219: I don't understand the example. :-(
Screen 0 is not necessarily the active one, is it?
I thought it's the first screen reported by Xinerama, so either the one on primary output or the first output.

XMonad/Util/Loggers.hs
```diff
@@ -182,6 +186,72 @@ logLayout = withWindowSet $ return . Just . ld
...
+
+-- | A screen dependent 'Logger', which is just a function that takes
+-- a Screen and returns a 'Logger'
+type ScreenLogger = W.Screen WorkspaceId (Layout Window) Window ScreenId ScreenDetail -> Logger
+
+-- | A helper function to create screen-specific loggers.
+withScreen :: ScreenLogger -> ScreenId -> Logger
+withScreen f n = do
+  ss <- withWindowSet $ return . W.screens
+  case filter ((== n) . W.screen) ss of
```
liskin, line 251: https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-List.html#v:find might be useful here

XMonad/Util/Loggers.hs
TheMC47: The comment is confusing. Basically, if screen 0 (the primary one) is active, the logger will return a "*".  I'll update it

XMonad/Util/Loggers.hs
TheMC47: Not quite, a `fromMaybe` unwraps what's inside the `Maybe` with a fallback, we don't want it to be unwrapped 

I guess the name is confusing, I'll go with one of the suggestions of @liskin

TheMC47:
@liskin  @slotThe  Thank you for the reviews! I'll try to resolve the comments ASAP

XMonad/Util/Loggers.hs
slotThe: I have no qualms about the order as such, but when another function is references as essentially "this is like that" I think the order of arguments should be the same :) 

How it is right now is completely fine!

XMonad/Util/Loggers.hs
```diff
@@ -182,6 +186,70 @@ logLayout = withWindowSet $ return . Just . ld
...
+    $ traverse (fmap show . getName)
+    . (W.focus <$>)
+    . W.stack
+    . W.workspace
+
+-- | Get the name of the visible workspace on the given screen.
+logCurrentOnScreen :: ScreenId -> Logger
+logCurrentOnScreen = withScreen $ logConst . W.tag . W.workspace
+
+-- | Get the name of the visible layout on the given screen.
```
slotThe, line 237: Funnily enough, while "current workspace" -> "visible workspace" felt very natural to me, "current layout" -> "visible layout" feels wrong.  A layout is not really "visible" per se, is it?

geekosaur: I'd agree. "Layout of the workspace visible on the given screen" is a mouthful but consistent with how we normally refer to such things.

TheMC47: I guess it does, I updated it

CHANGES.md
```diff
@@ -304,6 +304,12 @@
 
   * `XMonad.Util.Loggers`
     - Make `battery` and `loadAvg` distro-independent.
+    - Added `logTitleOnScreen`, `logCurrentOnScreen` and `logLayoutOnScreen`
+      as screen-specific variants of `logTitle`, `logCurrent` and `logLayout`.
+    - Added `logWhenActive` to have loggers active only when a certain
+      screen is active.
+    - Added `logConst` to log a constant `String`, and `logMaybe` (infix: `.|`)
+    to combine loggers.
```
slotThe, line 312:
```suggestion
    - Added `logConst` to log a constant `String`, and `logMaybe` (infix: `.|`)
      to combine loggers.
```

slotThe:
This seems fine to me now and all other objections also seem to have been fixed; will merge in a few days if there are no further objections

CHANGES.md
```diff
@@ -304,6 +304,12 @@
 
   * `XMonad.Util.Loggers`
     - Make `battery` and `loadAvg` distro-independent.
+    - Added `logTitleOnScreen`, `logCurrentOnScreen` and `logLayoutOnScreen`
+      as screen-specific variants of `logTitle`, `logCurrent` and `logLayout`.
+    - Added `logWhenActive` to have loggers active only when a certain
+      screen is active.
+    - Added `logConst` to log a constant `String`, and `logMaybe` (infix: `.|`)
```
liskin, line 311:
```suggestion
    - Added `logConst` to log a constant `String`, and `logDefault` (infix: `.|`)
```

XMonad/Util/Loggers.hs
```diff
@@ -182,6 +186,70 @@ logLayout = withWindowSet $ return . Just . ld
...
+
+-- | If the first logger returns @Nothing@, the default logger is used.
+-- For example, to display a quote when no windows are on the screen,
+-- you can do:
+--
+-- > logDefault logTitle (logConst "Hey, you, you're finally awake.")
+logDefault :: Logger -> Logger -> Logger
+logDefault l d = l >>= maybe d logConst
+
+-- | An infix operator for 'logMaybe', which can be more convenient to
```
liskin, line 201:
```suggestion
-- | An infix operator for 'logDefault', which can be more convenient to
```

liskin (approved):
A few leftovers but otherwise good to go.

TheMC47:
Awesome, thank you all for the review!

PR #453, opened 2021-01-28
Remove last occurences of non Data.Default defaults

slotThe:
### Description

1. Remove mentions of `defaultConfig`, as this was removed in favour of `def` a while ago.
2. Deprecate `X.A.TreeSelect.tsDefaultConfig` in favour of `def`, as this was apparently forgotten.

### Checklist

  This didn't happen with other `Data.Default` deprecations, so I don't think it's necessary in this case either.

### DIFF
--- a/XMonad/Actions/TreeSelect.hs
+++ b/XMonad/Actions/TreeSelect.hs
@@ -40,6 +40,7 @@ module XMonad.Actions.TreeSelect
 
     , TSConfig(..)
     , tsDefaultConfig
+    , def
 
       -- * Navigation
       -- $navigation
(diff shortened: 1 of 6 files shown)


XMonad/Actions/TreeSelect.hs
```diff
@@ -258,6 +259,7 @@ defaultNavigation = M.fromList
 -- Using nice alternating blue nodes
 tsDefaultConfig :: TSConfig a
 tsDefaultConfig = def
+{-# DEPRECATED tsDefaultConfig "Use def (from Data.Default, and re-exported by XMonad.Actions.TreeSelect) instead." #-}
```
liskin, line 262: Isn't the reexport from `XMonad` enough?

slotThe: Every module that has a `Data.Default` instance exports this again.  I wasn't even aware that `XMonad` re-exports it (it's quite well hidden, via `XMonad.Config`s re-export of `Default(...)`).  

I'm fine with removing the export, but if we are going for consistency then I think we'd have to remove it from all other modules as well.

liskin: Oh, I grepped for "re-exported" and missed the other re-exports then.
Let's keep consistency then.

slotThe: The grep for `re-exported` doesn't return anything because all of these deprecations were removed in 5140f5b5d06790e055eb7fb0cf3eccc4997aa736.  If you go to that commit you'll see that all of these modules had a deprecations message just like that and all of them exported `def` as well

PR #460, opened 2021-02-01
X.H.RescreenHook: New module (custom rescreen and randr event hooks)

liskin:
### Description

Inspired by
https://github.com/fis/dot-xmonad/blob/a8c8e804873f8657fc1f1a0d93ee3f58f7a56e22/zem/src/Zem/AutoRandr.hs
and my own
https://github.com/liskin/dotfiles/blob/43999573d006311292a9fadde02f3b234140d999/.xmonad/xmonad.hs#L308-L339,
this adds a generic hook module that allows attaching a custom hook to
rescreen and randr screen change notify event, hiding the complexity of
suppressing duplicate events (debouncing) and distinguishing actual
xrandr configuration changes from output connects/disconnects.

#### [X.H.{EwmhDesktops,ManageDocks}: Improve the usage doc a bit](../commit/0610a3c5a064f967402051ffe94f8c289cf1ba36)

Don't assume ewmh/docks are the only xmonad config combinator out there.

#### [X.H.RescreenHook: Merge the two hooks together and improve their behaviour](../commit/d106ea24a98a5f5d9ad6d0d0ed5ed39db96cedae)

Now that randrChangeHook is only invoked for changes that don't result
in rescreen, it can actually be used for autorandr.

### Checklist

### DIFF
--- a/XMonad/Hooks/Rescreen.hs
+++ b/XMonad/Hooks/Rescreen.hs
@@ -90,7 +90,9 @@ instance Monoid RescreenConfig where
 -- Note that 'rescreenHook' is safe to use several times, 'rescreen' is still
 -- done just once and hooks are invoked in sequence, also just once.
 rescreenHook :: RescreenConfig -> XConfig a -> XConfig a
-rescreenHook = flip XC.once rescreenHook'
+rescreenHook = XC.once $ \c -> c
+    { startupHook = startupHook c <> rescreenStartupHook
+    , handleEventHook = handleEventHook c <> rescreenEventHook }
 
 -- | Shortcut for 'rescreenHook'.
 addAfterRescreenHook :: X () -> XConfig a -> XConfig a
@@ -99,10 +101,6 @@ addAfterRescreenHook h = rescreenHook def{ afterRescreenHook = h }
 -- | Shortcut for 'rescreenHook'.
 addRandrChangeHook :: X () -> XConfig a -> XConfig a
 addRandrChangeHook h = rescreenHook def{ randrChangeHook = h }
-
-rescreenHook' :: XConfig a -> XConfig a
-rescreenHook' c = c{ startupHook = startupHook c <> rescreenStartupHook
-                   , handleEventHook = handleEventHook c <> rescreenEventHook }
 
 -- | Startup hook to listen for @RRScreenChangeNotify@ events.
 rescreenStartupHook :: X ()
--- a/XMonad/Hooks/Rescreen.hs
+++ b/XMonad/Hooks/Rescreen.hs
@@ -1,3 +1,6 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE RecordWildCards #-}
 -- |
 -- Module      :  XMonad.Hooks.Rescreen
 -- Copyright   :  (c) 2021 Tomáš Janoušek <tomi@nomi.cz>
@@ -9,15 +12,13 @@
 module XMonad.Hooks.Rescreen (
     -- * Usage
     -- $usage
+    RescreenConfig(..),
     rescreenHook,
+    rescreenStartupHook,
     rescreenEventHook,
-    randrHook,
-    randrEventHook,
-    randrStartupHook,
     ) where
 
-import Control.Monad.Fix (fix)
-import Control.Monad (when)
+import Control.Monad (void)
 import Data.Monoid (All(..))
 
 import Graphics.X11.Xrandr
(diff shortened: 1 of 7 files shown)


liskin:
Ping @fis

liskin:
Somewhat related: https://github.com/liskin/xmonad/commit/839d0d94788efaa0287326ca514d42259ec23def
I have no idea if that is actually necessary, though. Probably not, but it could be if someone uses https://hackage.haskell.org/package/X11/docs/Graphics-X11-Xrandr.html to get more information about the situation, like https://github.com/fis/dot-xmonad/blob/a8c8e804873f8657fc1f1a0d93ee3f58f7a56e22/zem/src/Zem/AutoRandr.hs does.

fis:
Heh. It may take me a day or three to actually get around to it, but I can definitely try to test my AutoRandr code on top of these hooks.

Regarding the debouncing logic, some context: my setup involves two monitors connected over a single (USB-C) cable with DisplayPort MST. If memory serves, to X plugging in the cable looks like two consecutive (but distinct) configuration changes: first a single monitor appears at the "DP-0" output, and then that changes into a DP "hub", with two monitors connected as "DP-0-1" and "DP-0-8". I think I added the sleep-based debouncing as an initial attempt to work around that. That said, I don't think it really was stable enough (the timing is unpredictable), and I ended up instead just not configuring any action for the "single external monitor" configuration. So it's quite possible just clearing the actually immediate X events would work well enough (for my purposes).

liskin:
Absolutely no need to rush it! :-)

I also have a DP MST hub here (Lenovo ThinkPad Ultra Dock) and I had all sorts of problems with it in the past, but right now it seems to work rather well. Your setup is probably different, though, as I guess one of your monitors acts as the MST hub, but it really could look like a single monitor for a brief moment. Not sure if this is something we can reliably handle in xmonad. We could probably add some configurable sleep to the event clearing or something to make sure the hook really fires only once it's settled. Or we can leave it up to users to add sleep and some actual debouncing to the hook if they need that.

In my own config, I added flock to my bash layout-auto: https://github.com/liskin/dotfiles/commit/706db34614194d60329435c6457bb895e7467b54#diff-c57c136cf207a3cd44f4fac23bde2146a5f8259c4391333981aec1a8aba66898 and now I can just freely add sleeps if they ever become needed. And my configuration also won't reconfigure xrandr for unknown monitor configuration, I still need to manually press a key for that.

XMonad/Hooks/Rescreen.hs
```diff
@@ -0,0 +1,146 @@
...
+    _ <- clearTypedWindowRREvents (ev_window e) rrScreenChangeNotify
+    -- If there were any ConfigureEvents, this is an actual screen
+    -- configuration change, so rescreen and fire rescreenHook. Otherwise,
+    -- this is just a connect/disconnect, fire randrChangeHook.
+    if ev_event_type e == configureNotify || moreConfigureEvents
+        then rescreen >> afterRescreenHook
+        else randrChangeHook
+
+-- | Remove all X events of a given window and type from the event queue,
+-- return whether there were any.
```
slotThe, line 123: This looks really cool; not only in the context of this module but in general even.  Do we have a place where these kinds of "XUtils" (sadly `XMonad.Util.XUtils` does not seem to be this place) can go?

geekosaur: Not currently. Sadly evicting and reclaiming `XMonad.Util.XUtils` would be a breaking change; but it's frankly misnamed.

That said, maybe it deserves to be in `XMonad.Util.EvUtils` (this doesn't exist yet).

liskin: What's wrong with `XMonad.Util.XUtils`?

On the other hand: should I really try to move a function that nobody needed for a decade somewhere where it's only slightly easier to find than where it is? After all, if anyone opens a PR that adds a similar function for their module, surely I'll notice and suggest factoring it out then. :-)

liskin: (Anyway, if we ever make the function available, it should come with some warnings as it may change the order of events. It does put them back in the correct order relative to one another, but the order is changed relative to events that weren't picked up by `XCheckTypedWindowEvent`. It is extremely unlikely to matter in this usecase, but with something else than `ConfigureEvent`, it may matter.)

slotThe: > What's wrong with `XMonad.Util.XUtils`?

The module description says "A module for painting on the screen" and that's pretty much what all of the functions do in some capacity; not really a place for general X utility functions.  Then again this is only there as a doc, so we may well repurpose it and give all the drawing stuff a heading of its own.

> On the other hand: should I really try to move a function that nobody needed for a decade somewhere where it's only slightly easier to find than where it is? After all, if anyone opens a PR that adds a similar function for their module, surely I'll notice and suggest factoring it out then. :-)

I'll give you that one; probably a bit of a premature optimization, especially since no good module already exists where we could put it.

liskin: > The module description says "A module for painting on the screen" and that's pretty much what all of the functions do in some capacity; not really a place for general X utility functions.

There are also a couple of X wrappers for IO functions with Display parameter. We can move those under a separate haddock heading and then add another heading for utilities (which I define as more than a wrapper) like my clearTypedWindowEvents. None of that needs to be done now, but I think this is the most sensible thing to do if we want a module for X utils.

XMonad/Hooks/Rescreen.hs
```diff
@@ -0,0 +1,165 @@
...
+-- 'afterRescreenHook' is useful for restarting/repositioning status bars and
+-- systray.
+--
+-- 'randrChangeHook' may be used to automatically trigger xrandr (or perhaps
+-- autorandr) when outputs are (dis)connected.
+--
+-- Note that 'rescreenHook' is safe to use several times, 'rescreen' is still
+-- done just once and hooks are invoked in sequence, also just once.
+rescreenHook :: RescreenConfig -> XConfig a -> XConfig a
+rescreenHook = flip XC.once rescreenHook'
```
liskin, line 96: Perhaps `XC.once` itself should be flipped…

liskin:
Two things are still missing here:

* a check that XineramaScreenInfo actually changed to prevent some unnecessary rescreens: https://github.com/xmonad/xmonad/issues/243#issuecomment-723470599
* a check that connected outputs actually changed to prevent some unnecessary invocations of `randrChangeHook` (e.g. after VT switch)

Even without these, it should be useful for many people and should be good enough to be used by default in `X.H.StatusBar`, so I'd like to get it merged soon.

TheMC47 (approved):
This is working beautifully with X.H.StatusBar! Good job

I really like the safety that the ExtensibleConfig gives us. LGTM

PR #390, opened 2020-10-14
Export clickableWrap

ivanbrennan:
### Description

Export `clickableWrap` from XMonad.Util.ClickableWorkspaces so users can more easily compose their own clickable pretty-printers.

Configs that apply WorkspaceId transformations, such as IndependentScreens (adding/removing a screen-number prefix) and NamedWorkspaces (adding/removing a name suffix), cannot use `clickablePP` as is, since they need to apply `clickableWrap` to an appropriately transformed WorkspaceId. Rather than force them to re-implement `clickableWrap`, export it.

An example use-case, where IndependentScreens has added a screen number prefix to the workspace ids (0_1, 0_2, ...), and we want the status-bar to show ids sans screen number (1, 2, ...) _and_ make them clickable:
```haskell
import XMonad.Util.ClickableWorkspaces (clickableWrap)

getClickable :: (WorkspaceId -> WorkspaceId) -> X (WorkspaceId -> String)
getClickable f = do
  wsIndex <- getWsIndex
  return $ \ws -> case wsIndex (f ws) of
                    Just idx -> clickableWrap idx ws
                    Nothing -> ws

composePP :: PP -> ScreenId -> X PP
composePP pp s = do
  clickable <- getClickable (marshall s)
  return
    . marshallPP s
    $ pp
      { ppCurrent         = ppCurrent         pp . clickable,
        ppVisible         = ppVisible         pp . clickable,
        ppHidden          = ppHidden          pp . clickable,
        ppHiddenNoWindows = ppHiddenNoWindows pp . clickable,
        ppUrgent          = ppUrgent          pp . clickable
      }
```

### Checklist

  - [ ] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

  - [ ] I updated the `CHANGES.md` file

  - [ ] I updated the `XMonad.Doc.Extending` file (if appropriate)

PR #462, opened 2021-02-03
X.U.ClickableWorkspaces: Integrations with X.L.IndependentScreens and X.A.WorkspaceNames

liskin:
### Description

Having these work together isn't entirely trivial (see my confusion in
https://github.com/xmonad/xmonad-contrib/pull/390) so let's provide the
integrations.

Also, switch from unmaintained wmctrl to xdotool (rationale in commit
msg).

### Checklist

    (I didn't, I think the entry for X.U.ClickableWorkspaces itself it sufficient.)

    (I didn't, I think the entry for X.U.ClickableWorkspaces itself it sufficient.)

### DIFF
--- a/XMonad/Util/ClickableWorkspaces.hs
+++ b/XMonad/Util/ClickableWorkspaces.hs
@@ -46,11 +46,20 @@ import XMonad.Util.WorkspaceCompare (getWsIndex)
 --   * "XMonad.Hooks.EwmhDesktops" for @xdotool@ support (see Hackage docs for setup)
 --   * use of UnsafeStdinReader/UnsafeXMonadLog in xmobarrc (rather than StdinReader/XMonadLog)
 
-
+-- | Wrap string with an xmobar action that uses @xdotool@ to switch to
+-- workspace @i@.
 clickableWrap :: Int -> String -> String
 clickableWrap i ws = xmobarAction ("xdotool set_desktop " ++ show i) "1" $ xmobarRaw ws
 
--- | Use index of workspace in users config to target workspace with @xdotool@ switch.
+-- | Return a function that wraps workspace names in an xmobar action that
+-- switches to that workspace. That workspace name must be exactly as
+-- configured in 'XMonad.Core.workspaces', so this takes an additional
+-- parameter that allows renaming/marshalling of the name for display, which
+-- is applied after the workspace's index is looked up.
+--
+-- This additionally assumes that 'XMonad.Hooks.EwmhDesktops.ewmhDesktopsEventHook'
+-- isn't configured to change the workspace order. We might need to add an
+-- additional parameter if anyone needs that.
 getClickable :: (WorkspaceId -> String) -> X (WorkspaceId -> String)
 getClickable ren = do
   wsIndex <- getWsIndex
--- a/XMonad/Util/ClickableWorkspaces.hs
+++ b/XMonad/Util/ClickableWorkspaces.hs
@@ -18,12 +18,20 @@ module XMonad.Util.ClickableWorkspaces (
   -- * Usage
   -- $usage
   clickablePP,
-  clickableWrap
+  clickableRenamedPP,
+  clickableWrap,
+
+  -- * Integrations
+  clickableWorkspaceNamesPP,
+  clickableMarshallPP,
+  clickableMarshallWorkspaceNamesPP
   ) where
 
 import XMonad
-import XMonad.Util.WorkspaceCompare (getWsIndex)
+import XMonad.Actions.WorkspaceNames
 import XMonad.Hooks.DynamicLog (xmobarAction, xmobarRaw, PP(..))
+import XMonad.Layout.IndependentScreens
+import XMonad.Util.WorkspaceCompare (getWsIndex)
 
 -- $usage
 -- However you have set up your PP, apply @clickablePP@ to it, and bind the result
@@ -41,17 +49,24 @@ clickableWrap :: Int -> String -> String
 clickableWrap i ws = xmobarAction ("xdotool set_desktop " ++ show i) "1" $ xmobarRaw ws
 
 -- | Use index of workspace in users config to target workspace with @xdotool@ switch.
-getClickable :: X (WorkspaceId -> String)
-getClickable = do
+getClickable :: (WorkspaceId -> String) -> X (WorkspaceId -> String)
+getClickable ren = do
   wsIndex <- getWsIndex
   return $ \ws -> case wsIndex ws of
-                    Just idx -> clickableWrap idx ws
+                    Just idx -> clickableWrap idx (ren ws)
                     Nothing -> ws
 
 -- | Apply clickable wrapping to all workspace fields in given PP.
 clickablePP :: PP -> X PP
-clickablePP pp = do
-  clickable <- getClickable
+clickablePP = clickableRenamedPP id
+
+-- | Alternative to 'clickablePP' that allows changing the visible workspace
+-- name. Useful for integration with modules that change workspace names, such
+-- as "XMonad.Layout.IndependentScreens" and "XMonad.Actions.WorkspaceNames".
+-- See "XMonad.Util.ClickableWorkspaces.Integrations".
+clickableRenamedPP :: (WorkspaceId -> String) -> PP -> X PP
+clickableRenamedPP ren pp = do
+  clickable <- getClickable ren
   return $
     pp { ppCurrent         = ppCurrent pp . clickable
        , ppVisible         = ppVisible pp . clickable
@@ -59,3 +74,21 @@ clickablePP pp = do
        , ppHiddenNoWindows = ppHiddenNoWindows pp . clickable
        , ppUrgent          = ppUrgent pp . clickable
        }
+
+-- | Integration with "XMonad.Actions.WorkspaceNames".
+clickableWorkspaceNamesPP :: PP -> X PP
+clickableWorkspaceNamesPP pp = do
+    rename <- getWorkspaceNames
+    clickableRenamedPP rename pp
+
+-- | Integration with "XMonad.Layout.IndependentScreens".
+clickableMarshallPP :: ScreenId -> PP -> X PP
+clickableMarshallPP s pp =
+    clickableRenamedPP unmarshallW pp{ ppSort = marshallSort s <$> ppSort pp }
+
+-- | Integration with both "XMonad.Actions.WorkspaceNames" and
+-- "XMonad.Layout.IndependentScreens".
+clickableMarshallWorkspaceNamesPP :: ScreenId -> PP -> X PP
+clickableMarshallWorkspaceNamesPP s pp = do
+    rename <- getWorkspaceNames
+    clickableRenamedPP (unmarshallW . rename) pp{ ppSort = marshallSort s <$> ppSort pp }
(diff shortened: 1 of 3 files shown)


XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -32,26 +33,33 @@ import XMonad.Hooks.DynamicLog (xmobarAction, xmobarRaw, PP(..))
 -- > logHook = clickablePP xmobarPP { ... } >>= dynamicLogWithPP
 --
 -- * Requirements:
---   * wmctrl on system (in path)
+--   * @xdotool@ on system (in path)
 --   * "XMonad.Hooks.EwmhDesktops" for wmctrl support (see Hackage docs for setup)
```
slotThe, line 37:
```suggestion
--   * "XMonad.Hooks.EwmhDesktops" for @xdotool@ support (see Hackage docs for setup)
```

Plus, does xdotool need emwh even?

XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -32,26 +33,33 @@ import XMonad.Hooks.DynamicLog (xmobarAction, xmobarRaw, PP(..))
...
 --   * use of UnsafeStdinReader in xmobarrc (rather than StdinReader)
```
slotThe, line 38: Does this work with `UnsafeXMonadLog` as well?  I feel like it should; I will try to test this

XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -32,26 +33,33 @@ import XMonad.Hooks.DynamicLog (xmobarAction, xmobarRaw, PP(..))
...
 
 
 clickableWrap :: Int -> String -> String
-clickableWrap i ws = xmobarAction ("wmctrl -s " ++ show i) "1" $ xmobarRaw ws
+clickableWrap i ws = xmobarAction ("xdotool set_desktop " ++ show i) "1" $ xmobarRaw ws
 
 -- Use index of workspace in users config to target workspace with wmctrl switch.
```
slotThe, line 44:
```suggestion
-- | Use index of workspace in users config to target workspace with @xdotool@ switch.
```

XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -32,26 +33,33 @@ import XMonad.Hooks.DynamicLog (xmobarAction, xmobarRaw, PP(..))
...
-getClickable :: X (WorkspaceId -> String)
-getClickable = do
+getClickable :: (WorkspaceId -> String) -> X (WorkspaceId -> String)
```
slotThe, line 45: If this is really a "renaming" function, then it should be `WorkspaceId -> WorkspaceId`, right?

xmonad-contrib.cabal
```diff
@@ -318,6 +318,7 @@ library
                         XMonad.Prompt.Zsh
                         XMonad.Util.ActionCycle
                         XMonad.Util.ClickableWorkspaces
+                        XMonad.Util.ClickableWorkspaces.Integrations
```
slotThe, line 321: I wonder what the rationale is for creating a whole different module for these?  Seems like they would fit in `XMonad.Util.ClickableWorkspaces` just fine

slotThe (commented):
Looks good to me; just some small nits I've noticed (I'm getting good at this!)

XMonad/Util/ClickableWorkspaces.hs
liskin: For set_desktop and windowactivate, it does. It uses the same clientmessages as wmctrl, just sets the source indication correctly, as explained in more detail in the commit message.

XMonad/Util/ClickableWorkspaces.hs
liskin: Yeah, wouldn't work for me otherwise. I'll update the comment.

XMonad/Util/ClickableWorkspaces.hs
liskin: I like it more this way, as it conveys the idea that the input should really be a valid workspace name (as opposed to something that's already been renamed) whereas the output can be anything at all.

xmonad-contrib.cabal
liskin: Probably premature and entirely unnecessary cyclic deps prevention. I couldn't decide whether to put these into ClickableWorkspaces or into IndependentScreen and WorkspaceNames, and just added a separate module instead. I can probably put these into ClickableWorkspaces if I'm the only one who thought it's somehow wrong. :-)

liskin:
@slotThe Thanks for the quick feedback and for catching those forgotten doc bits! 💖

XMonad/Util/ClickableWorkspaces.hs
slotThe: Aha okay, that makes sense

ivanbrennan:
@liskin Thanks for putting this together. I'm in the process of testing it with IndependentScreens. I'll report back shortly.

XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -18,44 +18,72 @@ module XMonad.Util.ClickableWorkspaces (
...
+-- | Integration with "XMonad.Actions.WorkspaceNames".
+clickableWorkspaceNamesPP :: PP -> X PP
+clickableWorkspaceNamesPP pp = do
+    rename <- getWorkspaceNames
+    clickableRenamedPP rename pp
+
+-- | Integration with "XMonad.Layout.IndependentScreens".
+clickableMarshallPP :: ScreenId -> PP -> X PP
+clickableMarshallPP s pp =
+    clickableRenamedPP (snd . unmarshall) pp{ ppSort = marshallSort s <$> ppSort pp }
```
ivanbrennan, line 89: This seems to work as far as hiding the ScreenId prefixes (e.g. `0_0` and `0_1` show as `0` and `1`). I can't seem to get it to play nicely with `clickableWorkspaceNamesPP` though -- workspaces aren't displaying their names, just their numbers.

I have an admittedly complicated use-case, integrating IndependentScreens, WorkspaceNames, and DynamicBars, so I'm not opposed to providing this helper as is, but I'm curious whether it's possible to support use-cases like mine (and whether it already can but I'm just misusing it).

Here's a snippet from my logHook, when using these integration helpers. The workspaces are clickable and ScreenId prefixes are hidden, but as I mentioned, workspace names aren't shown:
```haskell
logString :: PP -> ScreenId -> X String
logString pp =
  composePP pp >=> dynamicLogString

composePP :: PP -> ScreenId -> X PP
composePP pp s = do
  pp'  <- clickableWorkspaceNamesPP pp
  pp'' <- clickableMarshallPP s pp'
  pure (filterOutWsPP [scratchpadWorkspaceTag] pp'')
```
In case it's useful, a larger snippet offering more context is here: https://gist.github.com/ivanbrennan/f1e75efe0d8c8c969f256b53a93e3510

liskin: Oh, I totally forgot that you're using all three (clickable, independent, names)! This will need a functions that combines it all together, as these unfortunately don't compose nicely as they all need to see the original unmangled workspace tag. I'll push an update later today. Thanks for your feedback so far.

XMonad/Util/ClickableWorkspaces.hs
```diff
@@ -18,44 +18,72 @@ module XMonad.Util.ClickableWorkspaces (
...
+clickableWrap i ws = xmobarAction ("xdotool set_desktop " ++ show i) "1" $ xmobarRaw ws
 
--- Use index of workspace in users config to target workspace with wmctrl switch.
-getClickable :: X (WorkspaceId -> String)
-getClickable = do
+-- | Use index of workspace in users config to target workspace with @xdotool@ switch.
+getClickable :: (WorkspaceId -> String) -> X (WorkspaceId -> String)
+getClickable ren = do
   wsIndex <- getWsIndex
   return $ \ws -> case wsIndex ws of
```
ivanbrennan, line 56: In order to get IndepenentScreens and WorkspaceNames to play nicely together, I had to apply the rename at a different point,
```haskell
return $ \ws -> case wsIndex ws of
    case wsIndex (ren ws)
        Just idx -> clickableWrap idx ws
        Nothing -> ws
```
Maybe that's related to https://github.com/xmonad/xmonad-contrib/pull/462/files#r569591764

XMonad/Util/ClickableWorkspaces.hs
liskin: (It would all be much better if we had a PP where the ppCurrent, ppVisible, … functions were given the original workspace (e.g. `WorkspaceId → String → String`), then they'd compose better. But I'm not sure we can change the type now. :-/)

XMonad/Util/ClickableWorkspaces.hs
ivanbrennan: Ah, in my case the more appropriate type for `ren` was `String -> WorkspaceId`.

liskin: Yeah I remember we talked about this in https://github.com/xmonad/xmonad-contrib/pull/390 and I was confused then, but I think I understand it better now. You had to apply `ren` (`marshall`) before `wsIndex` because the `marshallPP` already mangled the workspace names for you. But `clickableMarshallPP` uses `unmarshall` only after the workspace name is looked up, so it should work well.

The only issue here, I believe, is that you use both IndependentWorkspaces and WorkspaceNames and there simply isn't any `clickableMarshallWorkspaceNamesPP`. :-)

I'll add one.

liskin: I pushed a `clickableMarshallWorkspaceNamesPP` which should make almost all of your https://github.com/ivanbrennan/nixbox/blob/c10c2d61b8d1bfffa0fb05afe721a6cf88e91844/services/xserver/xmonad.hs#L382-L393 and https://github.com/ivanbrennan/nixbox/blob/c10c2d61b8d1bfffa0fb05afe721a6cf88e91844/services/xserver/xmonad.hs#L418-L429 unnecessary. Just stick it in `logString`, compose with `namedScratchpadFilterOutWorkspacePP`, and I hope it'll work.

ivanbrennan: @liskin Thanks! I tried it out and it all seems to work perfectly now.

ivanbrennan (approved):

XMonad/Util/ClickableWorkspaces.hs
liskin: Awesome!

liskin:
Oh and I just realized I totally forgot that @geoffder is the maintainer of this module and I should ping them as well. :-)

geoffder:
> Oh and I just realized I totally forgot that @geoffder is the maintainer of this module and I should ping them as well. :-)

Hah! Thanks for the ping, and thanks for doing the work to integrate this better with other modules in the ecosystem!

XMonad/Util/ClickableWorkspaces.hs
liskin: I think I finally figured out a nice way to make the composition of `clickablePP`, `workspaceNamesPP` and `marshallPP` just work: https://github.com/xmonad/xmonad-contrib/pull/481

PR #463, opened 2021-02-07
Dynamic Status Bars support for XMonad.Hooks.StatusBar

TheMC47:
### Description

This provides dynamic status bars, that react to screen configuration changes. This needs the new interface from #443 and the splitting from #465, so I'm marking the PR as a draft till that's resolved.

Heavily inspired by "XMonad.Hooks.DynamicBars"
### Checklist

### DIFF
--- a/XMonad/Hooks/StatusBar.hs
+++ b/XMonad/Hooks/StatusBar.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE FlexibleContexts, TypeApplications #-}
+{-# LANGUAGE FlexibleContexts, TypeApplications, TupleSections  #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Hooks.StatusBar
@@ -39,6 +39,11 @@ module XMonad.Hooks.StatusBar (
   -- * Multiple Status Bars
   -- $multiple
 
+  -- * Dynamic Status Bars
+  -- $dynamic
+  dynamicSBs,
+  dynamicEasySBs,
+
   -- * Property Logging utilities
   xmonadPropLog,
   xmonadPropLog',
@@ -60,7 +65,7 @@ import System.Posix.Types (ProcessID)
 import Foreign.C (CChar)
 
 import XMonad
-import XMonad.Prelude ( traverse_, void )
+import XMonad.Prelude
 
 import XMonad.Util.Run
 import qualified XMonad.Util.ExtensibleState as XS
@@ -69,6 +74,8 @@ import XMonad.Layout.LayoutModifier
 import XMonad.Hooks.ManageDocks
 import XMonad.Hooks.StatusBar.PP
 import qualified XMonad.StackSet as W
+
+import Graphics.X11.Xrandr (xrrSelectInput)
 
 -- $usage
 -- You can use this module with the following in your @~\/.xmonad\/xmonad.hs@:
@@ -316,7 +323,7 @@ statusBarPipe cmd xpp  = do
 -- Here's an example of what such declarative configuration of multiple status
 -- bars may look like:
 --
--- > -- Make sure to setup the xmobar config accordingly
+-- > -- Make sure to setup the xmobar configs accordingly
 -- > xmobarTop    = statusBarPropTo "_XMONAD_LOG_1" "xmobar -x 0 ~/.config/xmobar/xmobarrc_top"    (pure ppTop)
 -- > xmobarBottom = statusBarPropTo "_XMONAD_LOG_2" "xmobar -x 0 ~/.config/xmobar/xmobarrc_bottom" (pure ppBottom)
 -- > xmobar1      = statusBarPropTo "_XMONAD_LOG_3" "xmobar -x 1 ~/.config/xmobar/xmobarrc1"       (pure pp1)
@@ -360,6 +367,105 @@ statusBarPipe cmd xpp  = do
 --
 -- By using the new interface, the config becomes more declarative and there's
 -- less room for errors.
+--
+-- The only *problem* now is that the status bars will not be updated when your screen
+-- configuration changes (by plugging in a monitor, for example). Check the section
+-- on dynamic status bars for how to do that.
+
+-- $dynamic
+-- Using multiple status bars by just combining them with '<>' works well
+-- as long as the screen configuration does not change often. If it does,
+-- you should use 'dynamicSBs': by providing a function that creates
+-- status bars, it takes care of setting up the event hook, the log hook
+-- and the startup hook necessary to make the status bars, well, dynamic.
+--
+-- > xmobarTop    = statusBarPropTo "_XMONAD_LOG_1" "xmobar -x 0 ~/.config/xmobar/xmobarrc_top"    (pure ppTop)
+-- > xmobarBottom = statusBarPropTo "_XMONAD_LOG_2" "xmobar -x 0 ~/.config/xmobar/xmobarrc_bottom" (pure ppBottom)
+-- > xmobar1      = statusBarPropTo "_XMONAD_LOG_3" "xmobar -x 1 ~/.config/xmobar/xmobarrc1"       (pure pp1)
+-- >
+-- > barSpawner :: ScreenId -> IO StatusBarConfig
+-- > barSpawner 0 = pure $ xmobarTop <> xmobarBottom -- two bars on the main screen
+-- > barSpawner 1 = pure $ xmobar1
+-- > barSpawner _ = mempty -- nothing on the rest of the screens
+-- >
+-- > main = xmonad $ dynamicSBs barSpawner (def { ... })
+--
+-- Make sure you specify which screen to place the status bar on (in xmobar,
+-- this is achieved by the @-x@ argument). In addition to making sure that your
+-- status bar lands where you intended it to land, the commands are used
+-- internally to keep track of the status bars.
+--
+-- Note also that this interface can be used with one screen, or if
+-- the screen configuration doesn't change.
+
+newtype ActiveSBs = ASB {getASBs :: [(ScreenId,  StatusBarConfig)]}
+
+instance ExtensionClass ActiveSBs where
+  initialValue = ASB []
+
+-- | Given a function to create status bars, 'dynamicSBs'
+-- adds the dynamic status bar capabilities to the config.
+-- For a version of this function that applies 'docks' and
+-- 'avoidStruts', check 'dynamicEasySBs'.
+--
+-- Heavily inspired by "XMonad.Hooks.DynamicBars"
+dynamicSBs :: (ScreenId -> IO StatusBarConfig) -> XConfig l -> XConfig l
+dynamicSBs f conf = conf
+  { startupHook     = startupHook conf
+                      >> setupEventHandler
+                      >> killAllStatusBars
+                      >> updateSBs f
+  , logHook         = logHook conf >> logSBs
+  , handleEventHook = eventHookSBs f <> handleEventHook conf
+  }
+
+-- | Like 'dynamicSBs', but applies 'docks' to the
+-- resulting config and adds 'avoidStruts' to the
+-- layout.
+dynamicEasySBs :: LayoutClass l Window
+               => (ScreenId -> IO StatusBarConfig)
+               -> XConfig l
+               -> XConfig (ModifiedLayout AvoidStruts l)
+dynamicEasySBs f conf =
+  docks . dynamicSBs f $ conf { layoutHook = avoidStruts (layoutHook conf) }
+
+-- | Given the function to create status bars, update
+-- the status bars by killing those that shouldn't be
+-- visible anymore and creates any missing status bars
+updateSBs :: (ScreenId -> IO StatusBarConfig) -> X ()
+updateSBs f = do
+  actualScreens    <- withWindowSet $ return . map W.screen . W.screens
+  (toKeep, toKill) <-
+    partition ((`elem` actualScreens) . fst) . getASBs <$> XS.get
+  -- Kill the status bars
+  cleanSBs (map snd toKill)
+  -- Create new status bars if needed
+  let missing = actualScreens \\ map fst toKeep
+  added <- io $ traverse (\s -> (s,) <$> f s) missing
+  traverse_ (sbStartupHook . snd) added
+  XS.put (ASB (toKeep ++ added))
+
+-- | Handles 'RRScreenChangeNotifyEvent' by updating the
+-- status bars.
+eventHookSBs :: (ScreenId -> IO StatusBarConfig) -> Event -> X All
+eventHookSBs f RRScreenChangeNotifyEvent{} = updateSBs f >> return (All True)
+eventHookSBs _ _                           = return (All True)
+
+-- | Run 'sbLogHook' for the saved 'StatusBarConfig's
+logSBs :: X ()
+logSBs = XS.get >>= traverse_ (sbLogHook . snd) . getASBs
+
+-- | Subscribe to the 'RRScreenChangeNotifyEvent'
+setupEventHandler :: X ()
+setupEventHandler = do
+  dpy  <- asks display
+  root <- asks theRoot
+  io $ xrrSelectInput dpy root rrScreenChangeNotifyMask
+
+-- | Kill the given 'StatusBarConfig's from the given
+-- list
+cleanSBs :: [StatusBarConfig] -> X ()
+cleanSBs = traverse_ sbCleanupHook
 
 -- | The default property xmonad writes to. (@_XMONAD_LOG@).
 xmonadDefProp :: String
@@ -428,7 +534,6 @@ spawnStatusBar :: String -- ^ The command used to spawn the status bar
 spawnStatusBar cmd = do
   newPid <- spawnPID cmd
   XS.modify (StatusBarPIDs . M.insert cmd newPid . getPIDs)
-
 
 -- | Kill all status bars started with 'spawnStatusBar'. Note the
 -- caveats in 'cleanupStatusBar'
--- a/XMonad/Hooks/StatusBar.hs
+++ b/XMonad/Hooks/StatusBar.hs
@@ -44,25 +44,23 @@ module XMonad.Hooks.StatusBar (
   xmonadPropLog',
   xmonadDefProp,
 
-  -- * Managing Status Bar Processes
-  spawnStatusBarAndRemember,
-  cleanupStatusBars,
-
-  -- * Manual Plumbing
-  -- $plumbing
+  -- * Managing status bar Processes
+  -- $sbprocess
+  spawnStatusBar,
+  killStatusBar,
+  killAllStatusBars,
   ) where
 
 import Control.Exception (SomeException, try)
 import qualified Codec.Binary.UTF8.String as UTF8 (encode)
+import qualified Data.Map as M
 import System.Posix.Signals (sigTERM, signalProcessGroup)
 import System.Posix.Types (ProcessID)
-
-import qualified Data.Map        as M
 
 import Foreign.C (CChar)
 
 import XMonad
-import XMonad.Prelude (void)
+import XMonad.Prelude ( traverse_, void )
 
 import XMonad.Util.Run
 import qualified XMonad.Util.ExtensibleState as XS
@@ -70,6 +68,7 @@ import qualified XMonad.Util.ExtensibleState as XS
 import XMonad.Layout.LayoutModifier
 import XMonad.Hooks.ManageDocks
 import XMonad.Hooks.StatusBar.PP
+import qualified XMonad.StackSet as W
 
 -- $usage
 -- You can use this module with the following in your @~\/.xmonad\/xmonad.hs@:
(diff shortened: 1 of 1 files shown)


XMonad/Hooks/DynamicStatusBarConfigs.hs
```diff
@@ -0,0 +1,125 @@
...
+
+-- | Given a function to create status bars, 'dynamicSBs'
+-- adds the dynamic status bar capabilities to the config.
+dynamicSBs :: (ScreenId -> IO StatusBarConfig) -> XConfig l -> XConfig l
+dynamicSBs f conf = conf
+  { startupHook     = startupHook conf
+                      *> setupEventHandler
+                      *> cleanAll
+                      *> updateSBs f
+  , logHook         = logHook conf *> logSBs
```
liskin, line 68: Just curious: why `*>` in here and `<>` in `handleEventHook`?

XMonad/Hooks/DynamicStatusBarConfigs.hs
```diff
@@ -0,0 +1,125 @@
+{-# LANGUAGE TupleSections #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Hooks.DynamicStatusBarConfigs
```
liskin, line 4: This looks like a nice and very useful module. It deserves a better name than this. :-)
(I'll keep it in mind and if I can think of something, I'll let you know.)

CHANGES.md
```diff
@@ -46,6 +46,10 @@
       in how the xmonad core handles XDG directories.
 
 ### New Modules
+  * `XMonad.Hooks.DynamicStatusBarConfigs`
```
liskin, line 49: Empty line above, pls.

Also, an entry in X.D.Extending.

liskin (commented):
First glance: nice. I love the simplicity of the user interface. I'll need to take a closer look at the internals later.

XMonad/Hooks/DynamicStatusBarConfigs.hs
```diff
@@ -0,0 +1,125 @@
...
+
+-- | Kill the given 'StatusBarConfig's from the given
+-- list
+cleanSBs :: [StatusBarConfig] -> X ()
+cleanSBs = traverse_ sbCleanupHook
+
+-- | Kill all status bars
+cleanAll :: X ()
+cleanAll =
+  getASBs <$> XS.get >>= traverse_ (sbCleanupHook . snd) >> XS.put (ASB [])
```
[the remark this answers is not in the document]
TheMC47: How did I miss that :sweat_smile:  It only occurred to me now that this is a NOOP

This also means, without changes, this won't work with property-based loggers :(

XMonad/Hooks/DynamicStatusBarConfigs.hs
TheMC47: I thought of the `logHook` and `startupHook` more like a sequence of actions, and `handleEventHook` as a single action that is composed.. I'll make them the same

liskin: Oh, that makes sense. Although I must admit that `*>` invokes more of a "give me the result on the right" vibe rather than "sequencing", for which `>>` is more common, at least among those who learned Haskell before `*>` was a thing. :-)

XMonad/Hooks/DynamicStatusBarConfigs.hs
TheMC47: I'm... bad at naming things

XMonad/Hooks/DynamicStatusBarConfigs.hs
TheMC47: I guess, I was influenced by the choices @slotThe  made in the original `makeStatusBar` and it kinda stuck with me.. I prefer `<>` though, but `>>` is more common

XMonad/Hooks/DynamicStatusBarConfigs.hs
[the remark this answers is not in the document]
TheMC47: After thinking about this, `cleanAll` should go into `DynamicLog` and kill all managed status bars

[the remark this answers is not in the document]
liskin: That might work.

XMonad/Hooks/DynamicStatusBarConfigs.hs
slotThe: > at least among those who learned Haskell before `*>` was a thing. :-)

Coming from someone who did not, I pretty much treat `>>` as a deprecated versin of `*>` (much like `liftM` -> `fmap`, `mapM` -> `traverse`, `return` -> `pure`, ... we really have a lot of those now) so both convey a sequencing vibe to me.  I do think that `<>` invokes more of a "this is a single thing" instead of "these are different, mutually independent steps" feeling

XMonad/Hooks/DynamicStatusBarConfigs.hs
```diff
@@ -0,0 +1,125 @@
...
+-- | Given a function to create status bars, 'dynamicSBs'
+-- adds the dynamic status bar capabilities to the config.
+dynamicSBs :: (ScreenId -> IO StatusBarConfig) -> XConfig l -> XConfig l
+dynamicSBs f conf = conf
+  { startupHook     = startupHook conf
+                      *> setupEventHandler
+                      *> cleanAll
+                      *> updateSBs f
+  , logHook         = logHook conf *> logSBs
+  , handleEventHook = eventHookSBs f <> handleEventHook conf
```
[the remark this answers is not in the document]
liskin: I could maybe synthesize the RRScreenChangeNotify event that was discarded by Rescreen and explicitly invoke handleEventHook on it. It's a horrible hack, though.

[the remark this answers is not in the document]
TheMC47: I only took that event because it was the one used in `DynamicBars`, but I don't really know if it's the best choice...

What's the problem here?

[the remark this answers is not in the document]
liskin: The problem is that my `X.H.Rescreen` module completely replaces xmonad's handling of Configure and RRScreenChangeNotify events in a way that doesn't compose well with other hooks (prevents other hooks from seeing all the events, prevent xmonad core from handling any of the events). It needs to do this because the X server sends several such events on every xrandr change, so

1. if one wants to attach a hook that restarts something (bars that don't reposition themselves, for example), the events need to be debounced for the hook to be invoked only once
2. if one wants to attach two different hooks, one for actual screen layout reconfiguration, another for just connect/disconnect of outputs, we need to look at all the events to detect which of the two situations it was

If this went into the core with two additional hooks in XConfig, everything would then compose nicely, but the xmonad philosophy (as I see it) has always been to keep the core minimal and try to do as much as possible in optional contrib modules. It usually works well, it just sometimes gets awkward if contirb modules interact/overlap as we never know if the user will use one, the other, or both.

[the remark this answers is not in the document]
TheMC47: I'm not sure I follow, don't all event handlers that are composed with `<>` see the same events? I think the `All False` will only prevent the xmonad default handler from being dispatched, and it doesn't affect the other events right?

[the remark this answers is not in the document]
geekosaur: Correct, you would need to add your own logic to suppress your own event handlers in that case. Note that xmonad has no idea what hooks return `All False` until the hook has executed, since they can contain arbitrary logic.

liskin:
Actually not really correct, as my hooks looks into the event queue directly and discards events. I'll explain more when I get back from a walk. Dne 13. 2. 2021 12:36 napsal uživatel brandon s allbery kf8nh <notifications@github.com>:
@geekosaur commented on this pull request.



In XMonad/Hooks/DynamicStatusBarConfigs.hs:
> +
+newtype ActiveSBs = ASB {getASBs :: [(ScreenId,  StatusBarConfig)]}
+
+instance ExtensionClass ActiveSBs where
+  initialValue = ASB []
+
+-- | Given a function to create status bars, 'dynamicSBs'
+-- adds the dynamic status bar capabilities to the config.
+dynamicSBs :: (ScreenId -> IO StatusBarConfig) -> XConfig l -> XConfig l
+dynamicSBs f conf = conf
+  { startupHook     = startupHook conf
+                      *> setupEventHandler
+                      *> cleanAll
+                      *> updateSBs f
+  , logHook         = logHook conf *> logSBs
+  , handleEventHook = eventHookSBs f <> handleEventHook conf

Correct, you would need to add your own logic to suppress your own event handlers in that case. Note that xmonad has no idea what hooks return All False until the hook has executed, since they can contain arbitrary logic.

XMonad/Hooks/DynamicStatusBarConfigs.hs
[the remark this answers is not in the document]
liskin: All events handlers do see the same events, but `X.H.Rescreen` looks into the X event queue directly and discard duplicate events, so other hooks only see the first Configure or RRScreenChangeNotify event, whichever comes first, and all the other events are discarded. I should probably explain this more clearly in the documentation there.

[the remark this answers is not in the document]
TheMC47: Alright that makes sense now

Well, I guess we should export the event handler (and the other hooks) from this module and explain how to make it work with `X.H.Rescreen` (Adding it to the `RescreenConfig`, and manually add the startup hook). I'm not sure that calling the `handleEvent` on it manually is a good idea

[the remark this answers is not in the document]
TheMC47: @liskin  so what should we do here?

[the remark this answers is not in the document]
liskin: Nothing for now. :-)

I got an idea that might improve composability of stuff like this (something like extensible state, just for XConfig, an "extensible config"). But I'd like to finish other stuff first, so let's not worry about the interaction here and deal with it later. I don't really need X.H.Rescreen in my config, auto-xrandr is more trouble than it's worth, so I'll just try using your code and see how it behaves. (But I won't get to it sooner than end of this week.)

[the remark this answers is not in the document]
TheMC47: This is not at all urgent, so take your time!

TheMC47:
Alright, I've been using this for a couple of months now, and it's been working great. Except for a minor issue: sometimes, some bars don't get started. Not sure why, it doesn't seem like a problem with the code itself.
Here are the logs from restarting xmonad. For the record, I restarted it like 5 times in a row to get it misbehaving. 
```
XMonad will use build script at "/home/yecinem/.xmonad/build" to recompile.
XMonad recompilation process exited with success!
XMonad is recompiling and replacing itself with another XMonad process because the current process is called "xmonad" but the compiled configuration should be called "xmonad-x86_64-linux"
XMonad will use build script at "/home/yecinem/.xmonad/build" to recompile.
XMonad recompilation process exited with success!
xxmmoobbaarr::  CCaauugghhtt  ssiiggnnaall  1155;;  eexxiittiinngg......

xmonad-x86_64-linux: failed to create OS thread: Resource temporarily unavailable
Bad _NET_DESKTOP with data[0]=-1
Bad _NET_DESKTOP with data[0]=-1
Bad _NET_DESKTOP with data[0]=-1
`````

**Comment by TheMC47** [comment]
> Alright, I've been using this for a couple of months now, and it's been working great. Except for a minor issue: sometimes, some bars don't get started. Not sure why, it doesn't seem like a problem with the code itself.
> Here are the logs from restarting xmonad. For the record, I restarted it like 5 times in a row to get it misbehaving.
> 
> ```
> XMonad will use build script at "/home/yecinem/.xmonad/build" to recompile.
> XMonad recompilation process exited with success!
> XMonad is recompiling and replacing itself with another XMonad process because the current process is called "xmonad" but the compiled configuration should be called "xmonad-x86_64-linux"
> XMonad will use build script at "/home/yecinem/.xmonad/build" to recompile.
> XMonad recompilation process exited with success!
> xxmmoobbaarr::  CCaauugghhtt  ssiiggnnaall  1155;;  eexxiittiinngg......
> 
> xmonad-x86_64-linux: failed to create OS thread: Resource temporarily unavailable
> Bad _NET_DESKTOP with data[0]=-1
> Bad _NET_DESKTOP with data[0]=-1
> Bad _NET_DESKTOP with data[0]=-1
> ```

As we discussed on the IRC, this was because of `-rtsopts -with-rtsopts=-N`

**Comment by oogeek** [comment]
Overall it looks good. But I have a issue here.
Use-case example:
```
Two monitors:
laptop 1920x1080 
2K display 2560x1440 (external)
```
And set external monitor to be primary.
Then in the definition of `barspawner`:
```
barSpawner :: ScreenId -> IO StatusBarConfig
barSpawner 0 = pure xmobar0  -- two bars on the main screen
barSpawner 1 = pure xmobar1
barSpawner _ = Data.Monoid.mempty -- nothing on the rest of the screens

```
`xmobar0` is used for external display.
But after unplugging the external display, the primary display becomes the laptop. But `xmobar0` is for external display.

**Comment by TheMC47** [comment]
Thanks for your feedback! For your use case, I think this is out of the scope of this PR (and I guess of xmonad? not sure). We can't control which id is given to which screen. But maybe I'm wrong, I know very little about X.

**Comment by oogeek** [comment]
> Thanks for your feedback! For your use case, I think this is out of the scope of this PR (and I guess of xmonad? not sure). We can't control which id is given to which screen. But maybe I'm wrong, I know very little about X.

So I think for my use-case, I need to do a bit of hack. Let me see. :)

File: XMonad/Hooks/StatusBar.hs [hunk 6]
```diff
@@ -358,6 +364,88 @@ statusBarPipe cmd xpp  = do
 --
 -- By using the new interface, the config becomes more declarative and there's
 -- less room for errors.
+--
+-- The only *problem* now is that the status bars will not be updated when your screen
+-- configuration changes (by plugging in a monitor, for example). Check the section
+-- on dynamic status bars for how to do that.
+
+-- $dynamic
+-- Using multiple status bars just by combining them with '<>' works well
```
slotThe, line 373:
```suggestion
-- Using multiple status bars by just combining them with '<>' works well
```

XMonad/Hooks/StatusBar.hs
```diff
@@ -358,6 +364,88 @@ statusBarPipe cmd xpp  = do
...
+-- and the startup hook necessary to make the status bars, well, dynamic.
+--
+-- > xmobarTop    = statusBarPropTo "_XMONAD_LOG_1" "xmobar -x 0 ~/.config/xmobar/xmobarrc_top"    (pure ppTop)
+-- > xmobarBottom = statusBarPropTo "_XMONAD_LOG_2" "xmobar -x 0 ~/.config/xmobar/xmobarrc_bottom" (pure ppBottom)
+-- > xmobar1      = statusBarPropTo "_XMONAD_LOG_3" "xmobar -x 1 ~/.config/xmobar/xmobarrc1"       (pure pp1)
+-- >
+-- > barSpawner :: ScreenId -> IO StatusBarConfig
+-- > barSpawner 0 = pure $ xmobarTop <> xmobarBottom -- two bars on the main screen
+-- > barSpawner 1 = pure $ xmobar1
+-- > barSpawner _ = memtpy -- nothing on the rest of the screens
```
slotThe, line 386:
```suggestion
-- > barSpawner _ = mempty -- nothing on the rest of the screens
```

XMonad/Hooks/StatusBar.hs
```diff
@@ -358,6 +364,88 @@ statusBarPipe cmd xpp  = do
...
+
+instance ExtensionClass ActiveSBs where
+  initialValue = ASB []
+
+-- | Given a function to create status bars, 'dynamicSBs'
+-- adds the dynamic status bar capabilities to the config.
+--
+-- Heavily inspired by "XMonad.Hooks.DynamicBars"
+dynamicSBs :: (ScreenId -> IO StatusBarConfig) -> XConfig l -> XConfig l
+dynamicSBs f conf = conf
```
slotThe, line 403: I got caught off-guard a bit by this not adding `docks` (I only read the `dynamicSB` docks, maybe it's mentioned somewhere else).

Is there a use-case where one would want to add status bars but _not_ have them recognized as such?

XMonad/Hooks/StatusBar.hs
```diff
@@ -386,41 +474,48 @@ xmonadPropLog' prop msg = do
...
-                          -> X ()
-spawnStatusBarAndRemember cmd = do
+spawnStatusBar :: String -- ^ The command used to spawn the status bar
+               -> X ()
+spawnStatusBar cmd = do
   newPid <- spawnPID cmd
-  XS.modify (StatusBarPIDs . (newPid :) . getPIDs)
+  XS.modify (StatusBarPIDs . M.insert cmd newPid . getPIDs)
+
+
```
slotThe, line 516:
```suggestion
```

slotThe (commented):
This looks cool :+1:  

I'll play around with the actual functionality a bit more in the coming days

XMonad/Hooks/StatusBar.hs
TheMC47: Probably not. Following the same argument, should we also add `docks` (the layout modifier too? not sure) to `withSB`.

Whatever option we choose, I think we should stay consistent - this should behave like `withSB`.

slotThe: Yeah I think we should add it to `withSB` as well; bars tend to behave very weirdly if they're not recognized as bars. 

I'm kind of torn about `avoidStruts`.  The vast majority of people will probably just use it for all layouts, but there is a legitimate use-case for having certain layouts just not respect the bar.  It's probably better to leave that one to the user.

TheMC47: Out of curiosity, why would one not apply `avoidStruts`?

geekosaur: https://github.com/geekosaur/xmonad.hs/blob/pyanfar/xmonad.hs#L142
The quick and dirty version being to not apply it at all, but then you can't toggle it either.

TheMC47: Interesting. So I guess we don't apply the layout modifier then?

slotThe: > Out of curiosity, why would one not apply `avoidStruts`?

In addition to what geekosaur already said, I can imagine someone wanting something like a `Full` layout to be completely free of struts.

I'm not sure if these use-cases are enough to justify us not applying `avoidStruts`; one may well achieve them using other means

liskin: > Yeah I think we should add it to `withSB` as well

That's why we have `withEasySB`. There's no place for `docks` or `avoidStruts` in `withSB` or `dynamicSBs`.

Maybe later, once we have extensible/introspectable XConfig and can make combinators like `docks` and `ewmh` idempotent, then we can use them all over the place, but right now it's not safe to do that.

slotThe: Seems like people's configs will break either way; be that because an extraneous `docks` is added or because nothing is added at all.  Questions seems to be for what kind of "error case" we want to optimize.

liskin: The "missing `docks`" case is more obvious and easier to debug, and seems to be in line with the overall xmonad philosophy of giving users a choice.

Anyway, the obvious solution here is to just make it consistent and provide both `dynamicSBs` and `dynamicEasySBs`.

slotThe: Works for me :+1:

TheMC47: Sounds good 👍 I'll update it

TheMC47: Alright, that should be taken care of

slotThe (approved):

TheMC47:
As we spoke in the IRC, I mentioned that the commands themselves are used to keep track of what's launched. If someone could give it a proof read, we can merge this tonight :tada:

XMonad/Hooks/StatusBar.hs
```diff
@@ -390,6 +390,11 @@ statusBarPipe cmd xpp  = do
 -- >
 -- > main = xmonad $ dynamicSBs barSpawner (def { ... })
 --
+-- Beware of one caveat: if you want to launch multiple status bars but use the
```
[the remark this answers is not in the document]
TheMC47: Fair point, is it better now?

XMonad/Hooks/StatusBar.hs
```diff
@@ -390,10 +390,10 @@ statusBarPipe cmd xpp  = do
 -- >
 -- > main = xmonad $ dynamicSBs barSpawner (def { ... })
 --
--- Beware of one caveat: if you want to launch multiple status bars but use the
--- same command, you need to differentiate between them (for example, by adding
--- extra white spaces at the end). The string used to launch the commands is
--- also used to keep track of the status bars.
+-- Make sure you specify which screen to place the the status bar on (in xmobar,
```
liskin, line 393: Undouble the "the", otherwise this is really good. Note that there's a similar comment somewhere down that should be adjusted too.

XMonad/Hooks/StatusBar.hs
```diff
@@ -389,41 +494,52 @@ xmonadPropLog' prop msg = do
...
+    XS.modify (StatusBarPIDs . M.delete cmd . getPIDs)
+
+killPid :: ProcessID -> IO ()
+killPid pidToKill = void $ try @SomeException (signalProcessGroup sigTERM pidToKill)
+
+-- | Spawns a status bar and saves its PID together with the commands that was
+-- used to start it. This is useful when the status bars should be restarted
+-- with xmonad. Use this in combination with 'killStatusBar'.
+--
+-- If you want to start multiple bars with the same command and selectively kill
```
liskin, line 530: ^ this one

XMonad/Hooks/StatusBar.hs
TheMC47: I think I'm gonna just remove the comment there

liskin (approved):
No further suggestions :-)

PR #464, opened 2021-02-08
Added javaHack

TheMC47:
### Description
Fixes Java applications that don't work well with xmonad, by setting `_JAVA_AWT_WM_NONREPARENTING=1`

I saw something similar in @geekosaur's config, and I think it's a pretty common solution. Having this in the `X.U.Hacks` can be convenient for some users.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,6 +52,8 @@
     Serves as a collection of hacks and fixes that should be easily acessible to users.
     The first element of this module is windowedFullscreenFix, which fixes fullscreen behaviour
     of chromium based applications when using windowed fullscreen.
+    A second entry is `javaHack`, which helps when dealing with Java applications that might
+    not work well with xmonad.
 
   * `XMonad.Util.ActionCycle`
 
--- a/XMonad/Util/Hacks.hs
+++ b/XMonad/Util/Hacks.hs
@@ -23,12 +23,16 @@ module XMonad.Util.Hacks
   ( -- * Windowed fullscreen
     -- $windowedFullscreenFix
     windowedFullscreenFixEventHook
+    -- * Java Hack
+    -- $java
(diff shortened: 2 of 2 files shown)


geekosaur:
For what it's worth, it looks like this is also https://github.com/xmonad/xmonad/issues/126.

TheMC47:
I wonder if we should add this to the core

geekosaur:
It's certainly arguable. I could wish we had a similar hack for Mono, but they don't seem to be playing along.

TheMC47:
It should be a net improvement, as it won't bother anyone not using java GUIs, and at the same time, spare some users a lot of frustration

geekosaur:
Actually, in a sense it will be less debuggable: consider use with session managers (as with xmonad with KDE or MATE) or with launchers run separately from xmonad in a `.xsession` file. They won't have it set. We may already have seen this one in the past, come to think of it.

(It shouldn't have been an environment variable but a root window property or something like the EWMH properties. But that ship has sailed.)

liskin:
Lots of stuff, like EWMH, EWMH fullscreen, ManageDocks, etc. is a net improvement for most people, yet we still keep the core minimal yet tremendously extensible, and it's worked well for over a decade. If anything, xmonad-contrib could ship some sort of "extended core default" config that enables things most people need while still allowing some configuration of how those things work (https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Util/ExtensibleConf.hs lets us add arbitrary hooks and knobs, and even without it, we could've always had an `XContribConfig` record or something).

(But yeah, if we were doing xmonad all over again today, we'd probably have just one repo. Still, the minimal extensible core idea is genius and it seems this repo separation helps us poor humans not forget that.)

PR #465, opened 2021-02-12
Split XMonad.Hooks.DynamicLog

TheMC47:
### Description

- XMonad.Hooks.DynamicLog.PP: the pretty-printing abstraction, with the
corresponding utilities
- XMonad.Hooks.StatusBar: A new module to provide a better interface to
manage external status bars.

Still a WIP (mainly: I won't touch `CHANGES.md` till everything else is done); I created the PR so we can discuss more which functions go where and what should be added/deleted. Also, not sure if we let #463 open or close it and move the implementation here.

Reference: https://github.com/xmonad/xmonad-contrib/pull/443#issuecomment-777041434

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,7 +50,8 @@
     - The type of `getWorkspaceNames` was changed to fit into the new `ppRename`
       field of `PP`.
 
-  * `XMonad.Hooks.DynamicLog` and `XMonad.Util.Run`
+  * `XMonad.Hooks.StatusBar`, `XMonad.Hooks.StatusBar.PP` (previously
+    `XMonad.Hooks.DynamicLog`) and `XMonad.Util.Run`
 
     - `spawnPipe` no longer uses binary mode handles but defaults to the
       current locale encoding instead.
@@ -77,9 +78,31 @@
     Originally contained inside `XMonad.Hooks.DynamicLog`, this module provides the
     pretty-printing abstraction and utilities, used primarly with `logHook`.
 
+    Below are changes from `XMonad.Hooks.DynamicLog`, that now are included in
+    this module:
+
+    - Added `shortenLeft` function, like existing `shorten` but shortens by
+      truncating from left instead of right. Useful for showing directories.
+
+    - Added `shorten'` and `shortenLeft'` functions with customizable overflow
+      markers.
+
+    - Added `filterOutWsPP` for filtering out certain workspaces from being
+      displayed.
+
+    - Added `ppTitleUnfocused` to `PP` for showing unfocused windows on
+      the current workspace in the status bar.
+
+    - Added `xmobarBorder` function to create borders around strings.
+
+    - Added `ppRename` to `PP`, which makes it possible for extensions like
+      `workspaceNamesPP`, `marshallPP` and/or `clickablePP` to compose
+      intuitively.
+
+
   * `XMonad.Hooks.StatusBar`
 
-    This module provides a new interface that replaces "XMonad.Hooks.DynamicLog",
+    This module provides a new interface that replaces `XMonad.Hooks.DynamicLog`,
     by providing composoble status bars. Supports property-based as well
     as pipe-based status bars.
 
@@ -161,8 +184,8 @@
   * `XMonad.Util.ClickableWorkspaces`
 
     Provides `clickablePP`, which when applied to the `PP` pretty-printer used by
-    `XMonad.Hooks.DynamicLog.dynamicLogWithPP`, will make the workspace tags
-    clickable in XMobar (for switching focus).
+    `XMonad.Hooks.StatusBar.PP`, will make the workspace tags clickable in XMobar
+    (for switching focus).
 
   * `XMonad.Layout.VoidBorders`
 
@@ -240,27 +263,11 @@
 
   * `XMonad.Hooks.DynamicLog`
 
-    - Added `shortenLeft` function, like existing `shorten` but shortens by
-      truncating from left instead of right. Useful for showing directories.
-
-    - Added `shorten'` and `shortenLeft'` functions with customizable overflow
-      markers.
-
-    - Added `filterOutWsPP` for filtering out certain workspaces from being
-      displayed.
-
     - Added `xmobarProp`, for property-based alternative to `xmobar`.
-
-    - Reworked the module documentation to suggest property-based logging
-      instead of pipe-based logging, due to the various issues associated with
-      the latter.
-
-    - Added `ppTitleUnfocused` to `PP` for showing unfocused windows on
-      the current workspace in the status bar.
-
-    - Added `xmobarBorder` function to create borders around strings.
-
+    
     - Add the -dock argument to the dzen spawn arguments
+
+    - The API for this module is frozen: this is now a compatibility wrapper.
 
   * `XMonad.Layout.BoringWindows`
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -72,6 +72,17 @@
 
 ### New Modules
 
+  * `XMonad.Hooks.StatusBar.PP`
+
+    Originally contained inside `XMonad.Hooks.DynamicLog`, this module provides the
+    pretty-printing abstraction and utilities, used primarly with `logHook`.
+
+  * `XMonad.Hooks.StatusBar`
+
+    This module provides a new interface that replaces "XMonad.Hooks.DynamicLog",
+    by providing composoble status bars. Supports property-based as well
+    as pipe-based status bars.
+
   * `XMonad.Util.Hacks`
 
     Serves as a collection of hacks and fixes that should be easily acessible to users.
@@ -229,10 +240,6 @@
 
   * `XMonad.Hooks.DynamicLog`
 
-    - Added `statusBar'` function, like existing `statusBar` but accepts a pretty
-      printing options argument embedded in the X monad, to allow for dynamically
-      modified options such as `workspaceNamesPP`.
-
     - Added `shortenLeft` function, like existing `shorten` but shortens by
       truncating from left instead of right. Useful for showing directories.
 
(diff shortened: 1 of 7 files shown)


liskin (commented):
Looks good.

There's still a lot of documentation in the DynamicLog module that, assuming we want DynamicLog to become a compat module, should probably be moved elsewhere and just referred to from DynamicLog.

And we should think a bit more about the naming. If DynamicLog is deprecated, then DynamicLog.PP could probably become StatusBar.PP? Would it perhaps make sense to move all this to something like `XMonad.Log.PP` and `XMonad.Log.StatusBar`? Anyone has any idea, I want to hear it!

TheMC47:
I was a bit lazy when it came to the documentation, I'll take care of that when we're sure about where each function goes. `PP` should be moved to somewhere "safe", you're right about that

Otherwise, if we go with `XMonad.Log.*`, I think there are some other modules that also are intended to be used with status bars (like `ClickableWorkspaces`), and I'm not sure if it would okay to also move them to `XMonad.Log`. It would definitely help the discoverability of the modules, particularly if we make sure that all the future features get added there.

And I like `XMonad.Hooks.StatusBar.*` a bit more, since it's quite obvious that it is about status bars.

liskin:
> And I like `XMonad.Hooks.StatusBar.*` a bit more, since it's quite obvious that it is about status bars.

Fair point. After all, one doesn't type imports very often. :-)

XMonad/Hooks/StatusBar.hs
```diff
@@ -0,0 +1,422 @@
...
+-- >   xmonad =<< makeStatusBar mySB myConf
+--
+-- which plays nice with other combinators that you might have already
+-- in your config:
+--
+-- > main = do
+-- >   mySB <- statusBarProp "xmobar" (pure myPP)
+-- >   xmonad =<< (makeStatusBar mySB . ewmh . docks $ def {...})
+--
+-- It is generally a good idea to also add
```
liskin, line 92: Something seems to be missing here. :-)

(Please, from this point on, push fixup commits instead of forcepushing. I'm hacking on it too. :-))

TheMC47: I have absolutely no idea what I wanted to say

liskin:
I'm terribly sorry for procrastinating this long. That's how this stupid brain of mine works: once it decides to postpone something and starts feeling bad about it, it becomes hard to get back to it. :-(

Anyway, I decided that it'll be more effective to checkout your branch locally and just fix things, rather than commenting on the diff, and that's what I did, and I pushed what I managed to fix today. It's not perfect yet, but I'd say it's almost good to go.

What's left:

* CHANGES mentions additions to DynamicLog which are either not there at all, or are just compat reexports, so this needs to be improved to be more clear about DynamicLog/StatusBar/…
* some modules mention DynamicLog in their documentation, and this would be nice to somehow adapt to the new interface (I did the ClickableWorkspaces module as an example, we should convert at least WorkspaceNamed and IndependentScreens)
* would be lovely if someone proofread the docs of these 3 modules top to bottom and tried to make it more coherent — it's noticeable that it was written by various people over more than a decade, so it's not as nice a read as @slotThe's tutorial; but let's not obsess over it, I guess

Thanks again for working on this! It's great to see improvements in these parts of the codebase that almost all users use yet few dared to touch in years.

TheMC47:
Awesome! Don't worry about it, it's totally fine :smile: We definitely need to update the docs, but we could open another issue and collect all modules that need update, and see if some new users take care of them while we prepare for the release (that way we also get feedback on the new interface). 

I'll take a closer look at CHANGES; it's also been a while since I read the docs so I'll give it a go again. Thanks for the review and for your fixes!

XMonad/Hooks/StatusBar.hs
```diff
@@ -74,22 +77,25 @@ import XMonad.Hooks.StatusBar.PP
...
--- below), is to use 'statusBarProp' with 'makeStatusBar', which takes care of
--- the necessary plumbing-no shell scripting required!
+-- below), is to use 'statusBarProp' with 'makeStatusBar'', which takes care of
+-- the necessary plumbing:
 --
 -- > main = do
--- >   mySB <- statusBarProp "xmobar" (pure myPP)
--- >   xmonad =<< makeStatusBar mySB myConf
+-- >   mySB <- statusBarProp "xmobar" (pure xmobarPP)
+-- >   xmonad =<< makeStatusBar' mySB def
```
TheMC47, line 92: I think this is `makeStatusBar` , not `makeStatusBar'` (the latter takes an additional keybinding)

liskin: Oh. The change to `makeStatusBar'` is intentional, I wanted the basic example to not rely on any custom `myPP` and `myConf`, but I forgot about the additional parameter, and now I see that `toggleStrutsKey` is still in `X.H.DynamicLog` so I can't easily use this. :-/

liskin: 9a4bfee205059f4f3d21d5d710f441fa615b3887, perhaps?

slotThe:
> We definitely need to update the docs, but we could open another issue and collect all modules that need update, and see if some new users take care of them while we prepare for the release (that way we also get feedback on the new interface).

I was about ready for another night of endless fun, but I think that sounds even better!  It would perhaps be good to throw a lot of modules in there (like we're currently doing with the layouts) but also to mark certain modules as "essential" (i.e. these _really_ need to be updated before the new release)

XMonad/Hooks/StatusBar.hs
TheMC47: Looks good :+1:

TheMC47:
I mean, if people decide to contribute to that then it's good, but we can always do them ourselves before the release

slotThe:
On Fri, Mar 26 2021 11:40, Yecine Megdiche wrote:
> I mean, if people decide to contribute to that then it's good, but we
> can always do them ourselves before the release 

But for that we need to know which ones we, at one point, deemed
important ;)

TheMC47:
Alright, I tried to take care of the `CHANGES.md` and reworked the docs a tiny bit.

slotThe:
I added a few docs nits.  I will at this point also lobby for the renaming of `makeStatusBar'` to something more human-readable

TheMC47:
maybe`makeStatusBarWithKey`? or something entirely different?

slotThe:
On Mon, Mar 29 2021 04:43, Yecine Megdiche wrote:
> maybe `makeStatusBarWithKey`? or something entirely different? 

Sounds good to me

liskin:
> maybe`makeStatusBarWithKey`? or something entirely different?

I find this a bit confusing, as it does more than just set the key, but by being explicit about the key, people are discouraged from looking up what it actually does and will be more likely to understand it wrong. Something like `makeAndManageStatusBar` might be closer to what it does. Unfortunately we don't apply the layout modifier so any name will be confusing in one way or another. :-/

liskin:
Oh, we do apply the layout modified, never mind the last sentence then. (I think `makeAndManageStatusBar` and still far from perfect, though.)

CHANGES.md
```diff
@@ -77,9 +78,26 @@
...
+      markers.
+
+    - Added `filterOutWsPP` for filtering out certain workspaces from being
+      displayed.
+
+    - Added `ppTitleUnfocused` to `PP` for showing unfocused windows on
+      the current workspace in the status bar.
+
+    - Added `xmobarBorder` function to create borders around strings.
+
```
liskin, line 97: It seems we lost the `ppRename` entry somewhere.

XMonad/Hooks/StatusBar.hs
```diff
@@ -124,12 +126,12 @@ import XMonad.Hooks.StatusBar.PP
 --
 -- If your status bar does not support property-based logging, you may also try
 -- 'statusBarPipe'.
--- It can be used in the same way as 'statusBarProp' above (for xmobar, you will now
--- use the @StdinReader@ plugin in your @.xmobarrc@).  Instead of writing to
--- a property, this function opens a pipe and makes the given status bar read
--- from that pipe.
+-- It can be used in the same way as 'statusBarProp' above (for xmobar, you now
+-- have to use the @StdinReader@ plugin in your @.xmobarrc@).  Instead of
```
liskin, line 130: Since StdinReader is default in xmobar, I'm not sure the strong "have to" wording here is appropriate.

liskin:
I spotted two minor issues but overall I think we're aiming for a merge today or something. :-)

CHANGES.md
```diff
@@ -95,6 +95,11 @@
 
     - Added `xmobarBorder` function to create borders around strings.
 
+    - Added `ppRename` to `PP`, which makes it possible for extensions like
+      `workspaceNamesPP`, `marshallPP` and/or `clickablePP` to compose
+      intuitively.
```
liskin, line 100: oh it's here, sorry!

XMonad/Hooks/StatusBar.hs
geekosaur: Seems appropriate to me, since using this without either `StdinReader` or `UnsafeStdinReader` will lead to a hang, which is why all this rewiring.

slotThe: It is the default, but that's void as soon as people start customizing their `template` string (which I suspect most will have done).  

Really, this part is written for people who may have tried the property-logging solution above but want to switch to pipes again for one reason or another.  Requiring people to read every little bit is perhaps not the best way to go about things, but it's quite hard to find a wording that satisfies both sides here I feel.

liskin: Fair enough :-)

slotThe:
To summarize what was discussed on IRC because I feel like names are the
only things holding us back at this point:

  - I think that @liskin's suggested `makeAndManageStatusBar` is still
    pretty good, even though it's a bit long.
    
  - @TheMc48 suggested `bindSB` instead of `makeStatusBar` and
    `bindAndManageSB` instead of `makeStatusBar'`.  Here I'm personally
    not sure what "bind" is supposed to signal (at least it doesn't
    signal "spawn this program" to me)

TheMC47:
Split the difference and go with `makeSB` and `makeAndManageSB`?

liskin:
I have some more ideas:

* we could follow the pattern of `ewmh` and `docks`, and use just `statusBars` and maybe something like `managedStatusBars` or `statusBarsEasy` (but we might need to invent better names for `statusBar{Pipe,Prop}` then
* we could also get inspiration from UrgencyHook and have `withStatusBars` or `withSB` and `withManagedStatusBar` or `withEasySB` or something

I'm also totally okay with the last suggestion by @TheMC47 but just wanted to add my today's thoughts on this.

And yeah, we should stop with the bikeshedding, choose something and merge it tomorrow (I'll be out in the sun most of the day again though). :-)

geekosaur:
Must be nice… we've had snow showers and the occasional squall all day.

TheMC47:
I like `withSB` and `withEasySB` :+1:

slotThe:
Works for me; let's do it! :)

TheMC47:
Alright, I've renamed them to `withSB` and `withEasySB`. Let's merge this tonight if there are not other comments :tada:

Issue #377, opened 2020-09-06
XMonad.Hooks.DynamicLog corrupt non-ASC characters


PRESFIL:
### Problem Description

I switched from `xmonad-contrib==0.16-31`+`xmonad==0.15-51` to [`xmonad-contrib-git`](https://aur.archlinux.org/packages/xmonad-contrib-git/)+[`xmonad-git`](https://aur.archlinux.org/packages/xmonad-git/). Now, in `xmobar==0.35.31` characters are corrupted:
![изображение](https://user-images.githubusercontent.com/20644917/92325341-3d45ad00-f039-11ea-99c0-c5e73e7dcf73.png)
![изображение](https://user-images.githubusercontent.com/20644917/92325353-53ec0400-f039-11ea-8a38-9064f6513a6e.png)

**NOTE:** last commit on git version - [58feba91d96fee8339d7cf56fb33537be819eb4b](https://github.com/xmonad/xmonad-contrib/commit/58feba91d96fee8339d7cf56fb33537be819eb4b)

Then, i switching back to `xmonad-contrib==0.16-31`+`xmonad==0.15-51` and:
![изображение](https://user-images.githubusercontent.com/20644917/92325393-d7a5f080-f039-11ea-8e50-2a40f06fd54f.png)
![изображение](https://user-images.githubusercontent.com/20644917/92325397-e9879380-f039-11ea-909e-fc2b605665bb.png)

Is there are some default behavior of `xmonad`/`X.H.DynamicLog` changed or maintainer has forgotten some compile parameters for support `utf-8`?

### Configuration File

My test configuration:
```haskell
import XMonad
import XMonad.Hooks.DynamicLog
import XMonad.Hooks.EwmhDesktops
import XMonad.Layout.NoBorders
import XMonad.Util.SpawnOnce

import qualified Data.Map as M

myStartupHook = do
  spawnOnce "systemctl --user start dunst"
  spawnOnce "trayer --edge top --align right --SetDockType true --SetPartialStrut true --expand true --height 20 --transparent true --alpha 0 --tint 0x000000 --widthtype request --monitor 0"
  spawnOnce "xss-lock --transfer-sleep-lock -- sh -c \"xkb-switch -s us ; exec i3lock --nofork -c \\#000000\""

myKeys conf@(XConfig {XMonad.modMask = modm}) = M.fromList
  [ ((modm .|. shiftMask, xK_z), spawn "loginctl lock-session")
  ]

[...]

PR #471, opened 2021-02-16
Fix encoding issues in spawnPipe/dynamicLogString/xmonadPropLog

liskin:
#### [X.U.Run: Clean up spawnPipes a bit](../commit/9e9f4937eae716a7e6f833a917d2112986ca3463)

This makes it easier to see the differences between these functions,
makes it less likely someone will change one and not the others, etc.
More importantly, the documentation doesn't contain circular references
any more. :-)

Also, let's just use hSetEncoding. The concern of this being stateful
and theoretically having something written in the wrong encoding is
pointless: nobody has the handle until we return it from `spawnPipe'`.

(This also means that spawnPipeWithNoEncoding is now a text handle that
possibly does newline translation, just with char8 encoding. There
should be no difference in practice.)

Fixes: 8b2bd3ae5cd1 ("Add new variants of spawnPipe functions with encoding support")

#### [X.H.DynamicLog: Move UTF8 encoding from dynamicLogString to xmonadPropLog'](../commit/cdd6ee57b310b859ce7b0dbf2a766e29c2a9a86f)

For many (10+) years, we had a cascade of ugly workarounds:

 * X.U.Run.spawnPipe returned a binary handle, so the String passed to
   it must have been encoded by C.B.UTF8.String.encodeString

 * X.H.DynamicLog.dynamicLogString therefore returned such an encoded
   String, so one could use it directly in a hPutStrLn, but literal
   Strings wouldn't work

 * xmonadPropLog' also expected an encoded String to make it easier to
   use together with dynamicLogString, again breaking usage with String
   literals and other normal unicode Strings

Then in 1d0eaddc2530 Sibi fixed spawnPipe to return a handle usable with
normal Strings, which then obviously broke the entire cascade. But,
instead of using the opportunity to remove all the ugly workarounds, he
decided to add some more on top, so now spawnPipe with dynamicLogString
outputs doubly encoded UTF-8 and xmobar has a hack to strip this double
encoding (https://github.com/jaor/xmobar/pull/482), which only works
when XFT is in use and breaks on some long unicode codepoints. :-(

There is a better way: make everything just use normal Strings and only
encode when it goes out the wire. This means dynamicLogString can be
freely mixed with String literals, manual uses of xmonadPropLog' don't
need encodeString, and everything just works nicely.

This obviously breaks configs that used some of these pieces in
isolation (like mine), but that's a small price to pay. After all, right
now all users of spawnPipe/dynamicLogString are getting doubly encoded
UTF-8 which might or might not work in xmobar and probably breaks
horribly everywhere else, so this fix should be a clear improvement. :-)

Fixes: 1d0eaddc2530 ("Make spawnPipe to use system's locale encoding")
Fixes: https://github.com/xmonad/xmonad-contrib/issues/377
Fixes: https://github.com/jaor/xmobar/issues/476
Related: https://github.com/xmonad/xmonad-contrib/pull/334
Related: https://github.com/jaor/xmobar/pull/482

### Checklist

  - n/a I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,6 +50,26 @@
     - The type of `getWorkspaceNames` was changed to fit into the new `ppRename`
       field of `PP`.
 
+  * `XMonad.Hooks.DynamicLog` and `XMonad.Util.Run`
+
+    - `spawnPipe` no longer uses binary mode handles but defaults to the
+      current locale encoding instead.
+
+      `dynamicLogString`, the output of which usually goes directly into such
+      a handle, no longer encodes its output in UTF-8, but returns a normal
+      `String` of Unicode codepoints instead.
+
+      When these two are used together, everything should continue to work as
+      it always has, but in isolation behaviour might change.
+
+      (To get the old `spawnPipe` behaviour, `spawnPipeWithNoEncoding` can now
+      be used, and `spawnPipeWithUtf8Encoding` was added as well to force
+      UTF-8 regardless of locale. These shouldn't normally be necessary, though.)
+
+    - `xmonadPropLog` and `xmonadPropLog'` now encode the String in UTF-8.
+      Again, no change when used together with `dynamicLogString`, but other
+      uses of these in user configs might need to be adapted.
+
 ### New Modules
 
   * `XMonad.Util.Hacks`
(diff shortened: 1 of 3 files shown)


psibi (approved):
LGTM. Thanks for fixing this in a much better way!

PRESFIL:
Working! See [answer](https://github.com/xmonad/xmonad-contrib/issues/377#issuecomment-780720503).

TheMC47:
This also solves another issue I had, where xmobar couldn't correctly parse a combination of `fn` and `raw` tags with a special font. Awesome :tada:

liskin:
Oh yeah, that's a good point possibly worth mentioning somewhere. xmobarRaw wouldn't work without this.

PRESFIL:
Any updates on this PR?

liskin:
@PRESFIL I've been busy hacking on other projects, and I'd like to at least have a look at #465 before merging this, so I'd expect this to be merged hopefully next week-ish.

slotThe:
Rebased and merged, see #481 for reasons to do this before #465

PR #474, opened 2021-03-09
New layout modifier: XMonad.Layout.FixedAspectRatio

TheMC47:
### Description

Layout modifier for user-provided per-window aspect ratios.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -130,6 +130,10 @@
 
 ### New Modules
 
+  * `XMonad.Layout.FixedAspectRatio`
+
+    Layout modifier for user provided per-window aspect ratios.
+
   * `XMonad.Hooks.TaffybarPagerHints`
 
     Add a module that exports information about XMonads internal state that is
--- a/XMonad/Layout/FixedAspectRatio.hs
+++ b/XMonad/Layout/FixedAspectRatio.hs
@@ -1,0 +1,161 @@
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE PatternGuards #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Layout.FixedAspectRatio
+-- Copyright   :  (c) Yecine Megdiche <yecine.megdiche@gmail.com>
+-- License     :  BSD3-style (see LICENSE)
+--
+-- Maintainer  :  Yecine Megdiche <yecine.megdiche@gmail.com>
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- Layout modifier for user provided per-window aspect ratios.
+--
+-----------------------------------------------------------------------------
+
+module XMonad.Layout.FixedAspectRatio
+  (
+    -- * Usage
+    -- $usage
+    fixedAspectRatio
+  , FixedAspectRatio
+  , ManageAspectRatio(..)
+  , doFixAspect
+  ) where
+
+
+import           Control.Arrow
+import qualified Data.Map                      as M
+import           Data.Ratio
+
+import           XMonad
+import           XMonad.Actions.MessageFeedback
+import           XMonad.Layout.Decoration
+import           XMonad.Layout.LayoutHints
+
+-- $usage
+-- You can use this module with the following in your @~\/.xmonad\/xmonad.hs@:
+--
+-- > import XMonad.Layout.FixedAspectRatio
+-- Then add it to your layout:
+--
+-- > myLayout = fixedAspectRatio (0.5, 0.5) $ Tall 1 (3/100) (1/2)  ||| Full ||| etc..
+-- > main = xmonad def { layoutHook = myLayout }
+--
+-- Which will center the (eventually) shrinked windows in their assigned
+-- rectangle.
+--
+-- For a layout modifier that automatically sets the aspect ratio
+-- depending on the size hints (for example for programs like mpv),
+-- see "XMonad.Layout.LayoutHints"
+--
+-- See "XMonad.Doc.Extending#Editing_the_layout_hook" for more info on
+-- the 'layoutHook'.
+--
+-- You also want to add keybindings to set and clear the aspect ratio:
+--
+-- >      -- Set the aspect ratio of the focused window to 16:9
+-- >   ,((modm, xK_a), withFocused $ sendMessage . FixRatio (16 / 9))
+-- >
+-- >      -- Clear the aspect ratio from the focused window
+-- >   ,((modm .|. shiftMask, xK_a), withFocused $ sendMessage . ResetRatio)
+--
+-- There's one caveat: to keep the usage of the modifier simple, it
+-- doesn't remove a window from its cache automatically. Which means
+-- that if you close a program window that has some fixed aspect ratios
+-- and relaunch it, sometimes it'll still have the fixed aspect ratio.
+-- You can try to avoid this by changing they keybinding used to kill
+-- the window:
+--
+-- >  , ((modMask .|. shiftMask, xK_c), withFocused (sendMessage . ResetRatio) >> kill)
+--
+-- See "XMonad.Doc.Extending#Editing_key_bindings" for more info
+-- on customizing the keybindings.
+--
+-- This layout also comes with a 'ManageHook' 'doFixAspect' to
+-- automatically fix the aspect ratio:
+--
+-- > myManageHook = composeOne [
+-- >   title =? "Netflix" <||> className =? "vlc" --> doFixAspect (16 / 9)
+-- >   ...
+-- > ]
+--
+-- Check "XMonad.Doc.Extending#Editing_the_manage_hook" for more information on
+-- customizing the manage hook.
+
+-- | Similar to 'layoutHintsWithReplacement', but relies on the user to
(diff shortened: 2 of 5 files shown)


CHANGES.md
```diff
@@ -47,6 +47,11 @@
 
 ### New Modules
 
+  * `XMonad.Layout.FixedAspectRatio`
+
+    Layout modifier for user provided per-window aspect ratios.
+
+
```
slotThe, line 54:
```suggestion
```

XMonad/Layout/FixedAspectRatio.hs
```diff
@@ -0,0 +1,135 @@
...
+-- (0, 0) places the window at the top left, (1, 0) at the top right,
+-- (0.5, 0.5) at the center, etc.
+fixedAspectRatio
+  :: (Double, Double) -> l a -> ModifiedLayout FixedAspectRatio l a
+fixedAspectRatio = ModifiedLayout . FixedAspectRatio
+
+newtype FixedAspectRatio a = FixedAspectRatio (Double, Double)
+  deriving (Read, Show)
+
+newtype SavedRatios = SavedRatios { unRatios:: M.Map Window Rational }
```
slotThe, line 81:
```suggestion
newtype SavedRatios = SavedRatios { unRatios :: M.Map Window Rational }
```

slotThe (approved):
LGTM!

TheMC47:
Thanks for the review @slotThe! I've been trying this for a while, and it mostly works good, except for a minor quirk: it doesn't remove a killed window from the cache. We could instruct the users to do that manually before killing a window because I'm not sure if it's worth  it to add an event hook for this

TheMC47:
There was no reason to have the layout not be pure, so I removed the state and simplified the layout

TheMC47:
I updated the docs by talking about the little corner case. It doesn't bother me much tbh, but if anyone else complains I'll add an event hook

XMonad/Layout/FixedAspectRatio.hs
```diff
@@ -0,0 +1,123 @@
...
+-- | Calculates the new width and height so they respect the
+-- aspect ratio.
+adj :: (Dimension, Dimension) -> Rational -> (Dimension, Dimension)
+adj (w, h) ar | ar' > ar  = (ceiling $ fi h * ar, h)
+              | otherwise = (w, ceiling $ fi w / ar)
+  where ar' = fi w % fi h
+
+--- Message handling
+data ManageAspectRatio =
+    Fix Rational Window -- ^ Set the aspect ratio for the window
```
liskin, line 119: If this module is meant to bude used unqualified (as most modules probably are), I'd consider `FixAspect` and `ResetAspect` to be clearer and less likely to conflict with other modules.

liskin: Unrelated idea: maybe a managehook helper `doFixAspect` might be a useful addition. Should probably use `sendMessageWithNoRefresh` to avoid refreshing before the window gets managed.

XMonad/Layout/FixedAspectRatio.hs
```diff
@@ -0,0 +1,123 @@
...
+-- You also want to add keybindings to set and clear the aspect ratio:
+--
+-- >      -- Set the aspect ratio of the focused window to 16:9
+-- >   ,((modm, xK_a), withFocused $ sendMessage . Fix (16 / 9))
+-- >
+-- >      -- Clear the aspect ratio from the focused window
+-- >   ,((modm .|. shiftMask, xK_a), withFocused $ sendMessage . Reset)
+--
+-- There's one caveat: to keep the usage of the modifier simple, it
+-- doesn't remove a window from its cache automatically. Which means
```
liskin, line 64: I was going to suggest to deal with this but I noticed this problem plagues many other modules as well, even ones like BoringWindows that should really know better. :-(
(I remember fixing a couple memory leaks back in 2009/2010 when I started using xmonad, but I don't use any of these modules so I didn't bother. It might still be a problem, but other people have different usage patterns, higher tolerance to breakage, and likely also smaller uptimes.)

On the other hand it seems that something like this may take care of the garbage collection:

```haskell
  modifyLayoutWithUpdate (FixedAspectRatio ratios placement) w r = do
    let ws = W.integrate' (W.stack w)
    let ratios' = M.filterWithKey (\k _ -> k `elem` ws) ratios
    let m' = FixedAspectRatio ratios' placement
    (, Just m') <$> runLayout w r
```

It may be too eager (e.g. when used as a SubLayout, not that I can imagine a scenario when that would make sense), so this might be a safer (and more complicated):

```haskell
  modifyLayoutWithUpdate (FixedAspectRatio ratios placement) w r = do
    ws <- gets (W.index . W.view (W.tag w) . windowset)
    …
```

liskin (approved):
Handly little module. Added a few suggestions, but nothing major.

XMonad/Layout/FixedAspectRatio.hs
TheMC47: And I suppose layouts have no way of knowing when windows are killed since they don't see the events? 

I could fix this here, but since this is a problem, why don't we think of a better alternative? I was thinking of adding a combinator that installs an handleEvent that simply broadcasts a message (for example `WindowKilled w`). That way we can patch layouts easily. Would this work? And which event should we listen to?

XMonad/Layout/FixedAspectRatio.hs
TheMC47: I like the idea with the manageHook :+1:

XMonad/Layout/FixedAspectRatio.hs
liskin: Layouts see only [events that aren't handled by the core](https://github.com/xmonad/xmonad/blob/4a0b166998a838d91f002991da82507a2d04d3bb/src/XMonad/Main.hs#L426) but [DestroyWindowEvent obviously is handled](https://github.com/xmonad/xmonad/blob/4a0b166998a838d91f002991da82507a2d04d3bb/src/XMonad/Main.hs#L315) so layouts don't see this one.

TheMC47: I've opened https://github.com/xmonad/xmonad/pull/284 to broadcast `DestroyWindowEvent`. If we decide against it, then I'll add one of your suggestions.

liskin: Yeah I've seen that, I just wanted to focus on the 0.17 stuff so I'd rather not dig deep into this just yet.

TheMC47:
Just rebased

TheMC47:
I added another message to toggle the ratios, which is probably more useful. Not sure if we keep the other Reset and Fix messages, but I don't think it hurts.

XMonad/Layout/FixedAspectRatio.hs
```diff
@@ -0,0 +1,161 @@
...
+    applyHint (win, r@(Rectangle x y w h)) =
+      let ar       = M.lookup win ratios
+          (w', h') = maybe (w, h) (adj (w, h)) ar
+      in  (win, if isInStack s win then Rectangle x y w' h' else r)
+
+  pureModifier _ _ _ xs = (xs, Nothing)
+
+  handleMess (FixedAspectRatio ratios placement) mess
+    | Just DestroyWindowEvent { ev_window = w } <- fromMessage mess
+    = return . Just $ FixedAspectRatio (deleted w) placement
```
TheMC47, line 122: I'm testing this with https://github.com/xmonad/xmonad/pull/284, and it doesn't seem to work. Is there somewhere else I also need to handle `DistroyWindowEvent`?

liskin: If you're using NewSelect from LayoutCombinators, then it needs to be added there. Or use https://github.com/xmonad/xmonad/pull/281

TheMC47: That's the thing, I'm not using any of those... I'll need to debug what's the problem here

TheMC47: alright, https://github.com/xmonad/xmonad/pull/284/commits/eda5dfbe288c4a0a0174dbfd65027386cc38266f fixes it

TheMC47:
(just rebasing)

TheMC47:
Any further comments regarding this PR? If not, I'd merge it in the next days

liskin (approved):
No further comments, feel free to squash the fixups and merge.

slotThe (commented):
There was someone on IRC asking about making the exact window dimensions configurable (I believe in connections with Tor browser suggesting that).  I don't know how generally useful this is, but if it can go into this module without too much modification it might be worth a shit.  If not then I suppose we can also just wait for someone to hack this in themselves.

TheMC47:
> There was someone on IRC asking about making the exact window dimensions configurable (I believe in connections with Tor browser suggesting that). I don't know how generally useful this is, but if it can go into this module without too much modification it might be worth a shit. If not then I suppose we can also just wait for someone to hack this in themselves.

I saw the conversation, and I don't think it's a good idea to include it here, for a couple of reasons: fixing the size of a specific window might require making other windows smaller, which isn't at all handled here (windows can only shrink in size, never getting bigger). Specifying an upper bound on height/width might be useful though, but I don't see it implemented in this module (it should be easy to do separately though, and when that's done we could see if we can improve the design to avoid code duplication)

PR #486, opened 2021-03-19
Fix issue with colors in X.A.TreeSelect

exorcist365:
### Description

Green and blue were swapped.

### Checklist

### DIFF
--- a/XMonad/Actions/TreeSelect.hs
+++ b/XMonad/Actions/TreeSelect.hs
@@ -137,8 +137,8 @@ import Graphics.X11.Xrender
 -- > def = TSConfig { ts_hidechildren = True
 -- >                , ts_background   = 0xc0c0c0c0
 -- >                , ts_font         = "xft:Sans-16"
--- >                , ts_node         = (0xff000000, 0xff50d0db)
--- >                , ts_nodealt      = (0xff000000, 0xff10b8d6)
+-- >                , ts_node         = (0xff000000, 0xff50dbd0)
+-- >                , ts_nodealt      = (0xff000000, 0xff10d6b8)
 -- >                , ts_highlight    = (0xffffffff, 0xffff0000)
 -- >                , ts_extra        = 0xff000000
 -- >                , ts_node_width   = 200
@@ -220,8 +220,8 @@ instance Default (TSConfig a) where
     def = TSConfig { ts_hidechildren = True
                    , ts_background   = 0xc0c0c0c0
                    , ts_font         = "xft:Sans-16"
-                   , ts_node         = (0xff000000, 0xff50d0db)
-                   , ts_nodealt      = (0xff000000, 0xff10b8d6)
+                   , ts_node         = (0xff000000, 0xff50dbd0)
+                   , ts_nodealt      = (0xff000000, 0xff10d6b8)
                    , ts_highlight    = (0xffffffff, 0xffff0000)
                    , ts_extra        = 0xff000000
                    , ts_node_width   = 200
--- a/XMonad/Actions/TreeSelect.hs
+++ b/XMonad/Actions/TreeSelect.hs
@@ -161,8 +161,8 @@ import Graphics.X11.Xrender
 -- white       = 0xffffffff
 -- black       = 0xff000000
 -- red         = 0xffff0000
--- blue        = 0xff00ff00
 -- green       = 0xff0000ff
+-- blue        = 0xff00ff00
 -- transparent = 0x00000000
 -- @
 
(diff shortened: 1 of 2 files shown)


exorcist365:
If you ask me RGB = (red green blue) makes more sense than RGB = (red blue green)

XMonad/Actions/TreeSelect.hs
```diff
@@ -656,7 +656,7 @@ drawStringXMF display window visual colormap gc font col x y text = case font of
 -- Note that it uses short to represent its components
 fromARGB :: Pixel -> XRenderColor
 fromARGB x = XRenderColor (fromIntegral $ 0xff00 .&. shiftR x 8)  -- red
-                          (fromIntegral $ 0xff00 .&. x)           -- green
-                          (fromIntegral $ 0xff00 .&. shiftL x 8)  -- blue
+                          (fromIntegral $ 0xff00 .&. shiftL x 8)  -- green
+                          (fromIntegral $ 0xff00 .&. x)           -- blue
```
liskin, line 660: Oh crap. :-(
This is actually a bug in https://hackage.haskell.org/package/X11-xft-0.3.1/src/Graphics/X11/Xrender.hsc, but that package is probably unmaintained, so we may need to work around it here, although it'd be nice to at least try contacting Clemens. :-/

Turns out the same problem is in xmobar: when alpha is less than 255 and bgcolor is green, the xft-drawn parts of xmobar are blue and the rest is green. It's really silly.

We definitely need a comment explaining why the colors are swapped here.

XMonad/Actions/TreeSelect.hs
```diff
@@ -220,8 +220,8 @@ instance Default (TSConfig a) where
     def = TSConfig { ts_hidechildren = True
                    , ts_background   = 0xc0c0c0c0
                    , ts_font         = "xft:Sans-16"
-                   , ts_node         = (0xff000000, 0xff50d0db)
-                   , ts_nodealt      = (0xff000000, 0xff10b8d6)
+                   , ts_node         = (0xff000000, 0xff50dbd0)
+                   , ts_nodealt      = (0xff000000, 0xff10d6b8)
```
liskin, line 224: The problem only affects foreground colors so I don't think this change is desirable.

XMonad/Actions/TreeSelect.hs
```diff
@@ -161,8 +161,8 @@ import Graphics.X11.Xrender
 -- white       = 0xffffffff
 -- black       = 0xff000000
 -- red         = 0xffff0000
--- blue        = 0xff00ff00
--- green       = 0xff0000ff
+-- blue        = 0xff0000ff
+-- green       = 0xff00ff00
```
liskin, line 165: Can we swap these two lines? :-)

exorcist365: Ah. Yeah I don't even know why I changed these.

XMonad/Actions/TreeSelect.hs
exorcist365: As far as I tested this fixes the foreground issue and does not break anything, but I might try to fix the Xrender thing as well.

liskin: Yeah, I tested this as well, because it all seemed very strange, and that led to me to the discovery of the root cause in X11-xft and the xmobar issue as well. If you decide to contact Clemens, feel free to Cc me. In any case, we need a comment here explaining what's going on.

exorcist365: What comment should I add?

liskin: One that summarizes what was said in this comment thread. If you don't know what to add, I guess I can do it later. :-/

XMonad/Actions/TreeSelect.hs
```diff
@@ -654,6 +654,8 @@ drawStringXMF display window visual colormap gc font col x y text = case font of
 -- | Convert 'Pixel' to 'XRenderColor'
 --
 -- Note that it uses short to represent its components
+--
+-- This fixes a bug with font rendering, caused by the haskell bindings for xorg-xft.
```
liskin, line 658: Oh come on, this doesn't even mention colors! Why don't you just take all those comments of ours and summarize them into a couple sentences. No need to invent new words, change package names, or anything.

liskin:
I cleaned this up and merged. I also sent a message to Clemens about the bug in X11-xft.

exorcist365:
Thanks, I was a bit busy the last 2 days and haven't visited github. I was going to do it today but seems like you already did it.

PR #490, opened 2021-03-22
X.U.Hacks: Add "Stacking trays (trayer) above panels (xmobar)"

liskin:
### Description

Placing `trayer` on top of `xmobar` is somewhat tricky:

- they both should be lowered to the bottom of the stacking order to avoid overlapping fullscreen windows

- the tray needs to be stacked on top of the panel regardless of which happens to start first

`trayerAboveXmobarEventHook` (and the more generic `trayAbovePanelEventHook`) is an event hook that ensures the latter: whenever the tray lowers itself to the bottom of the stack, it checks whether there are any panels above it and lowers these again.

### Checklist

### DIFF
--- a/XMonad/Util/Hacks.hs
+++ b/XMonad/Util/Hacks.hs
@@ -27,12 +27,17 @@ module XMonad.Util.Hacks (
   -- * Java Hack
   -- $java
   javaHack,
+
+  -- * Stacking trays (trayer) above panels (xmobar)
+  -- $raiseTrayer
+  trayerAboveXmobarEventHook,
+  trayAbovePanelEventHook,
   ) where
 
 
 import XMonad
 import Data.Monoid (All(All))
-import Control.Monad (when)
+import Control.Monad (when, filterM)
 import System.Posix.Env (putEnv)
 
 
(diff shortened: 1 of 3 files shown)


liskin:
(I'm fully aware that using xmobar with `overrideRedirect = True` currently requires restarting it after every rescreen, and I intend to submit another PR for that soon.)

XMonad/Util/Hacks.hs
```diff
@@ -92,3 +99,47 @@ javaHack conf = conf
...
+
+-- | Whenever a tray window lowers itself to the bottom of the stack, look for
+-- any panels above it and lower these.
+trayAbovePanelEventHook
+  :: Query Bool -- ^ tray
+  -> Query Bool -- ^ panel
+  -> (Event -> X All) -- ^ event hook
+trayAbovePanelEventHook trayQ panelQ ConfigureEvent{ev_window = w, ev_above = a} | a == none = do
+  -- when tray is lowered to the bottom of stack (a == none),
+  -- put all panels that are above it below
```
TheMC47, line 137: Maybe "move all panels below it"?

TheMC47 (approved):
LGTM

XMonad/Util/Hacks.hs
liskin: Hm, do I need this comment at all? It's saying the same thing as the haddock for the function itself (which I wrote today and then forgot to take another look at the comment in code).

liskin: I dropped it. And then I added a usage to the the doc which I forgot.

PR #492, opened 2021-03-23
X.H.ManageDocks: React to strut updates of override_redirect docks

liskin:
### Description

We only requested PropertyChange events from docks in the `manageDocks` manageHook, but that only gets called for normal windows, not override_redirect ones. Therefore, xmobar in its default configuration wouldn't get its struts refreshed on changes. This resulted in gaps not updating after xmobar repositions itself on xrandr changes.

If one wanted to handle that repositioning in xmonad, it was possible to configure xmobar with `overrideRedirect = False`, which is meant for window managers with [proper EWMH stacking order support][1], so in xmonad it resulted in xmobar covering fullscreen windows. That can be worked around by adding `checkDock --> doLower` to manageHook, but it starts to smell of too many workarounds.

[1]: https://specifications.freedesktop.org/wm-spec/wm-spec-1.5.html#STACKINGORDER

The fix is to request PropertyChange events for all windows that we treat as docks.

Related: https://github.com/xmonad/xmonad-contrib/pull/406
Related: https://github.com/xmonad/xmonad-contrib/pull/490

### Checklist

### DIFF
--- a/XMonad/Util/ExtensibleState.hs
+++ b/XMonad/Util/ExtensibleState.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE PatternGuards #-}
 -----------------------------------------------------------------------------
 -- |
@@ -22,6 +23,7 @@ module XMonad.Util.ExtensibleState (
                               , get
                               , gets
                               , modified
+                              , modifiedM
                               ) where
 
 import Data.Typeable (typeOf,cast)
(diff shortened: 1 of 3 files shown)


XMonad/Util/ExtensibleState.hs
```diff
@@ -121,8 +123,11 @@ remove :: (ExtensionClass a, XLike m) => a -> m ()
 remove wit = modifyStateExts $ M.delete (show . typeOf $ wit)
 
 modified :: (ExtensionClass a, Eq a, XLike m) => (a -> a) -> m Bool
-modified f = do
+modified = modifiedM . (pure .)
```
geekosaur, line 126: Arguably this refactor should be a separate PR.

geekosaur (approved):
Other than that (and a similar feeling about the `deleteFromStru[c]tCache` correction), this seems okay.

XMonad/Util/ExtensibleState.hs
liskin: I can definitely turn this into a commit on its own but a separate PR seems excessive.

liskin: > Other than that (and a similar feeling about the `deleteFromStru[c]tCache` correction), this seems okay.

I split both into separate commits.

PR #496, opened 2021-03-25
X.L.Magnifier: Add `magnify` combinator, magnification at `n` windows

slotThe:
### Description

Add the ability to only start magnifying when a certain number of windows is reached.  As a byproduct, cleans up the rather embarassing mistake made in bb205e92051de531c6c8b7eb1d692779969b6710 to not just transform everything into a record.

Also add a more general `magnify` function that subsumes (almost) all previously known combinators.  This needs exporting of some previously internal (but not crucial to the implementation) types, but the internals are still changeable easily enough. 

### Checklist

### DIFF
--- a/XMonad/Layout/Magnifier.hs
+++ b/XMonad/Layout/Magnifier.hs
@@ -1,4 +1,7 @@
-{-# LANGUAGE DeriveDataTypeable, MultiParamTypeClasses, TypeSynonymInstances, PatternGuards #-}
+{-# LANGUAGE DeriveDataTypeable         #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE PatternGuards              #-}
+{-# LANGUAGE TypeSynonymInstances       #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Layout.Magnifier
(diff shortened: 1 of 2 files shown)


XMonad/Layout/Magnifier.hs
```diff
@@ -39,6 +42,8 @@ module XMonad.Layout.Magnifier
 
       -- * Messages and Types
       MagnifyMsg (..),
+      MagnifyThis(..),
+      Start(..),
```
liskin, line 46: I'm a bit concerned about this: `Start`, `On`, `Off` don't seem like good/safe names to export. Also, I'm not sure if renaming this for easier understanding of the "constructor" function is worth having a name that's confusing in the actual implementation, where it signifies the current state, not initial state.

I'd almost think that consistently naming the record fields, just using Bool for this one, and exporting the fields may be better long term. We may add a default instance and not export the masterWins field, but note that both Tall and ResizableTall make it possible to set the initial number of masters, so we actually do want to export this one too!

(And then it becomes questionable whether the `magnify` function is worth it.)

liskin (commented):
Slight concern about 3f1889aebd97516f17582d0171c0632552f040a4, the rest is all right.

XMonad/Layout/Magnifier.hs
slotThe: Good point, thanks.  `On` and `Off` are probably a little bit too generic to export; using bool and then translating to the internal names sounds good to me (I do think it's nice that we have proper names for this internally and are not just using bool).

I'm kind of ambivalent about just exporting all of `Magnify`.  Best case we can deprecate almost all functions and tell people to use the data constructor directly instead (like X.L.Spacing essentially did), worst case there is a pressing need to change the internals for some new feature or fix and we can't do it because it'll break people's configurations.

liskin: I don't have any strong preference about (not) exporting either, so… can't help you there, or at least not immediately. :-)

liskin (approved):
Looks good to go.

PR #499, opened 2021-03-28
New module: XMonad.Prelude

slotThe:
### Description

As discussed on IRC, this is an idea for a prelude-like module for xmonad-contrib; hoping it'll improve the developer experience somewhat.  There were some concerns, so I decided to first post this here before I go about cleaning up imports.  Alternative preludes can seem pretty alien to new contributors and people unfamiliar with the code base in general, but since this (at this point, anyways) is mostly just "let's re-export the most important `base` modules". 

I wonder: As this is mostly about internal things, should we mention this in the changes file at all?  

### Checklist


geekosaur (commented):
Had you considered `XMonad.Docs.Developing`?

geekosaur:
Also, I'd been under the impression you had more things in mind than just `fi`; a  new module seems over the top just for that.

slotThe:
> Had you considered `XMonad.Docs.Developing`?

Oh that's a good point, thanks!

> Also, I'd been under the impression you had more things in mind than just `fi`; a new module seems over the top just for that.
> 
> To be more clear, I'd expected at least some comments with other things to consider, to start the conversation.

I did consider some other functions that are defined in multiple modules right now.  However, on IRC @liskin expressed some doubts about a custom prelude being difficult to work with for new contributors, so I thought at least getting rid of very commonly used base imports that people will know would work as a tradeoff (should people like this idea I would of course go through the fun exercise of doing this for all of contrib).   

Some functions I would personally consider adding because I've seen them in a few places:

  - `(!?) :: [a] -> Int -> Maybe a`
  -  `reverseStack :: Stack a -> Stack a`
  - `chunksOf :: Int -> [a] -> [[a]]`
  - `(.:) :: (a -> b) -> (c -> d -> a) -> c -> d -> b`

liskin:
> * `reverseStack :: Stack a -> Stack a`

There's `XMonad.Util.Stack.reverseS` and `reverseZ` already, better use that one.

:+1: for the rest.

XMonad/Layout/LayoutBuilder.hs
```diff
@@ -58,10 +58,23 @@ module XMonad.Layout.LayoutBuilder (
...
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
```
liskin, line 75: what's this? :-D

slotThe: Oh ^^'  I think that was `sed` not _quite_ doing what I wanted it to

slotThe:
@geekosaur what do you think of the other functions added/the use of `XMonad.Prelude` instead of some of the other base imports now?

XMonad/Actions/TiledWindowDragging.hs
```diff
@@ -97,5 +97,3 @@ performWindowSwitching win = do
 -- | shorthand for fromIntegral
```
geekosaur, line 97: We have another `fi` hiding here, it turns out.

XMonad/Config/Dmwit.hs
```diff
@@ -19,6 +16,7 @@ import System.Process
 
 -- xmonad core
 import XMonad
+import XMonad.Prelude
```
geekosaur, line 19: This is part of `xmonad-contrib` and perhaps shouldn't be part of a block whose comment is "xmonad core".

XMonad/Prelude.hs
```diff
@@ -0,0 +1,54 @@
...
+    chunksOf,
+    (.:),
+    (!?),
+) where
+
+import Control.Applicative as Exports
+import Control.Monad       as Exports
+import Data.Bool           as Exports
+import Data.Char           as Exports
+import Data.Foldable       as Exports
```
geekosaur, line 26: This is going to collide with `Data.List`, and it's not clear to me which one "wins" (there "should" be a multiple definition error in the export line, but the way GHC checks things, there won't be). This shouldn't affect behavior but may affect performance in some cases, since Foldable is more general and therefore less optimized… although they're lists, so they're not going to perform especially well anyway.

Come to think of it, if `Data.List` wins, that could cause some breakage for things expecting `Data.Foldable` to work on e.g. `Set`. But I assume CI would catch that.

XMonad/Prelude.hs
```diff
@@ -0,0 +1,54 @@
...
+import Data.Function       as Exports
+import Data.Functor        as Exports
+import Data.List           as Exports
+import Data.Maybe          as Exports
+import Data.Monoid         as Exports
```
geekosaur, line 31: I'm more generally worried about re-exporting too much for the same reason as above — the more reexports, the greater the chance of name collisions.

XMonad/Prelude.hs
```diff
@@ -0,0 +1,54 @@
...
+
+-- | Short for 'fromIntegral'.
+fi :: (Integral a, Num b) => a -> b
+fi = fromIntegral
+
+-- | Given a maximum length, splits a list into sublists
+--
+-- >>> chunksOf 5 (take 30 $ repeat 'a')
```
geekosaur, line 39: I'm still trying to figure out if we want to depend on `split` for this instead of reinventing one function.

geekosaur (approved):
I think that'll do for a start, and we can add more things as we run across them.

XMonad/Prelude.hs
geekosaur: Come to think of it, `Data.List` these days is mostly reexports from `Data.Foldable` so the above collision can't really occur (or can, but harmlessly). I still wonder if having both is safe in the long term.

slotThe: Yeah most of the things that are applicable to both lists and foldables is just re-exported from `Data.Foldable` now.  In general I think most of the `base` libraries work with each other more often than not.  There are certainly notable exceptions (like `Control.Category` having an instance of `(.)` that's incompatible with `base`s) but since a good number of modules simply import `XMonad.Prelude` unqualified and `xmonad-contrib` still compiles, everything seems fine here.

There's also the fact that `base` only changes with new versions of GHC, so if something were to break then I suspect we'd be the first to know and could fix things before users upgrade.

XMonad/Prelude.hs
slotThe: I was thinking about this but ultimately decided against pulling in a dependency (even though `split` is really small) for a single function.  If we find more functions that are just taken from that package then I think the incentive would be much higher for us to just add it as a dependency

XMonad/Prelude.hs
slotThe: Found one more: `X.P.Shell.split` is something like `Data.List.Split.splitOn . (:[])` with a special case for `[]`, returning `[]` instead of `[""]`.

XMonad/Prelude.hs
slotThe: Found one more: `X.P.Shell.split` is something like `Data.List.Split.splitOn . (:[])` with a special case for `[]`, returning `[]` instead of `[""]`.

slotThe:
I rebased the changes and squashed everything down.

The main question at this point is whether we want to pull in `split` as a dependency and expose it via X.Prelude (this would—at least—replace our `chunksOf` and some function in X.P.Shell) or if we would rather just define this ourselves.

@geekosaur @liskin @TheMC47  opinions?

TheMC47:
Thanks for your work! I think there's no harm in pulling split, it's tiny

geekosaur:
I was the one who originally asked about split, but I'm of two minds on it now: not quite enough use to justify it as a dependency, really.

liskin:
I have no strong opinions here. The split package is still maintained and has no dependencies other than base, so it's in some sense better than all the other deps we have, so I think including it is totally okay.

slotThe:
Much like the rest of you, I don't have any strong opinions on this one :)

Pulling in `split` is probably not very taxing, but that's a decision that we can also make at a later point, independently of this pr.

slotThe:
I added a small note in X.D.Developing (oh boy does that module need an overhaul).  

Other than that I think this is ready to go now; unless you have all changed your minds and definitely, positively want to use `Data.List.Split.chunksOf` now!

liskin (approved):

PR #500, opened 2021-03-29
New Module: XMonad.Prompt.OrgMode

slotThe:
### Description

A new prompt for interacting with [org-mode](https://orgmode.org/).
It can be used to quickly save TODOs, NOTEs, and the like with the
additional capability to schedule/deadline a task, or use the
primary selection as the contents of the note.

Right now `orgPromptPrimary` still relies on `X.U.XSelection.getSelection`, but the long-term plan would be to also rewrite that module to be less insane.  I do, however, think that the module in its current form is very benefitial already so I'm releasing it like this for now. 

### Checklist

### DIFF
--- a/XMonad/Prompt/OrgMode.hs
+++ b/XMonad/Prompt/OrgMode.hs
@@ -76,7 +76,8 @@ above you can write
 
 There is also some scheduling and deadline functionality present.  They
 are initiated by entering @+s@ or @+d@ into the prompt respectively,
-followed by a date.  Any of the following are valid dates:
+followed by a date and a time of day.  Any of the following are valid
+dates:
 
   - today
   - tomorrow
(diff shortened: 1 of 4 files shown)


geekosaur (approved):
I don't see anything problematic here, offhand.

TheMC47:
I don't have a comment on the code, but it's a pretty useful and awesome feature :+1:

XMonad/Prompt/OrgMode.hs
```diff
@@ -0,0 +1,391 @@
...
+above, pressed it, and are now confronted with a prompt:
+
+  - @hello +s today@ would create a TODO note with the header @hello@
+    and would schedule that for today's date.
+
+  - @hello +d today 12@ schedules the note for today at 12:00.
+
+  - @hello +d today 12:30@ schedules it for today at 12:30.
+
+  - @hello +d today 12:30@ works just like above, but creates a
```
psibi, line 108: I'm not sure how it's different from the above ? Even the above one seems to be a deadline ?

XMonad/Prompt/OrgMode.hs
```diff
@@ -0,0 +1,391 @@
...
+The time is specified in the @HH:MM@ format.  The minutes may be
+omitted, in which case @00@ will be substituted.
+
+A few examples are probably in order.  Suppose we have bound the key
+above, pressed it, and are now confronted with a prompt:
+
+  - @hello +s today@ would create a TODO note with the header @hello@
+    and would schedule that for today's date.
+
+  - @hello +d today 12@ schedules the note for today at 12:00.
```
psibi, line 104: Probably you can use the word `deadline` here to avoid confusion ?

psibi (approved):
This module looks amazing! Thanks for the nice documentation too! Couple of minor comments, but LGTM apart from that.

XMonad/Prompt/OrgMode.hs
slotThe: Whoops, you're right.  The two above should read `+s` instead of `+d`.  That way "schedules" also makes sense :)

slotThe:
Thanks for the reviews!  I will merge this this evening if there are no further objetcions

Issue #420, opened 2020-12-09
X.L.IndependentScreens' `marshallSort` always sorts lexically


geekosaur:
fizzie 16:16:42
I'm thinking the problem with IndependentScreens' marshallSort may be exactly that it handles marshalling and unmarshalling the physical workspace names, so that the sort function just sees the plain "1", "2", ..., but then the default sort function will fail to find indices for any of those because the config's workspace list will be "0_1", "0_2", ... and so on.

fizzie 16:17:53
In which case it would in fact fall back to lexicographically sorting the names, because of mconcat [indexCompare `on` wsIndex, compare], and indexCompare Nothing Nothing = EQ.

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

dmwit:
I guess I don't know the context of this, but my gut reaction is that this is why `workspaces'` exists: to get a list of the virtual workspaces available. So whatever sort function you're passing to `marshallSort` should use `workspaces'` instead of `workspaces`, and then this should Just Work?

liskin:
> I guess I don't know the context of this, but my gut reaction is that this is why `workspaces'` exists: to get a list of the virtual workspaces available. So whatever sort function you're passing to `marshallSort` should use `workspaces'` instead of `workspaces`, and then this should Just Work?

This would require substantial changes in `XMonad.Util.WorkspaceCompare` to allow overriding the `workspaces` it uses. Wouldn't it be better if `marshallPP` just worked with the default `ppSort`: `getSortByIndex`, which uses the real (not virtual) workspace names, and then have an optional pair of primed functions that work with virtual names for anyone who needs those?

The approach taken in #501 is probably also fine: keep the existing behaviour in `marshallPP` and `marshallSort` and add a primed pair of functions that work with real workspace names and clearly document that with the default `ppSort`, these should be used.

oogeek:
My way in the pull request is to keep the physical workspaces and let `getSortByIndex` or whatever sort it.
However, it is also find if we first sort it with `getSortByXXX`, then unmarshall it. I don't think it will make a big difference.

oogeek:
Any suggestions?

PR #438, opened 2020-12-31
Extend and fix issues with XMonad.Layout.IndependentScreens

elkowar:
### Description

This PR adds some general improvements to XMonad.Layout.IndependentScreens.
Specifically, it fixes the type-signature of `onCurrentScreen` being simply _wrong_, adds a few utility functions which make working with IndependentScreens more ergonomic, and adds a handleEventHook which makes `_NET_ACTIVE_WINDOW` events respect independent screens.
Without that eventHook, if you had a window on a hidden workspace on screen 2 request focus while focussing a workspace on screen 1, it would display that hidden workspace from screen 2 on screen 1, ends up being very confusing. The handleEventHook uses the also newly added `focusWindow'` function to shift the focus and view the workspace on the correct screen.

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

  - [x] I updated the `CHANGES.md` file

~~- [ ] I updated the `XMonad.Doc.Extending` file (if appropriate)~~

PR #501, opened 2021-03-31
Fix bug #420, add some functions to IndependentScreens

oogeek:
### Description
So basically I add:
1. `withNthWorkspaceScreen` to make it work with dynamicWorkspces
2. `copyWindowTo` to make it work with CopyWindow
3. `marshallSort'` (minor changes) to let it work properly with getSortByIndex
4. `wsContainingCopies` to let it detect copied windows properly (with CopyWindow)
5. `withScreen` (minor changes) to allow for different workspace names for each monitor

I am not sure if some function should be put in XMonad.Layout.IndependentScreens or other places.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -609,6 +609,22 @@
     - The decoration window now sets a `WM_CLASS` property.  This allows
       other applications, like compositors, to properly match on it.
 
+  * `XMonad.Layout.IndependentScreens`
+
+    - Fixed a bug where `marshallPP` always sorted workspace names
+      lexically.  This changes the default behaviour of `marshallPP`—the
+      given `ppSort` now operates in the _physical_ workspace names.
+      The documentation of `marshallSort` contains an example of how to
+      get the old behaviour, where `ppSort` operates in virtual names,
+      back.
+
+    - Added `workspacesOn` for filtering workspaces on the current screen.
+
+    - Added `withScreen` to specify names for a given single screen.
+
+    - Added new aliases `PhysicalWindowSpace` and `VirtualWindowSpace`
+      for a `WindowSpace` for easier to read function signatures.
+
 ## 0.16
 
 ### Breaking Changes
--- a/XMonad/Layout/IndependentScreens.hs
+++ b/XMonad/Layout/IndependentScreens.hs
@@ -17,15 +17,18 @@ module XMonad.Layout.IndependentScreens (
     -- * Usage
     -- $usage
     VirtualWorkspace, PhysicalWorkspace,
+    VirtualWindowSpace, PhysicalWindowSpace,
     workspaces',
-    withScreens, onCurrentScreen,
+    withScreen, withScreens,
+    onCurrentScreen,
     marshallPP,
     whenCurrentOn,
     countScreens,
+    workspacesOn,
     -- * Converting between virtual and physical workspaces
     -- $converting
     marshall, unmarshall, unmarshallS, unmarshallW,
-    marshallWindowSpace, unmarshallWindowSpace, marshallSort
+    marshallWindowSpace, unmarshallWindowSpace, marshallSort,
 ) where
 
 -- for the screen stuff
@@ -33,7 +36,7 @@ import Control.Arrow ((***))
 import Graphics.X11.Xinerama
 import XMonad
 import XMonad.Prelude
-import XMonad.StackSet hiding (filter, workspaces)
+import qualified XMonad.StackSet as W
 import XMonad.Hooks.DynamicLog
 
 -- $usage
@@ -78,6 +81,11 @@ import XMonad.Hooks.DynamicLog
 type VirtualWorkspace  = WorkspaceId
 type PhysicalWorkspace = WorkspaceId
 
+-- | A 'WindowSpace' whose tags are 'PhysicalWorkspace's.
+type PhysicalWindowSpace = WindowSpace
+-- | A 'WindowSpace' whose tags are 'VirtualWorkspace's.
+type VirtualWindowSpace = WindowSpace
+
 -- $converting
 -- You shouldn't need to use the functions below very much. They are used
 -- internally. However, in some cases, they may be useful, and so are exported
@@ -100,13 +108,20 @@ unmarshallW = snd . unmarshall
 workspaces' :: XConfig l -> [VirtualWorkspace]
 workspaces' = nub . map unmarshallW . workspaces
 
+-- | Specify workspace names for each screen
+withScreen :: ScreenId            -- ^ The screen to make workspaces for
+           -> [VirtualWorkspace]  -- ^ The desired virtual workspace names
+           -> [PhysicalWorkspace] -- ^ A list of all internal physical workspace names
+withScreen n vws = [marshall n pws | pws <- vws]
+
+-- | Make all workspaces across the monitors bear the same names
 withScreens :: ScreenId            -- ^ The number of screens to make workspaces for
             -> [VirtualWorkspace]  -- ^ The desired virtual workspace names
             -> [PhysicalWorkspace] -- ^ A list of all internal physical workspace names
-withScreens n vws = [marshall sc pws | pws <- vws, sc <- [0..n-1]]
+withScreens n vws = concatMap (`withScreen` vws) [0..n-1]
 
 onCurrentScreen :: (VirtualWorkspace -> WindowSet -> a) -> (PhysicalWorkspace -> WindowSet -> a)
-onCurrentScreen f vws = screen . current >>= f . flip marshall vws
+onCurrentScreen f vws = W.screen . W.current >>= f . flip marshall vws
 
 -- | In case you don't know statically how many screens there will be, you can call this in main before starting xmonad.  For example, part of my config reads
 --
@@ -121,20 +136,25 @@ onCurrentScreen f vws = screen . current >>= f . flip marshall vws
 countScreens :: (MonadIO m, Integral i) => m i
 countScreens = fmap genericLength . liftIO $ openDisplay "" >>= liftA2 (<*) getScreenInfo closeDisplay
 
--- | This turns a naive pretty-printer into one that is aware of the
--- independent screens. That is, you can write your pretty printer to behave
--- the way you want on virtual workspaces; this function will convert that
--- pretty-printer into one that first filters out physical workspaces on other
--- screens, then converts all the physical workspaces on this screen to their
--- virtual names.
+-- | This turns a pretty-printer into one that is aware of the independent screens. The
+-- converted pretty-printer first filters out physical workspaces on other screens, then
+-- converts all the physical workspaces on this screen to their virtual names.
+-- Note that 'ppSort' still operates on physical (marshalled) workspace names,
+-- otherwise functions from "XMonad.Util.WorkspaceCompare" wouldn't work.
+-- If you need to sort on virtual names, see 'marshallSort'.
 --
--- For example, if you have handles @hLeft@ and @hRight@ for bars on the left and right screens, respectively, and @pp@ is a pretty-printer function that takes a handle, you could write
+-- For example, if you have have two bars on the left and right screens, respectively, and @pp@ is
+-- a pretty-printer, you could apply 'marshallPP' when creating a @StatusBarConfig@ from "XMonad.Hooks.StatusBar".
 --
--- > logHook = let log screen handle = dynamicLogWithPP . marshallPP screen . pp $ handle
--- >           in log 0 hLeft >> log 1 hRight
+-- A sample config looks like this:
+--
+-- > mySBL = statusBarProp "xmobar" $ pure (marshallPP (S 0) pp)
+-- > mySBR = statusBarProp "xmobar" $ pure (marshallPP (S 1) pp)
+-- > main = xmonad $ withEasySB (mySBL <> mySBR) defToggleStrutsKey def
+--
 marshallPP :: ScreenId -> PP -> PP
 marshallPP s pp = pp { ppRename = ppRename pp . unmarshallW
-                     , ppSort = fmap (marshallSort s) (ppSort pp) }
+                     , ppSort = (. workspacesOn s) <$> ppSort pp }
 
 -- | Take a pretty-printer and turn it into one that only runs when the current
 -- workspace is one associated with the given screen. The way this works is a
@@ -161,7 +181,7 @@ whenCurrentOn s pp = pp
     { ppSort = do
         sortWs <- ppSort pp
         return $ \xs -> case xs of
-            x:_ | unmarshallS (tag x) == s -> sortWs xs
+            x:_ | unmarshallS (W.tag x) == s -> sortWs xs
             _ -> []
     , ppOrder = \i@(wss:_) -> case wss of
         "" -> ["\0"] -- we got passed no workspaces; this is the signal from ppSort that this is a boring case
@@ -171,11 +191,31 @@ whenCurrentOn s pp = pp
         _ -> ppOutput pp out
     }
 
--- | If @vSort@ is a function that sorts 'WindowSpace's with virtual names, then @marshallSort s vSort@ is a function which sorts 'WindowSpace's with physical names in an analogous way -- but keeps only the spaces on screen @s@.
-marshallSort :: ScreenId -> ([WindowSpace] -> [WindowSpace]) -> ([WindowSpace] -> [WindowSpace])
+-- | Filter workspaces that are on current screen.
+workspacesOn :: ScreenId -> [PhysicalWindowSpace] -> [PhysicalWindowSpace]
+workspacesOn s = filter (\ws -> unmarshallS (W.tag ws) == s)
+
+-- | @vSort@ is a function that sorts 'VirtualWindowSpace's with virtual names.
+-- @marshallSort s vSort@ is a function which sorts 'PhysicalWindowSpace's with virtual names,
+-- but keeps only the 'WindowSpace'\'s on screen @s@.
+--
+-- NOTE: @vSort@ operating on virtual names comes with some caveats, see
+-- <https://github.com/xmonad/xmonad-contrib/issues/420 this issue> for
+-- more information. You can use 'marshallSort' like in the following example:
+--
+-- === __Example__
+--
+-- > pp' :: ScreenId -> PP -> PP
+-- > pp' s pp = (marshallPP s pp) { ppSort = fmap (marshallSort s) (ppSort pp) }
+-- >
+-- > mySBL = statusBarProp "xmobar" $ pure (pp' (S 0) pp)
+-- > mySBR = statusBarProp "xmobar" $ pure (pp' (S 1) pp)
+-- > main = xmonad $ withEasySB (mySBL <> mySBR) defToggleStrutsKey def
+--
+-- In this way, you have a custom virtual names sort on top of 'marshallPP'.
+marshallSort :: ScreenId -> ([VirtualWindowSpace] -> [VirtualWindowSpace]) -> ([PhysicalWindowSpace] -> [PhysicalWindowSpace])
 marshallSort s vSort = pScreens . vSort . vScreens where
-    onScreen ws = unmarshallS (tag ws) == s
-    vScreens    = map unmarshallWindowSpace . filter onScreen
+    vScreens    = map unmarshallWindowSpace . workspacesOn s
     pScreens    = map (marshallWindowSpace s)
 
 -- | Convert the tag of the 'WindowSpace' from a 'VirtualWorkspace' to a 'PhysicalWorkspace'.
@@ -183,5 +223,5 @@ marshallWindowSpace   :: ScreenId -> WindowSpace -> WindowSpace
 -- | Convert the tag of the 'WindowSpace' from a 'PhysicalWorkspace' to a 'VirtualWorkspace'.
 unmarshallWindowSpace :: WindowSpace -> WindowSpace
 
-marshallWindowSpace s ws = ws { tag = marshall s  (tag ws) }
-unmarshallWindowSpace ws = ws { tag = unmarshallW (tag ws) }
+marshallWindowSpace s ws = ws { W.tag = marshall s  (W.tag ws) }
+unmarshallWindowSpace ws = ws { W.tag = unmarshallW (W.tag ws) }
--- a/XMonad/Actions/CopyWindow.hs
+++ b/XMonad/Actions/CopyWindow.hs
@@ -18,7 +18,7 @@ module XMonad.Actions.CopyWindow (
                                  -- * Usage
                                  -- $usage
                                  copy, copyToAll, copyWindow, runOrCopy
-                                 , killAllOtherCopies, kill1
+                                 , killAllOtherCopies, kill1, taggedWindows, copiesOfOn
                                  -- * Highlight workspaces containing copies in logHook
                                  -- $logHook
                                  , wsContainingCopies
--- a/XMonad/Config/Dmwit.hs
+++ b/XMonad/Config/Dmwit.hs
@@ -25,7 +25,7 @@ import XMonad.Hooks.DynamicLog
 import XMonad.Hooks.ManageDocks
 import XMonad.Hooks.ManageHelpers
 import XMonad.Layout.Grid
-import XMonad.Layout.IndependentScreens
+import XMonad.Layout.IndependentScreens hiding (withScreen)
 import XMonad.Layout.Magnifier
 import XMonad.Layout.NoBorders
 import XMonad.Prelude


XMonad/Layout/IndependentScreens.hs
```diff
@@ -186,3 +218,86 @@ unmarshallWindowSpace :: WindowSpace -> WindowSpace
...
+            windows $ copy $ wsID !! (j-1)
+
+-- return current and visible workspaces on a specific monitor
+lookupWorkspace' :: (Eq s)=> s -> StackSet i l a s sd-> [i]
+lookupWorkspace' sc w = [ tag j | Screen j s _ <- current w : visible w , s == sc ] 
+
+-- take a list of Strings such as ["1","2","3","4"] -> 1234
+fromDigits :: [Char] -> Int
+fromDigits = foldl addDigit 0
+   where addDigit num d = 10*num + ord d - 48
```
slotThe, line 270: I this guaranteed to always succeed?  If yes, then perhaps `read . head . sequence` would be clearer.  If not, then it's best to return a `Maybe Int` in either case (for the second option you can use `readMaybe`, for your option you'd need to see if `ord d` is actually a digit).

XMonad/Layout/IndependentScreens.hs
```diff
@@ -186,3 +218,86 @@ unmarshallWindowSpace :: WindowSpace -> WindowSpace
...
+marshallSort' s vSort = pScreens . vSort . vScreens where
+        onScreen ws = unmarshallS (tag ws) == s
+        vScreens    = filter onScreen
+        pScreens    = id
+
+-- do something with the Nth workspace
+-- can be used with >9 workspaces
+-- to let withNthWorkspace in XMonad.Actions.DynamicWorkspaces be screen aware
+withNthWorkspace' :: (String -> WindowSet -> WindowSet) -> Int -> X ()
+withNthWorkspace' job wnum = do
```
slotThe, line 245: Perhaps `withNthWorkspaceScreen` or something like that?  In general I think the `f'` idiom is only valid if there's exactly one way in which `f` may be changed, which is not the case here (nor is it for any of the functions below)

XMonad/Layout/IndependentScreens.hs
```diff
@@ -186,3 +218,86 @@ unmarshallWindowSpace :: WindowSpace -> WindowSpace
...
+wsContainingCopies' = do
+    ws <- gets windowset
+    sc <- gets (screen . current . windowset)
+    return $ copiesOfOn (peek ws) (taggedWindows $ workhidden ws sc )
+          where workhidden ws sc = [ Workspace i l a | Workspace i l a <-hidden ws , show' sc == unmarshall' (show i) ]
+
+-- both taggedWIndows and copiesOfOn are not changed, but these two functions are not exported by their original module, so added here for wsContainingCopies'
+-- | Get a list of tuples (tag, [Window]) for each workspace.
+taggedWindows :: [XMonad.StackSet.Workspace i l a] -> [(i, [a])]
+taggedWindows = map $ XMonad.StackSet.tag &&& XMonad.StackSet.integrate' . XMonad.StackSet.stack
```
slotThe, line 284: Perhaps you could import the StackSet qualified via `import qualified XMonad.StackSet as W`; this would be more in-tune with the rest of the code base.

slotThe (changes requested):
Thanks for the contribution!  I noted down a few things that I noticed.  I can't say much in terms of usability of these functions, as I don't use IndependentScreens myself.

The build failure on GHC 8.4.4 seems to be caused by the unecessary import of `Control.Monad`, so that should be fixed quite easily!

XMonad/Layout/IndependentScreens.hs
```diff
@@ -53,15 +68,15 @@ import XMonad.Hooks.DynamicLog
...
+-- > keyBindings conf = let m = modMask conf in fromList $
 -- >     {- lots of other keybindings -}
 -- >     [((m .|. modm, k), windows $ f i)
 -- >         | (i, k) <- zip (XMonad.workspaces conf) [xK_1 .. xK_9]
 -- >         , (f, m) <- [(W.greedyView, 0), (W.shift, shiftMask)]]
 --
 -- This should change to
 --
--- > keyBindings conf = let modm = modMask conf in fromList $
+-- > keyBindings conf = let m = modMask conf in fromList $
```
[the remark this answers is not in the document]
oogeek: Sorry for that, but when I was writing all these, I was using the 0.16 release and had done some testing with it. It is totally my fault that I am careless enough to copy the file straight up to the git version. You can just revert the changes.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -87,6 +102,8 @@ type PhysicalWorkspace = WorkspaceId
 -- inconvenient form (like \"2_web\") that xmonad uses internally. Similarly,
 -- the \"unmarshall\" functions convert in the other direction.
 
+-- S : ScreenId
+
```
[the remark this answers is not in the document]
oogeek: I simply want to add some note for myself and for some newbie who tries to figure out what it is. I find it hard to understand some code without knowing what S does here. Sorry I have messed up my first ever pull request. I should have done more thorough checking beforehand.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -99,13 +116,22 @@ unmarshallS = fst . unmarshall
...
+-- make all workspaces across the monitors bear the same names
 withScreens :: ScreenId            -- ^ The number of screens to make workspaces for
             -> [VirtualWorkspace]  -- ^ The desired virtual workspace names
             -> [PhysicalWorkspace] -- ^ A list of all internal physical workspace names
 withScreens n vws = [marshall sc pws | pws <- vws, sc <- [0..n-1]]
 
+-- workspace names are independently specified with each monitor 
+-- example:
+-- withScreens' 0 myWorkspaces ++ withScreens' 1 myWorkspaces'
+withScreens' :: ScreenId            -- ^ The screen to make workspaces for
```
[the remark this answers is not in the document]
oogeek: Yes you are correct. It should be renamed as withScreen.

XMonad/Layout/IndependentScreens.hs
oogeek: Yes, I should. The original code use import XMonad.StackSet, and I was hesitated to use either import ... as W or just leave the original version.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -186,3 +218,86 @@ unmarshallWindowSpace :: WindowSpace -> WindowSpace
...
+marshallPP' s pp = pp {
+    ppCurrent           = ppCurrent         pp . snd . unmarshall,
+    ppVisible           = ppVisible         pp . snd . unmarshall,
+    ppHidden            = ppHidden          pp ,
+    ppHiddenNoWindows   = ppHiddenNoWindows pp . snd . unmarshall,
+    ppUrgent            = ppUrgent          pp . snd . unmarshall,
+    ppSort              = fmap (marshallSort' s) (ppSort pp)
+    }
+
+-- if use original marshallSort, getSortByIndex will not work properly, since it sends the "video" instead of "0_video" back, and getSortByIndex has no way of knowing the order of "video" if the workspaces are declared with withScreens. However, getSortByTag works.
```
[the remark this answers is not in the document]
oogeek: Thanks for that. I also noticed the bug but I didn't read the issues very clearly. I mainly focused on another pull request which also add some new functionalities to the IndependentScreen.

XMonad/Layout/IndependentScreens.hs
oogeek: My idea is that if someone use getSortByIndex together with dynamicWorkspace (another module), these function will kind of serve as an replacement of the "withNthWorkspace". Your suggestion make sense.

XMonad/Layout/IndependentScreens.hs
slotThe: Yeah the original code was written at a time when this was probably not that common.  I think it's totally fine to change it at this point for more ergonomic use of the StackSet

XMonad/Layout/IndependentScreens.hs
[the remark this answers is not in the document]
slotThe: Ah, that's fair enough.  

You didn't mess up anything; that's the beauty of git!  You can push add another commit that removes these lines (and the `modm` -> `m` above) and I'll take care of squashing everything at the end.

oogeek:
I have tried my best to clean up some of the functions. But when testing, I found out that with ppRename, the order still does not work. So I think include the marshallSort' and I am testing to see if it works. For other things, they work fine.

XMonad/Layout/IndependentScreens.hs
oogeek: So I have deleted it from the latest commit.

XMonad/Layout/IndependentScreens.hs
oogeek: Have renamed it.

slotThe:
>     1. when `marshallSort` in `marshallPP` is replaced with `marshallSort'`, the order is correctly displayed in xmobar. So just change it to `marshallSort'` or remove the original one?

I think in that case this fixes #420; though I'm not quite sure if we can just change how `marshallSort` works without expecting people to complain.  We'll certainly need input from _someone_ who uses IndependentScreens first.

oogeek:
> > ```
> > 1. when `marshallSort` in `marshallPP` is replaced with `marshallSort'`, the order is correctly displayed in xmobar. So just change it to `marshallSort'` or remove the original one?
> > ```
> 
> I think in that case this fixes #420; though I'm not quite sure if we can just change how `marshallSort` works without expecting people to complain. We'll certainly need input from _someone_ who uses IndependentScreens first.

Yeah, it needs more testing. Right now I am using `IndependentScreens` with laptop + external. My config works fine with `marshallPP'` (combined with `marshallSort'`):
    1. since I am converting my config from 0.16 to git version
    2. the only change from the commit I made is I add an old `marshallPP`` from the 0.16 version.

I am adapting my config to the newly introduced `ppRename` and planning for more testing.

slotThe:
On Thu, Apr 01 2021 23:56, oogeek wrote:
> I am adapting my config to the newly introduced `ppRename` and
> planning for more testing. 

Please do; `ppRename' is very useful for the composition of different
modules, so we definitely don't want to lose it.

oogeek:
> Please do; `ppRename' is very useful for the composition of different modules, so we definitely don't want to lose it.

`ppRename` works really nice. I have finished adapting the new config and it works great.

slotThe:
>     1. If we use `ppRename`, then we don't need to use `unmarshallW` in the original `marshallSort`. My tests confirm that `ppRename` in `marshallPP` will handle the rename for us.

Well, it sort of depends.  `marshallSort` is exported and so it's feasible that people use it on its own, without going through `marshallPP`.

>     2. `ppRename` in `marshallPP` called `unmarshallW` for us. We can delete `unmarshallW` from `marshallSort`.

I'm not sure what you mean here; how does `ppRenamet` call `unmarshallW`?

>     3. Using `marshallSort'` will fix the issue you mentioned here #420
> 
>     4. If the change were to be done, some explanation should be added to the document.

After some deliberation I don't think #420 is technically a bug.  `marshallSort` does what it says, it's just perhaps not what some people expect it to do.  We can probably go with another version of it (and probably also `marshallPP` then); though we should think of a better name than `marshallSort'` :)

oogeek:
> I'm not sure what you mean here; how does `ppRenamet` call `unmarshallW`?


I guess it is here:
`marshallPP s pp = pp { ppRename = ppRename pp . unmarshallW`
                       ` , ppSort = fmap (marshallSort s) (ppSort pp) }`

oogeek:
Oh no, `WindowSwitcherDecoration` and `DraggingVisualizer` stopped working with `independentScreens`.  I don't think I have enough time to dig into the source code recently.

slotThe:
> Oh no, `WindowSwitcherDecoration` and `DraggingVisualizer` stopped working with `independentScreens`. I don't think I have enough time to dig into the source code recently.

Is this due to your change, or in general?  If it's the former then we should certainly look at it, in case of the latter I'd say let's concentrate on _this_ pr first and we may be able to tackle this at a later point.

oogeek:
> Is this due to your change, or in general? If it's the former then we should certainly look at it, in case of the latter I'd say let's concentrate on _this_ pr first and we may be able to tackle this at a later point.

I think it is "in general". I have tested a config without any `IndependentScreen.hs`, and it still does not work. After some diff, I find out that the two `WindowSwitcherDecoration` and `DraggingVisualizer` does not change, but the imports is use such as `XMonad.Layout.LayoutModifier` changes. So I think this issue can be dealt later.

slotThe (commented):
>     1. for example I add `wsContainingCopies` to the `IndependentScreens.hs`, after testing I confirm that the original example `copyLogHook` shown in `XMonad.Actions.CopyWindow` document needs modifications. Should I put the example inside the `IndependentScreens.hs` or somewhere else.

I think you should put it in IndependentScreens, they wouldn't make much sense in X.A.CopyWindow directly I think.  Perhaps in the documentation of `wsContainingCopies`?

>     2. Are there any improvements that I can make to polish this pull request, or does it need more testing? I am willing to test by the way and am energetic to make further improvements.

There are a few syntax inconsistencies that I noticed, but I think in the interest of getting this merged without too much fuss I'll just push a qick fix afterwards, so nothing to worry. 
 
>     4. In what direction or for what usage should I test the pull request. I mean I am using it right now and it seems ok, but I am afraid that it needs more testing.

This should perhaps be answered by someone who uses the module themselves :)

oogeek:
I have added some documentation and also tried to expose the two functions in X.A.CopyWindow. 
For the documentation, I have tested it with haddock and made some small changes. It seems ok.

slotThe:
@oogeek I have pushed some (mostly superficial) changes to your branch; try them and tell me what you think!  With that I think this pr looks pretty good—ready to merge as soon as perhaps someone else takes a look.

I admit that I haven't tested these at all; do scold me if I broke something :)

XMonad/Layout/IndependentScreens.hs
```diff
@@ -175,14 +247,39 @@ whenCurrentOn s pp = pp
 -- | If @vSort@ is a function that sorts 'WindowSpace's with virtual names, then @marshallSort s vSort@ is a function which sorts 'WindowSpace's with physical names in an analogous way -- but keeps only the spaces on screen @s@.
 marshallSort :: ScreenId -> ([WindowSpace] -> [WindowSpace]) -> ([WindowSpace] -> [WindowSpace])
 marshallSort s vSort = pScreens . vSort . vScreens where
-    onScreen ws = unmarshallS (tag ws) == s
+    onScreen ws = unmarshallS (W.tag ws) == s
     vScreens    = map unmarshallWindowSpace . filter onScreen
     pScreens    = map (marshallWindowSpace s)
 
+-- | Different from @marshallSort@, @marshallSort'@ does not convert 'WindowSpace'\'s physical names back to virtual names.
```
liskin: This comment seems wrong. Neither function is meant to change names, the only difference is what names the `vSort` functions works with.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -175,14 +247,39 @@ whenCurrentOn s pp = pp
...
+-- >    let copies' = map (drop 2) copies
+-- >    let check ws | ws `elem` copies' = pad . xmobarColor "red" "black" $ ws
+-- >                 | otherwise = pad ws
+-- >    dynamicLogWithPP $ marshallPP s myPP {ppHidden = check, ppOutput = hPutStrLn h}
+-- >
+-- > main = do
+-- >    h <- spawnPipe "xmobar"
+-- >    h1 <- spawnPipe "xmobar"
+-- >    xmonad def { logHook = sampleLogHook h 0 <+> sampleLogHook h1 1 }
+
```
liskin: Drop the empty line.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -137,6 +204,11 @@ marshallPP :: ScreenId -> PP -> PP
 marshallPP s pp = pp { ppRename = ppRename pp . unmarshallW
                      , ppSort = fmap (marshallSort s) (ppSort pp) }
 
+-- | @marshallPP'@ uses @marshallSort'@ instead of marshallSort, the rest is the same with marshallPP.
```
liskin: Would be lovely to have some sort of explanation why one might want that, perhaps an example…

XMonad/Layout/IndependentScreens.hs
```diff
@@ -101,13 +112,27 @@ unmarshallW = snd . unmarshall
 workspaces' :: XConfig l -> [VirtualWorkspace]
 workspaces' = nub . map unmarshallW . workspaces
 
+-- | wsConvertToNum takes a workspace and returns its ScreenId
```
liskin: `unmarshallS . W.tag` takes a workspace and returns its `ScreenId`.

Why do we need another function to do that?

XMonad/Layout/IndependentScreens.hs
```diff
@@ -122,6 +147,48 @@ onCurrentScreen f vws = screen . current >>= f . flip marshall vws
...
+             sort <- getSortByIndex
+             ws <- gets (sort . W.workspaces . windowset)
+             sc <- gets (W.screen . W.current . windowset)
+             let wsID = [ W.tag m | m<-ws, wsConvertToNum m == sc]
+             windows $ copy $ wsID !! (j-1)
+
+-- Take a 'ScreenId' and return a String.
+-- Using @show@ will get "S 0" for screen 0.
+showScreenId :: ScreenId->String
+showScreenId sc = [ last (show sc) ]
```
liskin: This is just wrong, drop this please. If anyone needs the number, `fromIntegral` works well.

liskin (changes requested):
Added a couple comments of my own as well.

Also, the commit message could be clearer: it's not clear what problem you're trying to solve, which part of the code is being changed, etc. This might be a good introduction to writing good commits messages: https://chris.beams.io/posts/git-commit/, and it's usually best to also look at other commits in a specific project you're contributing to to see what practices the devs there follow, so please take a look around here as well.

oogeek:
> Added a couple comments of my own as well.
> 
> Also, the commit message could be clearer: it's not clear what problem you're trying to solve, which part of the code is being changed, etc. This might be a good introduction to writing good commits messages: https://chris.beams.io/posts/git-commit/, and it's usually best to also look at other commits in a specific project you're contributing to to see what practices the devs there follow, so please take a look around here as well.

Thanks for your link. The only problem I send this commit message is that after my commit, I follow the guide and squash it. And I don't know how to deal with squash commit messages. Now I think I can learn more about it. I should be more careful on that.

oogeek:
> @oogeek I have pushed some (mostly superficial) changes to your branch in [this commit](https://github.com/xmonad/xmonad-contrib/pull/501/commits/36affe8bf1968327579132cbb7f40e468a9dd0d6); try them and tell me what you think! With that I think this pr looks pretty good—ready to merge as soon as perhaps someone else takes a look.
> 
> I admit that I haven't tested these at all; do scold me if I broke something :)

It does not break anything. Thanks for your fixup.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -175,14 +242,40 @@ whenCurrentOn s pp = pp
...
+-- > sampleLogHook h s = do
+-- >    copies <- wsContainingCopies
+-- >    let copies' = map (drop 2) copies
+-- >    let check ws | ws `elem` copies' = pad . xmobarColor "red" "black" $ ws
+-- >                 | otherwise = pad ws
+-- >    dynamicLogWithPP $ marshallPP s myPP {ppHidden = check, ppOutput = hPutStrLn h}
+-- >
+-- > main = do
+-- >    h <- spawnPipe "xmobar"
+-- >    h1 <- spawnPipe "xmobar"
```
liskin, line 280: Does this work without any `-x`?

XMonad/Layout/IndependentScreens.hs
```diff
@@ -122,6 +140,51 @@ onCurrentScreen f vws = screen . current >>= f . flip marshall vws
...
+wsContainingCopies = do
+    ws <- gets windowset
+    let sc = W.screen . W.current $ ws
+    return $ copiesOfOn (W.peek ws) (taggedWindows $ workspaceHidden ws sc)
+  where
+    workspaceHidden ws sc =
+        [ W.Workspace i l a
+        | W.Workspace i l a <- W.hidden ws
+        , sc == unmarshallS i
+        ]
```
liskin, line 170:
```suggestion
    workspaceHidden ws sc = [ w | w <- W.hidden ws, sc == unmarshallS (W.tag w) ]
```

XMonad/Layout/IndependentScreens.hs
```diff
@@ -175,14 +242,40 @@ whenCurrentOn s pp = pp
...
 -- | Convert the tag of the 'WindowSpace' from a 'PhysicalWorkspace' to a 'VirtualWorkspace'.
 unmarshallWindowSpace :: WindowSpace -> WindowSpace
 
-marshallWindowSpace s ws = ws { tag = marshall s  (tag ws) }
-unmarshallWindowSpace ws = ws { tag = unmarshallW (tag ws) }
+marshallWindowSpace s ws = ws { W.tag = marshall s  (W.tag ws) }
+unmarshallWindowSpace ws = ws { W.tag = unmarshallW (W.tag ws) }
+
+
+-- $logHook
```
liskin, line 265: Rename this section to `"XMonad.Actions.CopyWindow" integration` and move the relevant new functions there, please.

oogeek: I have done so.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -204,10 +211,9 @@ marshallPP :: ScreenId -> PP -> PP
 marshallPP s pp = pp { ppRename = ppRename pp . unmarshallW
                      , ppSort = fmap (marshallSort s) (ppSort pp) }
 
--- | @marshallPP'@ uses @marshallSort'@ instead of marshallSort, the rest is the same with marshallPP.
+-- | Like 'marshallPP', but uses 'marshallSort'' instead of 'marshallSort'.
 marshallPP' :: ScreenId -> PP -> PP
-marshallPP' s pp = pp { ppRename = ppRename pp . unmarshallW
-                     , ppSort = fmap (marshallSort' s) (ppSort pp) }
+marshallPP' s pp = (marshallPP s pp){ ppSort = fmap (marshallSort' s) (ppSort pp) }
```
dmwit, line 203: I like this change, good idea.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -276,6 +274,14 @@ unmarshallWindowSpace ws = ws { W.tag = unmarshallW (W.tag ws) }
 -- >    dynamicLogWithPP $ marshallPP s myPP {ppHidden = check, ppOutput = hPutStrLn h}
 -- >
 -- > main = do
--- >    h <- spawnPipe "xmobar"
--- >    h1 <- spawnPipe "xmobar"
+-- >    h <- spawnPipe "xmobar -x 0 <path to config>"
+-- >    h1 <- spawnPipe "xmobar -x 1 <path to config>"
 -- >    xmonad def { logHook = sampleLogHook h 0 <+> sampleLogHook h1 1 }
+
```
dmwit, line 280: Perhaps we should show the generalized form that works for an arbitrary number of screens, rather than its specialization to two screens. Something like this (untested):

    main = do
        n <- countScreens
        hs <- traverse (\(S n) -> spawnPipe ("xmobar -x " ++ show n ++ " ~/.xmobarrc-" ++ show n)) [0..n]
        xmonad def { logHook = mconcat (zipWith sampleLogHook hs [0..n]) }

dmwit (changes requested):

XMonad/Layout/IndependentScreens.hs
```diff
@@ -101,13 +112,27 @@ unmarshallW = snd . unmarshall
...
+-- | Make all workspaces across the monitors bear the same names
 withScreens :: ScreenId            -- ^ The number of screens to make workspaces for
             -> [VirtualWorkspace]  -- ^ The desired virtual workspace names
             -> [PhysicalWorkspace] -- ^ A list of all internal physical workspace names
-withScreens n vws = [marshall sc pws | pws <- vws, sc <- [0..n-1]]
+withScreens n vws = concatMap (`withScreen` vws) [0..n-1]
 
 onCurrentScreen :: (VirtualWorkspace -> WindowSet -> a) -> (PhysicalWorkspace -> WindowSet -> a)
-onCurrentScreen f vws = screen . current >>= f . flip marshall vws
+onCurrentScreen f vws = W.screen . W.current >>= f . flip marshall vws
```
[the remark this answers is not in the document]
dmwit: Seems like this comment got attached to the wrong line by accident, whoops!

XMonad/Layout/IndependentScreens.hs
dmwit: Okay, rereading, probably using `i` instead of `n` in the lambda is better, and both `[0..n]`'s should be `[0..n-1]`'s.

oogeek: Thanks for your suggestion. Have added it to the document and tested it.

oogeek:
Changes right now:
1. Rewritten `marshallSort` and `marshallSort'`
2. Add `workspacesOn`
3. fix explanation of `marshallSort'`
4. Add two new types: `PhysicalWindowSpace` `VirtualWindowSpace`

slotThe:
It may also be worth to check if there is some overlap with #438

dmwit:
> We may want to find some solutions that are general enough to convert functions from other modules to be screen-aware, but it might be impossible.

Yes, it's definitely the case that we can't hope to support *all* functions users might want to marshall/unmarshall. If we have a few common cases (seems we do now?) I think that's fine. I think one reasonable rule of thumb for "did we make it general enough yet" is "did we make it so we don't have to import the other module yet".

This is great so far, you're great, keep being great.

oogeek:
> It may also be worth to check if there is some overlap with #438

I have checked it and fairly speaking, I think there are some differences between our commits. 
Difference:
1. The author from https://github.com/xmonad/xmonad-contrib/pull/438 does not consider hidden workspaces, forcing the user to write his/her/their own code to deal with it. On the contrary, I consider that in my code. 
2. If `IndependentScreens` is used together with `WithNthWorkspaces` (function) and `DynamicWorkspaces` (module), then the 
```
-- | Get the nth virtual workspace
nthWorkspace :: Int -> X (Maybe VirtualWorkspace)
nthWorkspace n = (!!? n) . workspaces' <$> asks config
```
is not enough to deal with it. Dynamically added workspaces are not considered here. However, I consider it in my code, with similar reason as `Difference 1`.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -101,13 +112,27 @@ unmarshallW = snd . unmarshall
...
 workspaces' = nub . map unmarshallW . workspaces
 
+-- | wsConvertToNum takes a workspace and returns its ScreenId
+wsConvertToNum :: (Num b)=>W.Workspace [Char] l a-> b
+wsConvertToNum j = fromIntegral (toInteger $ digitToInt $ head $ takeWhile (/= '_') (W.tag j))
+
+-- | Workspace names are independently specified with each monitor.
+-- You can define your workspaces by calling withScreen for each screen:
+--
+-- > myConfig = def { workspaces = withScreen 0 ["web", "email", "irc"]  ++ withScreen 1 ["1", "2", "3"] }
```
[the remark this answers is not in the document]
oogeek: Can you share some idea as to how to implement it in a better way? My goal is to have it work like independent workspaces in Dwm.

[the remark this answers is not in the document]
dmwit: It's been a while since I used dwm. But if my memory serves, what you're describing in this documentation isn't anything like what dwm does. What dwm does is it has a fixed set of workspace names, and each screen gets their own independent copy of that name. The collection doesn't vary from screen to screen. That is what `IndependentScreens` is trying to duplicate.

If all the workspace names are different on each screen, then there's almost nothing to implement. You just have `workspaces = ["web", "email", "irc", "1", "2", "3"]` and then write your workspace-switching keybindings to only put `web` on the left and `1` on the right. You don't need any complicated marshalling and unmarshalling in sorts and displays and other mechanisms, because the fact that `web` only appears on the left is completely invisible to every other part of xmonad than the keybindings.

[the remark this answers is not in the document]
oogeek: > The collection doesn't vary from screen to screen. That is what `IndependentScreens` is trying to duplicate.
I see. So I think I will delete that line of documentation.

oogeek:
So it seems that I have the wrong notion that independentScreens can have variable number of workspaces on each screen. But it is fixed. So I think I will make some changes and simply some functions.

TheMC47:
Please do not merge master into your branch, rebase it instead. More details: https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md

oogeek:
> Please do not merge master into your branch, rebase it instead. More details: https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md

I think I screw up the rebase process. I will try my best to fix it.

oogeek:
I need some suggestions.
@slotThe @liskin 
I think that somehow my PR has diverged from the original aim of `X.L.IndependentScreens`, as pointed out by @dmwit. 
So shall I make a new layout called `X.L.MultiScreens` and put some functions there as example of Integration?
Meanwhile I can just work on cleaning up some type and functions in `X.L.IndependentScreens` and finish my PR.

liskin:
> I need some suggestions.
> @slotThe @liskin
> I think that somehow my PR has diverged from the original aim of `X.L.IndependentScreens`, as pointed out by @dmwit.
> So shall I make a new layout called `X.L.MultiScreens` and put some functions there as example of Integration? Like `wsContaingCopies` and `withNthWorkspaceScreens`
> Meanwhile I can just work on cleaning up some type and functions in `X.L.IndependentScreens` and finish my PR.

I don't think a separate module is the way to go here, it's not good for discoverability. As an example, consider a new user trying to use IndependentScreens with CopyWindow. They wouldn't go looking for hints about integration to a third module called MultiScreens, would they?

On the other hand, separating the integration stuff from this PR might be a good idea. The changes around `marshallSort` and `marshallPP` make sense, so a separate commit with those is definitely a way to get that part merged sooner. As noted earlier, this stuff is related to #420, so a commit that clearly references this via a `Closes: https://github.com/xmonad/xmonad-contrib/issues/420` or `Relates: https://github.com/xmonad/xmonad-contrib/issues/420` trailer would be nice. But I'm not really sure whether it's clear yet what the correct solution is? My own feeling (as someone who's never used dwm or IndependentScreens) is that `marshallPP` and `marshallSort` should just work with the default `ppSort` that uses `getSortByIndex`, so what you have as `marshallSort'` should in fact be the default, and this would become a fix that is also a breaking change in behaviour. But I have no idea how other people use IndependentScreens and if this is going to break their configs or not. I guess we better discuss this in #420?

oogeek:
Thank you for your suggestion. @liskin So I think I will first put the `Integration part` to another file. Then I finish this PR. After that, I will start a separate PR and commit the integration. 

Yeah, let's discuss in https://github.com/xmonad/xmonad-contrib/issues/420

oogeek:
I have deleted the integration and as @liskin has said, they can be delayed for the next PR. Right now I focus on two things, one is to fix the bug in https://github.com/xmonad/xmonad-contrib/issues/420. Another is the function I added `currentWorkspaces`, which can be used by later `Integration part`.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -134,8 +160,7 @@ countScreens = fmap genericLength . liftIO $ openDisplay "" >>= liftA2 (<*) getS
 -- > logHook = let log screen handle = dynamicLogWithPP . marshallPP screen . pp $ handle
 -- >           in log 0 hLeft >> log 1 hRight
 marshallPP :: ScreenId -> PP -> PP
-marshallPP s pp = pp { ppRename = ppRename pp . unmarshallW
-                     , ppSort = fmap (marshallSort s) (ppSort pp) }
+marshallPP s pp = pp { ppRename = ppRename pp . unmarshallW }
```
liskin, line 163: We still need something like `ppSort = ppSort pp . workspacesOn s` here.

oogeek: Thanks for pointing out, I have added it now.

liskin: Now the docs are out of date again. And the marshallPP doc needs to be adjusted, too, as it's no longer accurate.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -29,7 +29,7 @@ module XMonad.Layout.IndependentScreens (
     -- * Converting between virtual and physical workspaces
     -- $converting
     marshall, unmarshall, unmarshallS, unmarshallW,
-    marshallWindowSpace, unmarshallWindowSpace, marshallSort,
+    marshallWindowSpace, unmarshallWindowSpace, marshallSort
```
liskin: It's better to have the trailing comma there, actually, it makes future git diffs smaller.

XMonad/Layout/IndependentScreens.hs
oogeek: I have updated the docs. Does it look OK now?

XMonad/Layout/IndependentScreens.hs
```diff
@@ -101,13 +110,30 @@ unmarshallW = snd . unmarshall
...
-withScreens n vws = [marshall sc pws | pws <- vws, sc <- [0..n-1]]
+withScreens n vws = concatMap (`withScreen` vws) [0..n-1]
 
 onCurrentScreen :: (VirtualWorkspace -> WindowSet -> a) -> (PhysicalWorkspace -> WindowSet -> a)
-onCurrentScreen f vws = screen . current >>= f . flip marshall vws
+onCurrentScreen f vws = W.screen . W.current >>= f . flip marshall vws
+
+-- | 'currentWorkspaces' takes a 'PhysicalWindowSpace' sorting function, the workspaces on 
+-- current screen and returns the sorted results. Can be used by other integration functions later.
+currentWorkspaces :: ([PhysicalWindowSpace] -> [PhysicalWindowSpace]) -> X [PhysicalWorkspace]
```
liskin, line 130: What's the intended use of this function? Is it still useful or was that part of some integration stuff that was since removed from this PR?

XMonad/Layout/IndependentScreens.hs
```diff
@@ -122,20 +148,17 @@ onCurrentScreen f vws = screen . current >>= f . flip marshall vws
...
--- | This turns a naive pretty-printer into one that is aware of the
--- independent screens. That is, you can write your pretty printer to behave
--- the way you want on virtual workspaces; this function will convert that
--- pretty-printer into one that first filters out physical workspaces on other
--- screens, then converts all the physical workspaces on this screen to their
--- virtual names.
+-- | This turns a pretty-printer into one that is aware of the
+-- independent screens. The converted pretty-printer will first 
+-- filters out physical workspaces on other screens, then sort 
+-- all the physical workspaces on this screen.
```
liskin, line 154: We shouldn't drop the mention of virtual names, I think. They're not involved in sorting any more, but marshallPP still converts the names in ppRename.

How about this:

> This turns a pretty-printer into one that is aware of the independent screens. The converted pretty-printer first filters out physical workspaces on other screens, then converts all the physical workspaces on this screen to their virtual names.
> Note that 'ppSort' still operates on physical (marshalled) workspace names, otherwise functions from "XMonad.Util.WorkspaceCompare" wouldn't work. If you need to sort on virtual names, see 'marshallSort'.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -172,17 +195,33 @@ whenCurrentOn s pp = pp
...
-    vScreens    = map unmarshallWindowSpace . filter onScreen
+    vScreens    = map unmarshallWindowSpace . workspacesOn s
     pScreens    = map (marshallWindowSpace s)
+-- This marshallSort has a bug: see https://github.com/xmonad/xmonad-contrib/issues/420
+-- To use 'getSortByIndex', please use the better bug-fixed solution provided below. 
+
+-- This marshallSort is the better solution.
+-- It sorts directly the physical workspace names on the current screen.
+-- > marshallSort :: ScreenId -> ([PhysicalWindowSpace] -> [PhysicalWindowSpace]) -> ([PhysicalWindowSpace] -> [PhysicalWindowSpace])
+-- > marshallSort s vSort = vSort . workspacesOn s
```
liskin, line 215: I think we don't need to mention this as it's the default in `marshallPP`. Just add an example how to use `marshallSort` to provide a custom virtual names sort on top of `marshallPP`, and put that example in the haddock comment so that it's visible in the generated documentation.

XMonad/Layout/IndependentScreens.hs
oogeek: Thanks very much. I think your version is clearer than mime. I forgot to mention about 'ppSort' here.

XMonad/Layout/IndependentScreens.hs
oogeek: I have updated with an example, not sure if it is appropriate.

XMonad/Layout/IndependentScreens.hs
oogeek: Yeah, I intend to use it for later integration stuff. Right now the integration stuffs are removed from this PR.

liskin: In that case it would be best to drop this from the PR as well, it seems out of place in a "fix #420" PR.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -171,17 +191,36 @@ whenCurrentOn s pp = pp
...
+-- <https://github.com/xmonad/xmonad-contrib/issues/420 this issue>.
+-- Only use it if you want that behaviour!
+marshallSort :: ScreenId -> ([VirtualWindowSpace] -> [VirtualWindowSpace]) -> ([PhysicalWindowSpace] -> [PhysicalWindowSpace])
 marshallSort s vSort = pScreens . vSort . vScreens where
-    onScreen ws = unmarshallS (tag ws) == s
-    vScreens    = map unmarshallWindowSpace . filter onScreen
+    vScreens    = map unmarshallWindowSpace . workspacesOn s
     pScreens    = map (marshallWindowSpace s)
 
+-- | To still sort by 'VirtualWorkspaces', you can use 'marshallSort' like this:
```
liskin, line 210: This comment still won't be rendered by haddock.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -171,17 +191,36 @@ whenCurrentOn s pp = pp
...
-marshallSort :: ScreenId -> ([WindowSpace] -> [WindowSpace]) -> ([WindowSpace] -> [WindowSpace])
+-- | Filter workspaces that are on current screen.
+workspacesOn :: ScreenId -> [PhysicalWindowSpace] -> [PhysicalWindowSpace]
+workspacesOn s = filter (\ws -> unmarshallS (W.tag ws) == s)
+
+-- | @vSort@ is a function that sorts 'VirtualWindowSpace's with virtual names.
+-- @marshallSort s vSort@ is a function which sorts 'PhysicalWindowSpace's with virtual names,
+-- but keeps only the 'WindowSpace'\'s on screen @s@.
+--
+-- NOTE: This function always sorts the workspace names lexically, see
```
liskin, line 202: This isn't true, it uses the provided `vSort` for sorting.

XMonad/Layout/IndependentScreens.hs
slotThe: Oh, must have missed that.  I'll move it over to the `marshallSort` docs

CHANGES.md
```diff
@@ -113,6 +113,18 @@
...
+
+    - Added `workspacesOn` for filtering workspaces on the current screen.
+
+    - Added `withScreen` to specify names for a given single screen.
+
+    - Added new aliases `PhysicalWindowSpace` and `VirtualWindowSpace`
+      for a `WindowSpace` for easier to read function signatures.
+
+    - Fixed a bug where `marshallPP` always sorted workspace names
+      lexically.
```
liskin, line 126: This should be more prominent I think because while it does fix a bug with the Default ppSort, it also changes behaviour of marshallPP with non-default ppSorts.

CHANGES.md
```diff
@@ -621,6 +609,22 @@
...
       other applications, like compositors, to properly match on it.
 
+  * `XMonad.Layout.IndependentScreens`
+
+    - Fixed a bug where `marshallPP` always sorted workspace names
+      lexically.  This changes the default behaviour of `marshallPP`—the
+      given `ppSort` now operates in the _physical_ workspace names.
+      The documentation of `marshallSort` contains an example of how to
+      get the old behaviour, where `ppSort` operates in virtual names,
+      back.
```
liskin, line 619: Nice! :+1:

liskin (approved):

PR #502, opened 2021-04-03
`X.H.StatusBar`: Removed unnecessary IO

TheMC47:
### Description

This is a follow-up on the IRC [discussion](https://ircbrowse.tomsmeding.com/browse/xmonad?id=28392#trid28392)

This applies for `withSB` and `withEasySB`, as well as `statusBarProp` and
`statusBarPropTo`, making composability better. `statusBarPipe` is more
awkward to use now, but that's fine

### Checklist

### DIFF
--- a/XMonad/Hooks/DynamicLog.hs
+++ b/XMonad/Hooks/DynamicLog.hs
@@ -198,7 +198,7 @@ statusBar :: LayoutClass l Window
           -> IO (XConfig (ModifiedLayout AvoidStruts l))
 statusBar cmd pp k conf= do
   sb <- statusBarPipe cmd (pure pp)
-  withEasySB sb k conf
+  return $ withEasySB sb k conf
 
 -- |
 -- Helper function which provides ToggleStruts keybinding
@@ -260,7 +260,6 @@ dynamicLogXinerama = withWindowSet $ io . putStrLn . pprWindowSetXinerama
 -- the menu bar.
 xmobarProp :: LayoutClass l Window
            => XConfig l  -- ^ The base config
-           -> IO (XConfig (ModifiedLayout AvoidStruts l))
-xmobarProp conf = do
-    xmobarPropConfig <- statusBarProp "xmobar" (pure xmobarPP)
-    withEasySB xmobarPropConfig toggleStrutsKey conf
+           -> XConfig (ModifiedLayout AvoidStruts l)
+xmobarProp =
+  withEasySB (statusBarProp "xmobar" (pure xmobarPP)) toggleStrutsKey
--- a/XMonad/Hooks/StatusBar.hs
+++ b/XMonad/Hooks/StatusBar.hs
@@ -82,9 +82,8 @@ import XMonad.Hooks.StatusBar.PP
 -- status bar that supports property logging, is to use 'statusBarProp'
 -- with 'withEasySB'; these take care of the necessary plumbing:
 --
--- > main = do
--- >   mySB <- statusBarProp "xmobar" (pure xmobarPP)
--- >   xmonad =<< withEasySB mySB defToggleStrutsKey def
+-- > mySB = statusBarProp "xmobar" (pure xmobarPP)
+-- > main = xmonad $ withEasySB mySB defToggleStrutsKey def
 --
 -- You can read more about X11 properties
 -- [here](https://en.wikipedia.org/wiki/X_Window_System_core_protocol#Properties)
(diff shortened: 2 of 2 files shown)


XMonad/Hooks/DynamicLog.hs
```diff
@@ -261,6 +261,5 @@ dynamicLogXinerama = withWindowSet $ io . putStrLn . pprWindowSetXinerama
 xmobarProp :: LayoutClass l Window
            => XConfig l  -- ^ The base config
            -> IO (XConfig (ModifiedLayout AvoidStruts l))
-xmobarProp conf = do
-    xmobarPropConfig <- statusBarProp "xmobar" (pure xmobarPP)
-    withEasySB xmobarPropConfig toggleStrutsKey conf
+xmobarProp conf = return
```
slotThe, line 264: I wonder if dropping this would be a good idea as well, so it behaves more like `withEasySB`

slotThe (commented):
Works for me :+1:   Just one small thing I noticed

XMonad/Hooks/StatusBar.hs
```diff
@@ -314,11 +316,12 @@ statusBarPipe cmd xpp  = do
 -- Here's an example of what such declarative configuration of multiple status
 -- bars may look like:
 --
--- > main = do
--- >   xmobarTop    <- statusBarPipe "xmobar -x 0 ~/.config/xmobar/xmobarrc_top"    (pure ppTop)
--- >   xmobarBottom <- statusBarPipe "xmobar -x 0 ~/.config/xmobar/xmobarrc_bottom" (pure ppBottom)
--- >   xmobar1      <- statusBarPipe "xmobar -x 1 ~/.config/xmobar/xmobarrc1"       (pure pp1)
--- >   xmonad =<< withSB (xmobarTop <> xmobarBottom <> xmobar1) myConfig
+-- > -- Make sure to setup the xmobar config accordingly
```
liskin, line 319:
```suggestion
-- > -- Make sure to setup the xmobar configs accordingly
```

XMonad/Hooks/StatusBar.hs
```diff
@@ -314,11 +316,12 @@ statusBarPipe cmd xpp  = do
...
+-- > xmobarTop = statusBarPropTo "_XMONAD_LOG_1" "xmobar -x 0 ~/.config/xmobar/xmobarrc_top" (pure ppTop)
+-- > xmobarBottom = statusBarPropTo "_XMONAD_LOG_2" "xmobar -x 0 ~/.config/xmobar/xmobarrc_bottom" (pure ppBottom)
+-- > xmobar1 = statusBarPropTo "_XMONAD_LOG_3" "xmobar -x 1 ~/.config/xmobar/xmobarrc1" (pure pp1)
```
liskin, line 322: The horizontal alignment was nice. It's not git/diff-friendly but in the docs it's worth it I think.

XMonad/Hooks/StatusBar.hs
```diff
@@ -132,6 +130,12 @@ import XMonad.Hooks.StatusBar.PP
 -- bar read from that pipe.
 -- Please be aware that this kind of setup is very bug-prone and hence is
 -- discouraged: if anything goes wrong with the bar, xmonad will freeze!
+-- Also note that 'statusBarPipe' returns 'IO StatusBarConfig', so
```
liskin, line 133: I'd add an empty line above this to make it more visible.

XMonad/Hooks/DynamicLog.hs
liskin: `xmobarProp` was merged in late December and we're hoping people won't use DynamicLog any more, so it doesn't matter much. We might break a config or two, but then this might be considered a good thing as it will notify people who follow the bleeding edge stuff that there is a bleedinger edger stuff in X.H.StatusBar. :-)

(I'm not really sure why we decided to keep this one addition in DynamicLog, so if anyone knows, they're in a better position to decide whether to change this type or not.)

liskin (approved):
Just minor nits.

Overall this makes the interface even better. Thanks!

XMonad/Hooks/DynamicLog.hs
TheMC47: Hmm, good points. Should we move it then? And maybe re-export it?

liskin: I meant "keep" as in "not remove because it wasn't there at 0.16". It definitely belongs in DynamicLog, it's just a question whether it should exist at all.

TheMC47: Oh, I think it might nudge people to checkout `X.H.StatusBar`, so maybe it stays?

slotThe: I think it's nice having it there as a sort of "look what `X.H.StatusBar` can do".  Perhaps this is wishful thinking, but I think it would drive people exploring the docs towards at least checking out these modules

PR #505, opened 2021-04-06
X.Prompt: Special handling of one highlighted item

slotThe:
### Description

Fixes a bug introduced in f2cfaa33980061f4fb39b689403701d36babe33f.

The changes there allowed the prompt to cycle through its input; it now
becomes necessary to single out the case of only having a single
suggestion that's also highlighted.  Otherwise, `hlComplete` (condition:
`alwaysHighlight` is enabled) will loop forever.  This case can't be
handled from within hlComplete, as we are giving it incomplete
information to start with (by only calling it with the last word of the
current completion in `handleCompletion`), which is necessary for things
like completing arguments.

Sigh.

### Checklist

  I don't think this is really necessary.

### DIFF
--- a/XMonad/Prompt.hs
+++ b/XMonad/Prompt.hs
@@ -410,9 +410,10 @@ highlightedItem st' completions = case complWinDim st' of
       (_,_,_,_,xx,yy) = winDim
       complMatrix = splitInSubListsAt (length yy) (take (length xx * length yy) completions)
       (col_index,row_index) = complIndex st'
-    in case completions of
-      [] -> Nothing
-      _  -> complMatrix !? col_index >>= (!? row_index)
+    in case length completions of
+      0 -> Nothing
+      1 -> Just $ complMatrix !! col_index !! row_index
+      _ -> complMatrix !? col_index >>= (!? row_index)
  where
   -- | Safe version of '(!!)'.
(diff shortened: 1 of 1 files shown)


XMonad/Prompt.hs
```diff
@@ -410,9 +410,10 @@ highlightedItem st' completions = case complWinDim st' of
       (_,_,_,_,xx,yy) = winDim
       complMatrix = splitInSubListsAt (length yy) (take (length xx * length yy) completions)
       (col_index,row_index) = complIndex st'
-    in case completions of
-      [] -> Nothing
-      _  -> complMatrix !? col_index >>= (!? row_index)
+    in case length completions of
+      0 -> Nothing
+      1 -> Just $ complMatrix !! col_index !! row_index
```
liskin, line 415: This seems really weird. Whenever `complMatrix !? col_index >>= (!? row_index)` returns `Just`, this behaves the same, and when it would return `Nothing`, this one is a bottom. Or am I missing something?

slotThe: Fixed in 344ace492b5e03769ccea334be78661b411c465d

liskin: :-)

I'll just trust on this one then, I know very little about the Prompt internals and nothing really weird stands out any more.

PR #514, opened 2021-04-12
Add 'proportional' to the export list of X.A.EasyMotion

exorcist365:
### Description

behind them.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -232,9 +232,6 @@
 
 ### Bug Fixes and Minor Changes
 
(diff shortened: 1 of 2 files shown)


slotThe:
Oh good catch, thanks!  I don't think we really need the `CHANGES.md` change; it's a new module and so these kinds of changes are to be expected until we make a new release.

exorcist365:
> Oh good catch, thanks! I don't think we really need the `CHANGES.md` change; it's a new module and so these kinds of changes are to be expected until we make a new release.

All right.

Issue #518, opened 2021-04-14
X.H.WallpaperSetter with PNG support?


oogeek:
### Problem Description
I am wondering why as stated in `X.H.WallpaperSetter`, the background image should be JPG format.
I have read the source code and can not figure it out. 

Specifically, I don't understand why jpg is used here.
```
-- |returns the default association list (maps name to name.jpg, non-alphanumeric characters are omitted)
defWPNames :: [WorkspaceId] -> WallpaperList
defWPNames xs = WallpaperList $ map (\x -> (x,WallpaperFix (filter isAlphaNum x++".jpg"))) xs
```
We can actually take the last few characters and use it as suffix. Then we can have `abc.png` or `abc.jpg`

Another question is, is it necessary to `convert` the image now:
```
-- | Gets a list of geometry rectangles and filenames, builds and sets wallpaper
applyWallpaper :: [(Rectangle, FilePath)] -> X ProcessHandle
applyWallpaper parts = do
  winset <- gets windowset
  let (vx,vy) = getVScreenDim winset
  layers <- liftIO $ mapM layerCommand parts
  let basepart ="convert -size "++show vx++"x"++show vy++" xc:black "
      endpart =" jpg:- | feh --no-xinerama --bg-tile --no-fehbg -"
      cmd = basepart ++ (concat $ intersperse " " layers) ++ endpart
  liftIO $ runCommand cmd

```
I believe that `feh` can handle that if other options are used. And it is unnecessary if the background picture is just the same size as the screen.
### Checklist

  - [X] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

TheMC47:
I think you're making valid points, but there's always the concern of staying backward-compatible (not that I think a lot of people are using the module).
I say we break compatibility because it feels like a net improvement.

slotThe:
On Fri, Apr 16 2021 16:42, Yecine Megdiche wrote:
> I think you're making valid points, but there's always the concern of
> staying backward-compatible (not that I think a lot of people are
> using the module).
> I say we break compatibility because it feels like a net improvement. 

I feel like the .jpg -> allow more img formats direction should be able
to stay backwards compatible.  

I admit that I haven't looked over the code very much, but I'm not
certain that the second point holds and feh can do everything that
[...]

PR #521, opened 2021-04-17
add support for multiple formats for X.H.Wallpapersetter

oogeek:
### Description
As discussed in https://github.com/xmonad/xmonad-contrib/issues/518
I just clean a bit of the syntax and have tested the module. It is painfully slow with noticeable lagging. As I don't know how to implement a cache for the images, I can not improve the performance. The `todo` by previous author(s) also mentioned about caching.

Use `feh` without `convert` will see an small boost in performance. But that is up to the users to decide. 

### Checklist

### DIFF
--- a/XMonad/Hooks/WallpaperSetter.hs
+++ b/XMonad/Hooks/WallpaperSetter.hs
@@ -19,7 +19,7 @@ module XMonad.Hooks.WallpaperSetter (
 , Wallpaper(..)
 , WallpaperList(..)
 , defWallpaperConf
-, defWPNames
+, defWPNamesJpg, defWPNamesPng, defWPNames
   -- *TODO
   -- $todo
 ) where
@@ -34,7 +34,7 @@ import System.FilePath ((</>))
 import System.Random (randomRIO)
 
 import qualified Data.Map as M
-import Data.List (intersperse, sortBy)
+import Data.List (sortBy)
 import Data.Char (isAlphaNum)
 import Data.Ord (comparing)
 
@@ -82,7 +82,7 @@ newtype WallpaperList = WallpaperList [(WorkspaceId, Wallpaper)]
 instance Monoid WallpaperList where
   mempty = WallpaperList []
   mappend (WallpaperList w1) (WallpaperList w2) =
-    WallpaperList $ M.toList $ (M.fromList w2) `M.union` (M.fromList w1)
+    WallpaperList $ M.toList $ M.fromList w2 `M.union` M.fromList w1
 
 instance Semigroup WallpaperList where
   (<>) = mappend
@@ -100,9 +100,17 @@ defWallpaperConf = WallpaperConf "" $ WallpaperList []
 instance Default WallpaperConf where
     def = defWallpaperConf
 
--- |returns the default association list (maps name to name.jpg, non-alphanumeric characters are omitted)
+{-# DEPRECATED defWPNames "Use defWPNamesJpg instead" #-}
 defWPNames :: [WorkspaceId] -> WallpaperList
-defWPNames xs = WallpaperList $ map (\x -> (x,WallpaperFix (filter isAlphaNum x++".jpg"))) xs
+defWPNames = defWPNamesJpg
+
+-- | Return the default association list (maps @name@ to @name.jpg@, non-alphanumeric characters are omitted)
+defWPNamesJpg :: [WorkspaceId] -> WallpaperList
+defWPNamesJpg xs = WallpaperList $ map (\x -> (x, WallpaperFix (filter isAlphaNum x ++ ".jpg"))) xs
+
+-- | Like 'defWPNamesJpg', but map @name@ to @name.png@ instead.
+defWPNamesPng :: [WorkspaceId] -> WallpaperList
+defWPNamesPng xs = WallpaperList $ map (\x -> (x, WallpaperFix (filter isAlphaNum x ++ ".png"))) xs
 
 -- | Add this to your log hook with the workspace configuration as argument.
 wallpaperSetter :: WallpaperConf -> X ()
@@ -158,7 +166,6 @@ getPicRes picpath = do
     [[(w,"")],[(h,"")]] -> Just (w,h)
     _ -> Nothing
 
-
 -- |complete unset fields to default values (wallpaper directory = ~/.wallpapers,
 --  expects a file "NAME.jpg" for each workspace named NAME)
 completeWPConf :: WallpaperConf -> X WallpaperConf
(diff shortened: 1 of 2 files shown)


XMonad/Hooks/WallpaperSetter.hs
```diff
@@ -102,7 +102,9 @@ instance Default WallpaperConf where
 
 -- |returns the default association list (maps name to name.jpg, non-alphanumeric characters are omitted)
 defWPNames :: [WorkspaceId] -> WallpaperList
-defWPNames xs = WallpaperList $ map (\x -> (x,WallpaperFix (filter isAlphaNum x++".jpg"))) xs
+defWPNames xs = WallpaperList $ map (\x -> (x, WallpaperFix (filter isAlphaNum x ++ "." ++ getFormat x))) xs
+    where getFormat :: [Char] -> [Char]
+          getFormat x = reverse . takeWhile isAlphaNum $ reverse x
```
slotThe, line 107: I think using something like `System.FilePath.Posix.takeExtension` here would be better

XMonad/Hooks/WallpaperSetter.hs
```diff
@@ -198,18 +200,18 @@ applyWallpaper parts = do
   layers <- liftIO $ mapM layerCommand parts
   let basepart ="convert -size "++show vx++"x"++show vy++" xc:black "
       endpart =" jpg:- | feh --no-xinerama --bg-tile --no-fehbg -"
```
slotThe, line 202: There's still this in here; will this interfere with using non-jpg image formats?

XMonad/Hooks/WallpaperSetter.hs
TheMC47: Now that I looked at it more, I feel like we should keep the `.jpg` here. The function is useful if you want to get the wallpaper directly from the workspace name. In this version, you'd need to write the extension in your workspace name, which isn't a good UX :frowning_face: So having `.jpg` is a sensible default IMO.

One way to allow for other types, is to create another function that takes the extension as an extra parameter to the workspace list, and maybe also to the `WallpaperConf` (as `defWPNames` is called by `completeWPConf` in `wallpaperSetter`), but I'm not sure if it's worth it (I'll probably not use the module)

XMonad/Hooks/WallpaperSetter.hs
oogeek: I am testing for this now.

oogeek:
Here is what I got from man page `convert`

```
By  default,  the image format of `file' is de‐
termined by its magic  number.   To  specify  a
particular  image  format, precede the filename
with an image format name  and  a  colon  (i.e.
ps:image)  or  specify  the  image  type as the
filename  suffix  (i.e.   image.ps).    Specify
'file' as '-' for standard input or output.
```

XMonad/Hooks/WallpaperSetter.hs
slotThe: Good point.  I think this is something that the documentation doesn't really explain fully; might be worth updating that part as well

XMonad/Hooks/WallpaperSetter.hs
oogeek: Sadly, I found out that this can not be altered. It can only be set to one among `png` `jpg` .... 
The operation is basically this
1. if there are multiple monitors, convert the two pictures into one big picture, then use `feh` to display it. For example `1920x1080` + `2560x1440` -> `4480x1440`

oogeek: convert -size 4480x1440 xc:black \( '<path to file>'  -scale 2560x1440^ -gravity center -extent 2560x1440 +gravity \) -geometry +1920+0 -composite   \( '<path to file>'  -scale 1920x1080^ -gravity center -extent 1920x1080 +gravity \) -geometry +0+0 -composite  jpg:- | feh --no-xinerama --bg-tile --no-fehbg -

oogeek: So I think besides adding `defWPName`, there isn't much that can be done here.

slotThe: Will this work even with the `jpg:-` part?  Or do you need to set it to `png:-` when `defWPName'` (n.b.: perhaps `defWPNamePng`) is used?

oogeek: Actually it is ok to use either `png:-` or `jpg:-`. For example, when specifying `wallpaperbasedir` and let the random function pick one image, it is actually using `png` and convert it to `jpg` (in my case).

oogeek: > Will this work even with the `jpg:-` part? Or do you need to set it to `png:-` when `defWPName'` (n.b.: perhaps `defWPNamePng`) is used?

I think this is really an good suggestion. Maybe I can rename both defWPName and defWPName' to

oogeek: > Will this work even with the `jpg:-` part? Or do you need to set it to `png:-` when `defWPName'` (n.b.: perhaps `defWPNamePng`) is used?

`jpg:-` and `png:-` only specify the final format. Actually as the example I have given, you can mix jpg and png and finally say that the result is a png file.

slotThe: > I think this is really a good suggestion. Maybe I can rename both `defWPName` and `defWPName'` to `defWPNameJpg` and `defWPNamePng`

I think `defWPName` should stay, for backwards compatibility.  `defWPName' -> defWPNamePng` sounds good though!

oogeek: Thanks, I have added it.

XMonad/Hooks/WallpaperSetter.hs
```diff
@@ -100,9 +100,16 @@ defWallpaperConf = WallpaperConf "" $ WallpaperList []
 instance Default WallpaperConf where
     def = defWallpaperConf
 
--- |returns the default association list (maps name to name.jpg, non-alphanumeric characters are omitted)
 defWPNames :: [WorkspaceId] -> WallpaperList
-defWPNames xs = WallpaperList $ map (\x -> (x,WallpaperFix (filter isAlphaNum x++".jpg"))) xs
+defWPNames = defWPNamesJpg
```
slotThe, line 104: You should add something like a 

``` haskell
{-# DEPRECATED defWPNames "Use defWPNamesJpg instead" #-}
```

here instead

XMonad/Hooks/WallpaperSetter.hs
```diff
@@ -100,9 +100,16 @@ defWallpaperConf = WallpaperConf "" $ WallpaperList []
...
+
+-- |returns the default association list (maps name to name.jpg, non-alphanumeric characters are omitted)
+defWPNamesJpg :: [WorkspaceId] -> WallpaperList
+defWPNamesJpg xs = WallpaperList $ map (\x -> (x, WallpaperFix (filter isAlphaNum x ++ ".jpg"))) xs
+
+-- |returns the default association list (maps name to name.png, non-alphanumeric characters are omitted)
+defWPNamesPng :: [WorkspaceId] -> WallpaperList
+defWPNamesPng xs = WallpaperList $ map (\x -> (x, WallpaperFix (filter isAlphaNum x ++ ".png"))) xs
```
slotThe, line 112: Perhaps instead of replicating almost the same documentation, highlight the differences instead.

For example,

``` haskell
-- | Like `defWPNamesJpg`, but map @name@ to @name.png@ instead.
```

slotThe (commented):
Aside from a few smal things, this looks good now.  @TheMC47 what do you think?

XMonad/Hooks/WallpaperSetter.hs
oogeek: Thanks for pointing out. Should I change the CHANGES.md for that? Like `defWPNames` is deprecated?

slotThe: Yeah we usually include this in the `CHANGES.md`, good thinking :)

PR #533, opened 2021-05-01
Update issue and pull-request templates

slotThe:
### Description

- Clarify that we would really like a full template and not just a snippet where we'll have to guess the context.
- Change the `xmonad-testing` checkbox to something more generic, so people are more inclined to actually fill it in. 

### Checklist

  ~~- [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)~~

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,7 +7,8 @@ behind them.
 
   - [ ] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)
 
-  - [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)
+  - [ ] I've considered how to best test these changes (property, unit,
+        manually, ...) and concluded: XXX
 
   - [ ] I updated the `CHANGES.md` file
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -9,8 +9,8 @@ Give detailed step-by-step instructions on how to reproduce the problem.
 
 ### Configuration File
 
-Please include the smallest configuration file that reproduces the
-problem you are experiencing:
+Please include the smallest _full_ configuration file that reproduces
+the problem you are experiencing:
 
 ```haskell
 module Main (main) where
@@ -25,4 +25,6 @@ main = xmonad def
 
   - [ ] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)
 
-  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)
+  - I tested my configuration
+    - [ ] With `xmonad` version XXX (commit XXX if using git)
+    - [ ] With `xmonad-contrib` version XXX (commit XXX if using git)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,7 +7,7 @@ behind them.
 
   - [ ] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)
 
-  - [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)
+  - [ ] I tested my changes with commit XXX
```
oogeek, line 10: Maybe also add something like:
`I tested my changes with the latest git version`

Is that appropriate?

slotThe: > `I tested my changes with the latest git version`

I was thinking about this, but what is "the latest git version" may very well change in the course of the pr, especially when it's a longer affair, so I think explicitly specifying the commit is better. 

> Also specify `xmonad` and `xmobar` version.
> For example, some users will mix `xmonad-contrib git` with `xmonad 0.15`.

That's a good suggestion, thanks!  I don't think xmobar is necessary, but the version of the xmonad core would certainly be interesting.

TheMC47: I think mentioning the commit is better in this case, but maybe also mention a version? Something like

  - I tested my changes
    - [ ] With `xmonad`: version XX  (commit XXX, if building from git)
    - [ ] With `xmonad-contrib`: version XX (commit XXX, if building from git)


that might be too verbose though

slotThe: I don't think verbosity is really something that we should care about in a template like this; as long as we don't paste entire essays there :)

I'm curious how common it is to be on a non-git xmonad version when submitting pr's.  If that's a thing that people generally do then it would definitely be worth adding (not for contrib though, I'd expect people to have tested their pr's against _some_ version of contrib git)

TheMC47 (approved):
I think this is good to go. We could wait for more feedback if you want

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,7 +7,9 @@ behind them.
 
   - [ ] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)
 
-  - [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)
+  - I tested my changes
+    - [ ] With `xmonad` version XXX (commit XXX if using git)
+    - [ ] With `xmonad-contrib` commit XXX
```
liskin, line 12: What's the point of asking for xmonad-contrib commit in a pull request? All the info is available in the git history itself. This whole `I tested my changes with xmonad/xmonad-contrib version XXX` thing makes sense only in the issue template (where it's sorely missing), but not here, IMO.

(xmonad version would possibly make sense in the PR template, if xmonad was actually changing, but it isn't really, and people who make PRs that actually depend on some recently merged or work in progress functionality of xmonad core or X11 usually mention it anyway)

slotThe: > What's the point of asking for xmonad-contrib commit in a pull request? All the info is available in the git history itself. 

I saw it as more of a bookkeeping device for longer PRs.  This way, we at least know a version where it definitely worked, should any regressions be introduced later.  Maybe that's overthinking it...

> (xmonad version would possibly make sense in the PR template, if xmonad was actually changing, but it isn't really, and people who make PRs that actually depend on some recently merged or work in progress functionality of xmonad core or X11 usually mention it anyway)

You're probably right on that one. Xmonad doesn't change too much and so perhaps it's superfluous to ask a specific commit there.

> This whole `I tested my changes with xmonad/xmonad-contrib version XXX` thing makes sense only in the issue template (where it's sorely missing), but not here, IMO.

That's a good idea!

slotThe: @liskin so what do you think?  Keep it like this?  Completely remove the "I've tested my changes" from the pull request?  Have a passive-aggressive "I've tested my changes at all" there? :)

liskin: Unless we have an alternative that is better, I'd say drop it.

We could possibly try encouraging people to add property/unit tests where that makes sense, but then we risk "incidents" like #407/#430 stalling for half a year when the submitter (in this case me) procrastinates on adding the tests. :-)

Or we could try to improve xmonad-testing into something that is actually useful. Having contrib modules be tested somehow, even if that's just testing that configuration samples typecheck, would be nice, but right now it's more hassle than it's worth. Ideally this would be much less pain, like a doctest or something that tests the docs in xmonad-contrib effortlessly. But all this requires hard work that someone would have to put in, and I don't think I have the time for that, so right now the best way forward is to either drop the line from the PR template, or replace with something like "I've considered how to best test this (property, unit, manually, …) and concluded that …" to let people know that some testing is encouraged and that they should at least think about it for a while.

slotThe: > or replace with something like "I've considered how to best test this (property, unit, manually, …) and concluded that …" to let people know that some testing is encouraged and that they should at least think about it for a while.

Oh I really like that, thanks!

PR #534, opened 2021-05-03
update sampleHook doc in X.A.CopyWindow

oogeek:
### Description
So to update X.A.CopyWindow to fit with the newly introduced X.H.StatusBar, I rewrite the `sampleHook` for checking multiple copies of windows on different workspaces.

Also note that this is a `draft`. Right now we need to wait for #463 to be merged.

I am wondering if there are easier way to implement this. So that I don't need a new `StatusBarPropTo'`.

### Checklist


XMonad/Actions/CopyWindow.hs
```diff
@@ -88,6 +90,35 @@ import qualified XMonad.StackSet as W
...
+--
+-- > statusBarPropTo' :: String -- ^ Property to write the string to
+-- >                 -> String -- ^ The command line to launch the status bar
+-- >                 -> PP   -- ^ The pretty printing options
+-- >                 -> StatusBarConfig
+-- > statusBarPropTo' prop cmd pp = def
+-- >     { sbLogHook     = xmonadPropLog' prop =<< dynamicLogString . (\w -> pp { ppHidden = w }) =<< checkCopy
+-- >     , sbStartupHook = spawnStatusBarAndRemember cmd
+-- >     , sbCleanupHook = cleanupStatusBars
+-- >     }
```
TheMC47, line 119: The whole point of the abstractions is to not fiddle with low-level things anymore. All the `statusBar*` functions accept an `X PP`, so it's best to write a function that takes a `PP` and makes it aware of copies. Something like `copiesPP :: (WorkspaceId -> String) -> PP -> X PP`, and then apply it like so `statusBarProp "xmobar" (copiesPP myContainsCopies def)`

I also think that the examples should be minimal: No need to use `statusBarPropTo` with `xmobar -x 0`. A better way is to prefer `statusBarProp "xmobar"` and make the readers read the documentation to see what other options are possible

XMonad/Actions/CopyWindow.hs
```diff
@@ -79,6 +79,8 @@ import qualified XMonad.StackSet as W
 -- To distinguish workspaces containing copies of the focused window use
 -- something like:
 --
+-- Case 1: Using 'spawnPipe'
```
TheMC47, line 82: I say we drop all the `spawnPipe` stuff. We don't want people using that

TheMC47 (changes requested):
Thanks for your work! I quickly took a look at the changes, and I left you some comments. I still haven't tested any of the code. I'll tell you when I do.

XMonad/Actions/CopyWindow.hs
oogeek: Thanks for your suggestion. I have deleted that.

XMonad/Actions/CopyWindow.hs
oogeek: I have upated the example, thanks.

XMonad/Actions/CopyWindow.hs
```diff
@@ -76,18 +76,20 @@ import qualified XMonad.StackSet as W
...
+-- First use a check copy function that takes a pretty printer and makes it aware of copies.
+--
+-- > copiesPP :: PP -> X PP
+-- > copiesPP pp = do
+-- >     copies <- wsContainingCopies
+-- >     let check ws | ws `elem` copies = xmobarColor orange "" ws 
+-- >                  | otherwise = xmobarColor yellow "" ws
+-- >     return pp { ppHidden = check }
+-- 
+-- Then use @copiesPP@ with 'statusBarProp':
```
slotThe, line 89: You need to do 'XMonad.Hooks.StatusBar.statusBarProp' here, otherwise I don't think Haddock will link that (since the module itself does not import it). 

Might also be worth to explicitly import it, not sure

oogeek: Thanks!!! I just ignore it.

XMonad/Actions/CopyWindow.hs
TheMC47: I think you missed my point, why not just add `copiesPP` as a library function?

XMonad/Actions/CopyWindow.hs
```diff
@@ -76,18 +77,20 @@ import qualified XMonad.StackSet as W
...
+copiesPP wtoS pp = do
+    copies <- wsContainingCopies
+    let check ws | ws `elem` copies = wtoS ws 
+                 | otherwise = (ppHidden pp) ws
+    return pp { ppHidden = check }
+
+-- Then use @copiesPP@ with 'XMonad.Hooks.StatusBar.statusBarProp' :
+--
+-- > mySB = statusBarProp "xmobar" (copiesPP xmobarPP)
+-- > main = xmonad $ withEasySB mySB defToggleStrutsKey def
```
slotThe, line 93: This doesn't do what you think it does; these kinds of haddock headings have to be unboken comments.  In this case, the first part gets rendered under `$logHook`, then there's the function, and no user will ever see the bottom part.

TheMC47: I forgot to hit "submit review". @oogeek try to generate the docs yourself and see how they look

liskin: Same problem is here, btw: https://github.com/xmonad/xmonad-contrib/pull/501/files#diff-8b3eac5d6797ca9ec49c63a17e33213c05a92f32212bacd27f90e30b163c9b15R202

oogeek: Thanks for pointing out. Really sorry for my carelessness.

XMonad/Actions/CopyWindow.hs
```diff
@@ -76,18 +76,21 @@ import qualified XMonad.StackSet as W
...
--- >                 | otherwise = pad ws
--- >    dynamicLogWithPP myPP {ppHidden = check, ppOutput = hPutStrLn h}
--- >
--- > main = do
--- >    h <- spawnPipe "xmobar"
--- >    xmonad def { logHook = sampleLogHook h }
+--
+-- To distinguish workspaces containing copies of the focused window, use the functions below.
+-- First use 'copiesPP' that takes a pretty printer and makes it aware of copies.
+-- Then use 'copiesPP' with 'XMonad.Hooks.StatusBar.statusBarProp' :
```
slotThe, line 82: I'm not sure I follows these instructions very well; first use `copiesPP` and then use it again?  Do you want to say first build a `StatusBarConfig` with `copiesPP` and then use _that_ with `statusBarProp`?

oogeek: Thanks I have updated that. Let me generate the doc again to see if it looks OK.

XMonad/Actions/CopyWindow.hs
```diff
@@ -76,18 +76,23 @@ import qualified XMonad.StackSet as W
...
--- >    xmonad def { logHook = sampleLogHook h }
+--
+-- To distinguish workspaces containing copies of the focused window, use 'copiesPP'.
+-- 'copiesPP' takes a pretty printer and makes it aware of copies of the focused window.
+-- It can be applied when creating a @StatusBarConfig@ from "XMonad.Hooks.StatusBar".
+--
+-- A sample config looks like this:
+--
+-- > mySB = statusBarProp "xmobar" (copiesPP xmobarPP)
+-- > main = xmonad $ withEasySB mySB defToggleStrutsKey def
```
TheMC47, line 87: This doesn't compile though. `copiesPP` takes another parameter before the `PP`. To be consistent, I think we should use the same function as previously (e.g `pad . xmobarColor "red" "black"`). I'd also use `def` instead of `xmobarPP`

TheMC47 (approved):
Please squash the commits

oogeek:
> Please squash the commits

I have squashed it. Do I need to update `CHANGES.md`, since `CopiesPP` is introduced and exported.

TheMC47:
Yes that also needs to be done

TheMC47:
I'll merge this soon if no one objects

XMonad/Actions/CopyWindow.hs
```diff
@@ -76,18 +76,23 @@ import qualified XMonad.StackSet as W
...
+-- To distinguish workspaces containing copies of the focused window, use 'copiesPP'.
+-- 'copiesPP' takes a pretty printer and makes it aware of copies of the focused window.
+-- It can be applied when creating a @StatusBarConfig@ from "XMonad.Hooks.StatusBar".
+--
+-- A sample config looks like this:
+--
+-- > mySB = statusBarProp "xmobar" (copiesPP (pad . xmobarColor "red" "black") xmobarPP)
+-- > main = xmonad $ withEasySB mySB defToggleStrutsKey def
+
+-- | Takes a pretty printer and makes it aware of copies.
```
liskin, line 89: "Takes a pretty printer and makes it aware of copies by using the provided function to show hidden workspaces that contain copies of the focused window." might be a better doc here.

liskin (changes requested):

PR #538, opened 2021-05-09
`X.A.DynamicWorkspaceGroups`: TopicSpace support

TheMC47:
### Description

This adds `viewTopicGroup` and a corresponding prompt. This is similar
to `viewWSGroup`, but it calls `switchTopic` instead of `W.greedyView`,
in order to run the topic action on the workspace.

### Checklist

        manually, ...) and concluded: it works

### DIFF
--- a/XMonad/Actions/DynamicWorkspaceGroups.hs
+++ b/XMonad/Actions/DynamicWorkspaceGroups.hs
@@ -34,6 +34,10 @@ module XMonad.Actions.DynamicWorkspaceGroups
     , promptWSGroupForget
 
     , WSGPrompt
+     -- * TopicSpace Integration
+     -- $topics
+    , viewTopicGroup
+    , promptTopicGroupView
     ) where
 
 import Data.List (find)
@@ -45,6 +49,7 @@ import qualified XMonad.StackSet as W
 
 import XMonad.Prompt
 import qualified XMonad.Util.ExtensibleState as XS
+import XMonad.Actions.TopicSpace
 
 -- $usage
 -- You can use this module by importing it into your ~\/.xmonad\/xmonad.hs file:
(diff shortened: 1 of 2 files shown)


slotThe (approved):
I like this!  

I won't have a multi-head setup ready for some days (and the code looks straightforward enough really), but I'll test this asap.  If someone else can confirm that this works then feel free to merge it before that though :)

TheMC47:
Awesome! take your time with it, this isn't at all urgent

XMonad/Actions/DynamicWorkspaceGroups.hs
```diff
@@ -102,21 +107,25 @@ forgetWSGroup :: WSGroupId -> X ()
 forgetWSGroup = XS.modify . withWSG . M.delete
 
 -- | View the workspace group with the given name.
-viewWSGroup :: WSGroupId -> X ()
-viewWSGroup name = do
+viewWSGroup ::  WSGroupId -> X ()
```
liskin, line 110: Why the added space?

TheMC47: Because I trained my brain to auto-format each file and not check to format manually... and I can't autoformat files that I didn't write

XMonad/Actions/DynamicWorkspaceGroups.hs
```diff
@@ -145,3 +158,20 @@ promptWSGroupForget :: XPConfig -> String -> X ()
 promptWSGroupForget xp s = do
   gs <- fmap (M.keys . unWSG) XS.get
   mkXPrompt (WSGPrompt s) xp (mkComplFunFromList' xp gs) forgetWSGroup
+
+-- $topics
+-- You can use this module with "XMonad.Actions.TopicSpace" - just replace
```
slotThe, line 163: If you want to be correct to an obnoxious degree, this should be an em dash (—).  This is usually emulated in ASCII as three dashes `---`.  If not just ignore :)

slotThe (commented):
@TheMC47 when testing this just now I noticed that things get a bit weird when the topics that one visits all have startup actions that spawn things. 

Say I have two monitors `a` and `b` and two workspaces `1` and `2`.  Both workspaces spawn a terminal when they're visited and empty.  If I save the configuration `*a:1, b:2` (`a` focused) and then switch to that while both workspaces are empty, the terminals will both spawn on workspace `1`. 

This is a failing of the respective topic actions, just spawning a terminal and not thinking about the workspaces at all.  They should instead spawn this via `spawnOn` or similar functionality. I think this will probably occur often enough that it's at least worth noting in the docs somewhere.

TheMC47:
> This is a failing of the respective topic actions, just spawning a terminal and not thinking about the workspaces at all. They should instead spawn this via `spawnOn` or similar functionality. I think this will probably occur often enough that it's at least worth noting in the docs somewhere.

That's definitely something that needs to be mentioned

PR #539, opened 2021-05-10
X.A.TopicSpace: Add 'TopicItem'

slotThe:
### Description

1. Add a new type 'TopicItem', as well as several helpers.  This is so users don't have to specify things in two places and then use some extra function to check if they've not made any mistakes.
2. Update the `Usage` documentation of TopicSpace; fix some inconsistencies and tell users that there is an alternative way.

Concerning the second part, I wonder if we should just drop "the old way" and tell users that going through 'TopicItem' is just how one should configure the module.  Personally, I think it's a straight upgrade over having to manually specify everything.  Opinions?

Cc: @byorgey because I originally took the definition of `TopicItem` out of [your configuration](https://github.com/byorgey/dotfiles/blob/master/xmonad.hs#L170).  

### Checklist

        manually, ...) and concluded: most of this is docs and the added functions are trivial

### DIFF
--- a/XMonad/Actions/TopicSpace.hs
+++ b/XMonad/Actions/TopicSpace.hs
@@ -26,8 +26,12 @@ module XMonad.Actions.TopicSpace
   , TopicConfig(..)
   , TopicItem(..)
 
-    -- * Default Topic Config
-  , def
+    -- * Managing 'TopicItem's
+  , topicNames
+  , tiActions
+  , tiDirs
+  , noAction
+  , inHome
 
     -- * Switching and Shifting Topics
   , switchTopic
@@ -40,13 +44,6 @@ module XMonad.Actions.TopicSpace
   , topicActionWithPrompt
   , topicAction
   , currentTopicAction
-
-    -- * Managing 'TopicItem's
-  , topicNames
-  , tiActions
-  , tiDirs
-  , noAction
-  , inHome
 
     -- * Getting the Topic History
   , getLastFocusedTopics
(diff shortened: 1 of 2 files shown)


TheMC47:
> Concerning the second part, I wonder if we should just drop "the old way" and tell users that going through 'TopicItem' is just how one should configure the module. Personally, I think it's a straight upgrade over having to manually specify everything. Opinions?

I also think it's better this way, especially with the helpers you added. It might be worth it to make  `TopicConfig` private and add a `[TopicItem] -> XConfig l -> XConfig l` modifier that generates the `TopicConfig` in the startupHook, and saves it the state. That way we can drop `TopicConfig` as a parameter of the other functions, and I think it's a better interface overall. I'm not sure about this though, and we might want to wait for another release to do a proper deprecation, since a lot of people's configurations will break

slotThe:
@TheMC47 would you care to give these new docs a read?

XMonad/Actions/TopicSpace.hs
```diff
@@ -105,158 +112,126 @@ import XMonad.Util.Run (spawnPipe)
...
+-- > promptedGoto = workspacePrompt def goto
 -- >
 -- > promptedShift :: X ()
--- > promptedShift = workspacePrompt myXPConfig $ windows . W.shift
+-- > promptedShift = workspacePrompt def $ windows . W.shift
 -- >
--- > -- Toggle between the two most recently used topics while filtering
--- > -- out the scratchpad topic.
+-- > -- Toggle between the two most recently used topics, but keep
+-- > -- screens separate.  This needs @workspaceHistoryHook@.
```
TheMC47, line 192:
```suggestion
-- > -- screens separate. This needs @workspaceHistoryHook@.
```

XMonad/Actions/TopicSpace.hs
```diff
@@ -105,158 +112,126 @@ import XMonad.Util.Run (spawnPipe)
...
 --
--- We can now put the whole configuration together with the following (while
--- also checking that we haven't made any mistakes):
+-- This makes @M-1@ to @M-9@ switch to the first nine topics that we
+-- have specified in @topicItems@.
+--
+-- You can also switch to the nine last-used topics instead:
+--
+-- >   [ ("M-" ++ k, switchNthLastFocused myTopicConfig i)
+-- >   | (i, k) <- zip [1..] (map show [1 .. 9 :: Int])
```
TheMC47, line 225: Why not use `show i` instead of k? I think the zip is unnecessary here

TheMC47 (approved):
Sorry for taking too long. I just have some minor nitpicks

XMonad/Actions/TopicSpace.hs
slotThe: Putting two spaces after a period is a habit I don't think I will be able to kick :)  It is at least consistent throughout the module

XMonad/Actions/TopicSpace.hs
slotThe: You're completely right, I'll change that

XMonad/Actions/TopicSpace.hs
TheMC47: If it's consistent, then it's fine. I only looked at the diff and it felt odd. Don't change this then

PR #544, opened 2021-05-14
Add Sponsor button to GitHub

liskin:
### Description

These past months I've spent a lot of time working on xmonad and I feel like I've done a lot. This is, however, not sustainable long term. :-(

I'd like to try making my GitHub Sponsors profile a bit more visible, hoping that would allow me to continue dedicating time to xmonad.

I know that the correct approach probably is for the xmonad project to find a fiscal sponsor like the Software Freedom Conservancy or Open Collective or something and accept donations as a project, and then redistribute that to people, but I don't think the project has enough momentum to do something as complicated as that. :-/

---

@xmonad/core @slotThe @TheMC47 Are you folks okay with me doing this? Can I put that on https://github.com/xmonad/xmonad as well? Would you like to add your profile as well (I checked all of your profiles and none of you have a Sponsors profile yet, though)?

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,2 @@
+github:
+  - liskin


geekosaur:
I'm not averse to it and in fact kinda wish I had my shit together enough to do so myself and maybe get working on some of the ideas I've had including proper EWMH layer (including floating layer) support.

TheMC47:
I'm personally fine with this, and I think it's fair regarding the amount of work you're putting in. 

> I know that the correct approach probably is for the xmonad project to find a fiscal sponsor like the Software Freedom Conservancy or Open Collective or something and accept donations as a project, and then redistribute that to people, but I don't think the project has enough momentum to do something as complicated as that. :-/

Well, at least not for time being. However, If a big number of people want to also get sponsored, then it might be more fair and transparent to set up something like that. But it's not worth the hassle yet

Although not directly related, we should also review the [credits section in the website](https://github.com/xmonad/xmonad-web/blob/gh-pages/community.md#credits) and update it (maybe split it into a "Hall of Fame" for the no-longer active maintainers and a "Team" section for the current maintainers, but I haven't been around that long to know how to classify them)

liskin:
> However, If a big number of people want to also get sponsored, then it might be more fair and transparent to set up something like that.

Not just that, it will actually be necessary as GitHub won't let us put more than 4 accounts in there anyway. I was thinking that if this problem ever arises, we might just sponsor one another to distribute the income. It'd be lovely to have that problem, but I wouldn't kid myself just now… :-)

psibi:
> @xmonad/core @slotThe @TheMC47 Are you folks okay with me doing this?

I'm fine with it and don't see any issues.

slotThe:
On Fri, May 14 2021 09:21, Tomáš Janoušek wrote:
> @xmonad/core @slotThe @TheMC47 Are you folks okay with me doing this?
> Can I put that on https://github.com/xmonad/xmonad as well?

:+1:

Issue #517, opened 2021-04-14
X.L.WindowSwitcherDecoration and X.L.DraggingVisualizer stops working


oogeek:
### Problem Description
`X.L.WindowSwitcherDecoration` and `X.L.DraggingVisualizer stops working` work well with
```
xmonad 0.15
xmonad-contrib 0.16
```
But not working for
```
xmonad git
xmonad-contrib git
```
or 
```
xmonad 0.15
xmonad-contrib git
```
### Configuration File
```
myvisualL = windowSwitcherDecoration shrinkText def (draggingVisualizer myLayoutHook)

myLayoutHook = avoidStruts $ mouseResize $ windowArrange $ T.toggleLayouts floats
               $ mkToggle (NBFULL ?? NOBORDERS ?? EOT) myDefaultLayout
             where
               myDefaultLayout =     tall
                                 ||| floats
                                 ||| noBorders tabs
                                 ||| grid
                                 ||| spirals
                                 ||| Accordion
```
### Checklist

  - [X] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

oogeek:
I have tried to `git diff` for a while but can not find the issue.

liskin:
Can you please elaborate on what `stops working work well` means?

oogeek:
> Can you please elaborate on what "stops working work well" means?
> 
> (Also, I believe `git bisect` might be a better tool for this than `git diff`. ;-))

Behavior it used to be:

    left click the window bar (top) and the hold it while moving the whole window to overlap another one

    these two windows change position

Now it simply does not work.

liskin:
Yes, I can reproduce this. As this seems to be a regression since last version, we definitely want to fix this in the next release, so I added it to the milestone/project.

If you have some free time, a git bisect of xmonad-contrib between the v0.16 and master would definitely help us. I (or anyone who beats me to it) will get to this sooner or later, but there's loads of other higher priority stuff, some of which is even in progress.

oogeek:
> If you have some free time, a git bisect of xmonad-contrib between the v0.16 and master would definitely help us. I (or anyone who beats me to it) will get to this sooner or later, but there's loads of other higher priority stuff, some of which is even in progress, so I'd rather not put more on my plate right now.

I am currently working on it. Still git bisecting

[...]

PR #545, opened 2021-05-14
X.H.ManageDocks: Fix button events on decoration windows

liskin:
### Description

Decoration windows are created using XMonad.Util.XUtils.createNewWindow
which happens to set _NET_WM_WINDOW_TYPE to _NET_WM_WINDOW_TYPE_DESKTOP,
and ManageDocks considers such windows candidates for struts and
therefore requests property events, thus overriding the original event
mask requested by decorations.

The fix is to first obtain the current event mask, set the required bits
in it only then reset the mask.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/517
Fixes: ec146171238b ("X.H.ManageDocks: React to strut updates of override_redirect docks")
Related: https://github.com/xmonad/X11/pull/77

### Checklist

        manually, ...) and concluded:

    I tested this manually and I've verified that dragging works now and
    that https://github.com/xmonad/xmonad-contrib/pull/492 isn't
    regressed either. We do not and will not have the infrastructure to
    write an automated test for this, unfortunately.

    This is a recent regression of git master, no need to clutter the
    CHANGES.

### DIFF
--- a/XMonad/Hooks/ManageDocks.hs
+++ b/XMonad/Hooks/ManageDocks.hs
@@ -142,7 +142,8 @@ manageDocks = checkDock --> (doIgnore <+> doRequestDockEvents)
 -- (Only if not already a client to avoid overriding 'clientMask')
 requestDockEvents :: Window -> X ()
 requestDockEvents w = whenX (not <$> isClient w) $ withDisplay $ \dpy ->
-    io $ selectInput dpy w (propertyChangeMask .|. structureNotifyMask)
+    withWindowAttributes dpy w $ \attrs -> io $ selectInput dpy w $
+        wa_your_event_mask attrs .|. propertyChangeMask .|. structureNotifyMask
 
 -- | Checks if a window is a DOCK or DESKTOP window
 checkDock :: Query Bool


geekosaur:
Should `createNewWindow` even be doing that? It sounds wrong for most decorations.

liskin:
CI is expected to fail because this bumps the X11 dependency, but we first need to merge https://github.com/xmonad/X11/pull/77, release X11 0.10, bump the upper bound in https://github.com/xmonad/xmonad (including adding a revision for old releases on Hackage), and only then will this build.

geekosaur:
There's also a typo in the cabal file.

liskin:
> Should `createNewWindow` even be doing that? It sounds wrong for most decorations.

https://github.com/xmonad/xmonad-contrib/blob/246926911940ccff1f6fad13d82ceb5c91cd01dc/XMonad/Util/XUtils.hs#L66:
`-- @@@ ugly hack to prevent compositing`

liskin:
> There's also a typo in the cabal file.

Good catch! Force pushed (sorry).

Issue #526, opened 2021-04-22
XMonad.Prompt xprop


ceigh:
### Problem Description

I recently started using xmonad prompt, and I also use picom as a composite manager.
So, using picom, I set shadows for windows, but on xmonad prompt, I would like to remove the shadow, usually this is done by adding the window class to the list in picom.

But when I try to get information about the prompt window (as usual with xprop), xprop returns an empty result.
If someone has encountered a similar problem, please tell me how to define prompt in picom. Thanks.

### Configuration File

[xmonad](https://github.com/ceigh/dotfiles/blob/f137cf4ccd55a4469af03710f40811a21bef83d1/.xmonad/xmonad.hs#L80)
[picom](https://github.com/ceigh/dotfiles/blob/master/.config/picom/picom.conf)

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

geekosaur:
I don't think the prompt window currently gets any identifying information; ideally it should at least have a valid `WM_CLASS` assigned (`XMonad`,`xmonad-prompt` or similar) so things like this will work.

geekosaur:
Since it's unusual for windows to not have `WM_CLASS`, you may be able to match the window with `!WM_CLASS:8s`. This might also match other windows, but it seems unlikely. (Normally the pattern would have an `@`, but xmonad is non-reparenting so there should be no distinction.)

ceigh:
Yes, thanks it works!
According to this guide, I used `"!WM_CLASS:s"`:

```conf
shadow-exclude = [
  "!WM_CLASS:s"
];
```

geekosaur:
I'm going to leave this open because we should really have a proper `WM_CLASS` on the window.

ceigh:
Yes, I think this is correct, thank you again.

geekosaur:
I also think I reversed the resource and class names above, whoops.

slotThe:
Seems like `xSetClassHint` is not implemented by X11, so that would
probably have to be done first.

slotThe:
The necessary functionality is now implemented by xmonad/X11/pull/76; but that's not released quite yet.  

[...]

Issue #369, opened 2020-08-07
XMonad.Layout.Tabbed doesn't have WM_CLASS


Tanish2002:
### Problem Description

I am using the tabbed layout with a fork of picom(compositor) which provides rounded corner, and I want to disable rounded corners for the tabs but when I use `xprop` to find the `WM_CLASS` string there is none

Output of `xprop`
https://i.fiery.me/nR4vo.png

### Configuration File

Here is how I use the layout

```haskell
import XMonad.Layout.Tabbed

myLayout = fullscreenFull $ avoidStruts $ (dwindle ||| tabs ||| tall ||| Mirror tall ||| Full)
  where
    tabs =
      named "Tabbed"
      . spacingRaw False (Border 20 20 20 20) True (Border 0 20 0 20) False
      $ tabbedBottom shrinkText tabTheme
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

liskin:
Fixed by https://github.com/xmonad/xmonad-contrib/pull/546

PR #546, opened 2021-05-15
Set `WM_CLASS` for some windows

slotThe:
### Description

Fixes:
  - https://github.com/xmonad/xmonad-contrib/issues/526
  - https://github.com/xmonad/xmonad-contrib/issues/369

Set the WM_CLASS property for long-living windows that need it.  This
helps users to properly match on these windows for certain tasks.

This needs a new X11 release, hence it's a draft for now.  I also didn't change the bounds in the cabal file, as #545 has already done this and I expect this will get merged sooner. 

I was a bit conservative with what to set properties for.  Things like X.U.EZConfig also create windows that may very well have such a property, but I didn't think they were long-lived enough to warrant it really.  Feel free to point me to more modules that need it and I'll add them.

### Checklist

        manually, ...) and concluded: 

    I tested this by hand and it works.  I don't think we can really tests these kinds of interaction with the X server.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,9 @@
       constraint (was: `IO`), due to changes in how the xmonad core handles XDG
       directories.
 
+    - The prompt window now sets a `WM_CLASS` property.  This allows
+      other applications, like compositors, to properly match on it.
+
   * `XMonad.Hooks.EwmhDesktops`
 
     - It is no longer recommended to use `fullscreenEventHook` directly.
@@ -250,6 +253,10 @@
   * `XMonad.Actions.TreeSelect`
 
     - Fix swapped green/blue in foreground when using Xft.
+
+    - The spawned tree-select window now sets a `WM_CLASS` property.
+      This allows other applications, like compositors, to properly
+      match on it.
 
   * `XMonad.Layout.Fullscreen`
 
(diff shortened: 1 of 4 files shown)


liskin (approved):
Looks okay, might be nice to mention the motivation in the changelog as well so that people know what they can use it for.

slotThe:
Sure, I've rebased and mentioned this in the CHANGES

CHANGES.md
```diff
@@ -27,6 +27,9 @@
       constraint (was: `IO`), due to changes in how the xmonad core handles XDG
       directories.
 
+    - The prompt window now sets a `WM_CLASS` property.  This allows to
+      match on it in a manageHook.
```
liskin, line 31: Seems to me you also forgot what the motivation was. xmonad wouldn't be stupid enough to try to manage its own windows would it? This was more about shadows and rounded corners in a compositor, I think.

slotThe: Oh :)  Very good point yes, I'll change that

liskin:
Note that https://github.com/xmonad/xmonad-contrib/issues/526 and https://github.com/xmonad/xmonad-contrib/issues/369 as GitHub apparently doesn't understand your multiline trailers. Might want to consider using the standard git trailers instead.

slotThe:
On Wed, Jun 02 2021 03:49, Tomáš Janoušek wrote:
> Note that https://github.com/xmonad/xmonad-contrib/issues/526 and
> https://github.com/xmonad/xmonad-contrib/issues/369 didn't get closed
> as GitHub apparently doesn't understand your multiline trailers. Might
> want to consider using the standard git trailers instead.

Oh good point, thanks.  I'll do that next time

Issue #550, opened 2021-05-27
flameshot: keyboard shortcuts don't work


liskin:
flameshot is a popular screenshot tool. Unfortunately its keyboard shortcuts don't seem to work in xmonad.
More here: https://github.com/flameshot-org/flameshot/issues/773

An important clue is here: https://github.com/flameshot-org/flameshot/issues/773#issuecomment-698892896
Turns out flameshot doesn't seem to grab the keyboard, it probably just takes input focus and then sends a ClientMessage with _NET_WM_STATE_FULLSCREEN, which should be pointless (its window isn't managed), but EwmhDesktops processes that message anyway and the (useless) call to `windows` resets the input focus.

Adding an `isClient` check to `fullscreenEventHook` should fix this, and we should probably add that check to all the other places that process ClientMessages (see e.g. https://github.com/xmonad/xmonad-contrib/commit/9e6e521fbba2be9d7a675a1e99dcaaf8ace54283 why this might be a good idea).
This is only a partial fix, however, as any other event that causes xmonad to refresh its state and reset focus will break those keyboard shortcuts. Often there won't be any such event, but we can't prevent it. I wonder

davama:
Had similar issues, but specifically with ESC key: https://github.com/flameshot-org/flameshot/issues/1072

honestly havent tried the shortcuts.... 

using xmonad from latest github master branch 

Looking forward to see what becomes of this

liskin:
PR: https://github.com/xmonad/xmonad-contrib/pull/552

l29ah:
Tried flameshot-0.10.1 on todays master of xmonad and xmonad-contrib, yet still `flameshot gui` ignores my Enter hits, and it turns out they are delivered to the window that was (is?) in focus when flameshot was started.

liskin:
That's really weird. You're right that Enter doesn't work (I have flameshot 0.9.0), but it's not getting delivered to anything else here, and the other keys (Escape) work.

liskin:
Oh, it's actually not weird at all, I disabled the imgur uploader manually, and since 0.10.1 its hotkey is disabled for everyone. So it's just the help text of flameshot which is wrong. :-)

PR #552, opened 2021-05-28
X.H.EwmhDesktops: Ignore _NET_WM_STATE_FULLSCREEN from unmanaged windows

liskin:
### Description

This prevents an unnecessary refresh.

That refresh would normally be harmless but it does reset the input focus, which happens to upset some non-conforming clients such as https://github.com/flameshot-org/flameshot. Flameshot is an interactive screenshot tool that creates an override-redirect fullscreen window to let the user select a rectangle to capture and then allows drawing and adding text and so on, but it unfortunately doesn't follow ICCCM recommendations:

> If it is necessary for a client to receive keystrokes on an override-redirect window, either the client must grab the keyboard, or the client must have another top-level window that is not override-redirect and that has selected the Locally Active or Globally Active focus model.

Instead, it just takes input focus and hopes for the best. And it also sends an entirely useless `_NET_WM_STATE_FULLSCREEN` request, which would trigger a refresh and take that focus away.

This commit works around that by not handling that useless `_NET_WM_STATE_FULLSCREEN` for unmanaged windows and thus preventing that refresh. It's just a workaround, however: if a legitimate refresh is necessary at any point, the focus _will_ be taken away.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/550
Fixes: https://github.com/flameshot-org/flameshot/issues/773

### Checklist

        manually, ...) and concluded:

    I tested this manually.

    Not sure if this is worth a changelog entry, as fullscreen handling
    was refactored a bit since 0.16 and will be refactored even more
    before 0.17.

### DIFF
--- a/XMonad/Hooks/EwmhDesktops.hs
+++ b/XMonad/Hooks/EwmhDesktops.hs
@@ -307,6 +307,7 @@ fullscreenStartup = setFullscreenSupported
 -- Note this is not included in 'ewmh'.
 fullscreenEventHook :: Event -> X All
 fullscreenEventHook (ClientMessageEvent _ _ _ dpy win typ (action:dats)) = do
+  managed <- isClient win
   wmstate <- getAtom "_NET_WM_STATE"
   fullsc <- getAtom "_NET_WM_STATE_FULLSCREEN"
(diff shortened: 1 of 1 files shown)


geekosaur (approved):
lgtm

PR #559, opened 2021-06-13
X.P.OrgMode: Add property tests

slotThe:
### Description

Adds property tests for the parser in X.P.OrgMode by constructing a
pretty printer that is a proper inverse of the parser (modulo the fact
that `Maybe A ≅ A`).

This also gave me the opportunity to specify some corner cases that
users are unlikely to run into, but of course quick-check happily
notices :)

- X.P.OrgMode: Only parse actual words

  Instead of trying to find a prefix and then killing the rest of the
  word, actually see whether it at least fits the pattern.  This means
  that

      message +s saturated

  will no longer parse as a scheduled item for saturday, while

      message +s satur

  still will.

- X.P.OrgMode: Parse empty message

  Empty messages seem quite useless, but it's an easy fix for the parser
  to be able to deal with them, so do it.

- X.P.OrgMode: Start counting years at 25

  This is for disambiguation purposes.  Otherwise, there is no way to
  decide whether

      message +s 17 jul 12

  wants to schedule the note for the 17th of july at 12:00, or for
  the 17th of july in the year 12.  The easiest way around this is to
  stipulate that people want to make these notes for the future more
  often than not and to simply prohibit note before the year 25 (as that
  is where the valid times end).

- X.P.OrgMode: Appropriately pad output time

  The standard formatting for org is to left-pad single digits with a
  zero; do that.

### Checklist

        manually, ...) and concluded that it's about time to add tests!

### DIFF
--- a/tests/OrgMode.hs
+++ b/tests/OrgMode.hs
@@ -1,0 +1,157 @@
+{-# OPTIONS_GHC -Wno-orphans             #-}
+{-# OPTIONS_GHC -Wno-incomplete-patterns #-}
+{-# LANGUAGE InstanceSigs       #-}
+{-# LANGUAGE TypeApplications   #-}
+{-# LANGUAGE TupleSections      #-}
+{-# LANGUAGE LambdaCase         #-}
+module OrgMode where
+
+import XMonad.Prelude hiding ((!?))
+import XMonad.Prompt.OrgMode
+
+import qualified Data.Map.Strict as Map
+
+import Data.Map.Strict (Map, (!), (!?))
+import Test.Hspec
+import Test.Hspec.QuickCheck
+import Test.QuickCheck
+
+
+spec :: Spec
+spec = do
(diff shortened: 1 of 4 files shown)


tests/OrgMode.hs
```diff
@@ -0,0 +1,157 @@
...
+
+minuteInt :: Gen Int
+minuteInt = posInt `suchThat` (<= 59)
+
+mbPos :: Num a => Gen (Maybe a)
+mbPos = fmap (fromIntegral . getPositive) <$> arbitrary @(Maybe (Positive Int))
+
+infixr 6 <<>>
+(<<>>) :: (Applicative f, Monoid a) => f a -> f a -> f a
+(<<>>) = liftA2 (<>)
```
slotThe, line 157: I wondered whether this would be something nice for `X.Prelude` for a second (we use this in at least three other modules), but it's perhaps a bit too "out there" for that

Issue #548, opened 2021-05-23
Remove derivation of `Typeable`


slotThe:
### Problem Description

The derivation of `Typeable` has been automatic for quite a while now, yet we still have over a hundred `deriving Typeable` in our codebase.  It would be nice if someone could clean this up!

  - [x] xmonad
  - [ ] xmonad-contrib

PR #561, opened 2021-06-17
Remove all derivations of Typeable

exorcist365:
### Description

behind them.

### Checklist

        manually, ...) and concluded: It works.


slotThe (commented):
This looks nice, thanks!

It would be nice if you could mention the relevant issue (https://github.com/xmonad/xmonad-contrib/issues/548) in the description of the removal commit (backlinks are always nice to have).  We usually do this in the form of

    Related: https://github.com/xmonad/xmonad-contrib/issues/548

at the end of the commit description.

Further, there seem to be a lot of instances of `DeriveDataTypeable` that are now redundant¹, which is why the [CI fails](https://github.com/xmonad/xmonad-contrib/pull/561/checks?check_run_id=2847397529); would be nice if you could remove these as well.

¹: Well, they were redundant before, just that hlint didn't pick up that fact :)

XMonad/Actions/DynamicProjects.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveDataTypeable #-}
+
```
slotThe, line 1: It would be nice if these kinds of things were just the removal of a line and didn't introduce an empty line as well

XMonad/Actions/WorkspaceCursors.hs
```diff
@@ -199,7 +199,7 @@ workspaceCursors :: Cursors String -> l a -> ModifiedLayout WorkspaceCursors l a
 workspaceCursors = ModifiedLayout . WorkspaceCursors
 
 newtype ChangeCursors = ChangeCursors { unWrap :: Cursors String -> X (Cursors String) }
-    deriving (Typeable)
+
```
slotThe, line 202: Same here, these removals should not introduce extra whitespace

slotThe (commented):
Very nice, thanks!  I just noticed a few formatting nits but other than that this is good to go

XMonad/Actions/WorkspaceCursors.hs
exorcist365: Sorry, I fixed it in the latest commit, but for all files. :)

XMonad/Actions/CycleWorkspaceByScreen.hs
```diff
@@ -1,5 +1,3 @@
-
------------------------------------------------------------------------------
```
slotThe, line 2: Seems like this is removing a little too much sometimes :)

exorcist365: Sometimes automation plays a bad joke on us. I am bringing these lines back.

XMonad/Actions/KeyRemap.hs
```diff
@@ -1,4 +1,4 @@
- {-# LANGUAGE DeriveDataTypeable #-}
+
```
slotThe, line 1: Some of these are still in here

XMonad/Actions/LinkWorkspaces.hs
```diff
@@ -14,7 +14,7 @@
 --
 -----------------------------------------------------------------------------
 
-{-# LANGUAGE DeriveDataTypeable  #-}
+
```
slotThe, line 17: I also think these (when the pragma is declared after the module header) weren't picked up

XMonad/Hooks/Focus.hs
```diff
@@ -1,6 +1,6 @@
 {-# LANGUAGE FlexibleContexts       #-}
 {-# LANGUAGE MultiParamTypeClasses  #-}
-{-# LANGUAGE DeriveDataTypeable     #-}
+
```
slotThe, line 3: Also there when the pragma part of a list of pragmas

XMonad/Layout/Mosaic.hs
```diff
@@ -67,7 +67,7 @@ data Aspect
     | Wider
     | Reset
     | SlopeMod ([Rational] -> [Rational])
-    deriving (Typeable)
+
```
slotThe, line 70: Some of these also still remain

slotThe (commented):
I like these quick revisions, thanks :+1: 

At some point, I think just `sed`ing it will always not _quite_ be what one wants.  Finding a good way to actually make it aware of all cases then sounds almost as bad as going through these files one-by-one and fixes the inconsistencies  manually (although I of course will not tell you what to do :>)

exorcist365:
seding is exactly what i did but im not a regex master or sed master so some things turned out fucked up, it didnt work when i put \n in the replace expression.

exorcist365:
I will fix these in a minute.

XMonad/Actions/CycleWorkspaceByScreen.hs
lua-latex: automation is bad just manually compile

slotThe:
Thank you for the contribution!  I just fixed a few outliers that were still in there, squashed everything down and gave the commit message some more love

exorcist365:
No, I thank you.

Issue #416, opened 2020-12-03
Leakage of swallowed windows when X.H.WindowSwallowing used with X.L.TrackFloating + X.H.EwmhDesktops.ewmhFullScreen 


PRESFIL:
### Problem Description

If you run the `mpv video_file` in the terminal and the terminal is swallowed by `mpv` + `mpv` will expand to full screen with **EWMH**, the terminal will not be restored after `mpv` is finished.

It is also somehow affected by `X.H.DynamicLog` (possibly due to *struts*).

### Configuration File (minimized)

```haskell
import XMonad
import XMonad.Hooks.DynamicLog
import XMonad.Hooks.EwmhDesktops
import XMonad.Hooks.WindowSwallowing
import XMonad.Layout.TrackFloating

myLayouts = trackFloating $ layoutHook def

mySwallowHook = swallowEventHook ( className =? "Alacritty" <||> className =? "Termite" ) ( return True )

main = do
  xmonad =<< xmobar (ewmhFullscreen (ewmh def
    { modMask = mod4Mask
    , layoutHook = myLayouts
    , handleEventHook = mySwallowHook <+> ewmhDesktopsEventHook <+> handleEventHook def
    }))
```

My current terminal is `alacritty`. On `termite` issue also reproducible.
[xmonad](https://aur.archlinux.org/packages/xmonad-git/) version: `v0.15.r25.g5be975b-1`
[xmonad-contrib](https://aur.archlinux.org/packages/xmonad-contrib-git/) version: `v0.16.r197.g6381a468-1`

I'm using fresh `-git` version of packages.

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

liskin:
@elkowar can you please take a look at this one? Thanks.

elkowar:
Sure, I'll take a look!

PRESFIL:
Thanks! I'm experiencing this behavior not only with `mpv`, if it's important.

I will wait very carefully.

PRESFIL:
I think I confused `mpv` with `ffplay`. Externally, they are both like a window without buttons :sweat_smile: .
To play it, you need to allow the video to be played to the end.

PRESFIL:
This also happens sometimes with `mpv'.

slotThe:
[...]

PR #562, opened 2021-06-18
Window sublayouting

4caraml:
### Description

behind them.

### Checklist

_Originally proposed by @liskin in https://github.com/xmonad/xmonad-contrib/issues/416#issuecomment-777400194_ and framed as a (small) project for ZuriHac 2021.

It is probably a good idea to rename the module to something else, as it does not only do *swallowing* in case of the added feature.

### DIFF
--- a/XMonad/Hooks/WindowSwallowing.hs
+++ b/XMonad/Hooks/WindowSwallowing.hs
@@ -40,12 +40,13 @@
 module XMonad.Hooks.WindowSwallowing
   ( -- * Usage
     -- $usage
-    swallowEventHook
+    swallowEventHook, swallowEventHookSub
   )
 where
 import           XMonad
 import           XMonad.Prelude
 import qualified XMonad.StackSet               as W
+import           XMonad.Layout.SubLayouts
 import qualified XMonad.Util.ExtensibleState   as XS
 import           XMonad.Util.WindowProperties
 import           XMonad.Util.Run                ( runProcessWithInput )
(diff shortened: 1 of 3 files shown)


XMonad/Hooks/WindowSwallowing.hs
```diff
@@ -60,9 +61,54 @@ import qualified Data.Map.Strict               as M
 --
 -- > myHandleEventHook = swallowEventHook (className =? "Alacritty" <||> className =? "Termite") (return True)
 --
+-- The variant 'swallowEventHookSub' can be used if a layout from -- "XMonad.Layouts.SubLayouts" is used;
```
slotThe, line 64:
```suggestion
-- The variant 'swallowEventHookSub' can be used if a layout from "XMonad.Layouts.SubLayouts" is used;
```

XMonad/Hooks/WindowSwallowing.hs
```diff
@@ -60,9 +61,54 @@ import qualified Data.Map.Strict               as M
...
+                --   Set this to @return True@ to run swallowing for every parent.
+  -> Query Bool -- ^ query the child window has to match for window swallowing to occur.
+                --   Set this to @return True@ to run swallowing for every child
+  -> Event      -- ^ The event to handle.
+  -> X All
+swallowEventHookSub parentQ childQ event =
+  All True <$ case event of
+    -- This is called right before a window gets opened. We intercept that
+    -- call to possibly open the window ourselves, swapping out
+    -- it's parent processes window for the new window in the stack.
```
slotThe, line 107: I would move this to the top-level docs of `handleMapRequestEvent`, so we can get rid of the two identical comments here and in `swallowEventHook`

XMonad/Hooks/WindowSwallowing.hs
```diff
@@ -60,9 +61,54 @@ import qualified Data.Map.Strict               as M
 --
 -- > myHandleEventHook = swallowEventHook (className =? "Alacritty" <||> className =? "Termite") (return True)
 --
+-- The variant 'swallowEventHookSub' can be used if a layout from -- "XMonad.Layouts.SubLayouts" is used;
+-- instead of swallowing the window it will merge the child window with the parent. (this does not work with floating windows)
+--
 -- For more information on editing your handleEventHook and key bindings,
 -- see "XMonad.Doc.Extending".
 
+-- | run @x@ iff both parent- and child queries match and the child is a child by PID
```
slotThe, line 70: `x` seems like a weird name for a function

XMonad/Hooks/WindowSwallowing.hs
```diff
@@ -60,9 +61,54 @@ import qualified Data.Map.Strict               as M
...
+  -> Event      -- ^ The event to handle.
+  -> X All
+swallowEventHookSub parentQ childQ event =
+  All True <$ case event of
+    -- This is called right before a window gets opened. We intercept that
+    -- call to possibly open the window ourselves, swapping out
+    -- it's parent processes window for the new window in the stack.
+    MapRequestEvent{ev_window=childWindow} ->
+      handleMapRequestEvent parentQ childQ childWindow $ \parentWindow -> do
+        manage childWindow >> sendMessage (Merge parentWindow childWindow)
```
slotThe, line 110: This `do` is unnecessary here.  You can either write 

```suggestion
      handleMapRequestEvent parentQ childQ childWindow $ \parentWindow ->
        manage childWindow >> sendMessage (Merge parentWindow childWindow)
```

or 

```suggestion
      handleMapRequestEvent parentQ childQ childWindow $ \parentWindow -> do
        manage childWindow
        sendMessage (Merge parentWindow childWindow)
```

slotThe (commented):
Thank you for the contribution :tada:  Using sublayouts for this is much easier than I thought it would be!

There is also `CHANGES.md`, there essentially the same change as in X.D.Extending is needed

liskin:
The idea of invoking code after managing the window may be useful in general: https://github.com/xmonad/xmonad-contrib/issues/372#issuecomment-864372552

slotThe:
@4caraml friendly ping :)

slotThe:
@liskin I think this is ready to go now

XMonad/Hooks/WindowSwallowing.hs
```diff
@@ -60,9 +61,57 @@ import qualified Data.Map.Strict               as M
...
+      -- read the windows _NET_WM_PID properties
+      childWindowPid  <- getProp32s "_NET_WM_PID" childWindow
+      parentWindowPid <- getProp32s "_NET_WM_PID" parentWindow
+      case (parentWindowPid, childWindowPid) of
+        (Just (parentPid : _), Just (childPid : _)) -> do
+          -- check if the new window is a child process of the last focused window
+          -- using the process ids.
+          isChild <- liftIO $ fi childPid `isChildOf` fi parentPid
+          when isChild $ do
+            query parentWindow
```
liskin, line 95: This is more of an `action` than a `query`

slotThe: Ah, good point; especially since query in this context definitely means something different.  I'll fix that and then merge

PR #563, opened 2021-06-21
added function shiftWin to PureX

ss7m:
### Description

Added the function `shiftWin` to `Xmonad.Util.PureX` as a refresh-tracking version of `W.shiftWin`.
This was a gap in the currently privded interface that I needed. 

### Checklist

        manually, ...) and concluded:

This PR only adds the function `shiftWin`, so unit tests on this function would suffice. For example, if the target tag or target window don't exist, then it should return `Any False`, if the window is already on the target tag, it should return `Any False`, in all other cases it should return `Any True`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -491,6 +491,8 @@
     - Added `focusWindow` and `focusNth` which don't refresh (and thus
       possibly flicker) when they happen to be a no-op.
 
+    - Added `shiftWin` as a refresh tracking version of `W.shiftWin`.
+
   * Several `LayoutClass` instances now have an additional `Typeable`
     constraint which may break some advanced configs. The upside is that we
     can now add `Typeable` to `LayoutClass` in `XMonad.Core` and make it
--- a/XMonad/Util/PureX.hs
+++ b/XMonad/Util/PureX.hs
@@ -46,7 +46,7 @@ module XMonad.Util.PureX (
   getStack, putStack, peek,
   focusWindow, focusNth,
   view, greedyView, invisiView,
-  shift, curScreen, curWorkspace,
+  shift, shiftWin, curScreen, curWorkspace,
   curTag, curScreenId,
 ) where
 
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -483,6 +483,8 @@
     - Added `focusWindow` and `focusNth` which don't refresh (and thus
       possibly flicker) when they happen to be a no-op.
 
+    - Added `shiftWin` as a refresh tracking version of `W.shiftWin`
```
slotThe, line 486:
```suggestion
    - Added `shiftWin` as a refresh tracking version of `W.shiftWin`.
```

slotThe (approved):
LGTM!

XMonad/Util/PureX.hs
```diff
@@ -271,6 +271,14 @@ shift tag = withFocii $ \ctag fw ->
     mfw' <- peek
     return (Any $ Just fw /= mfw')
 
+-- | A refresh tracking version of @W.shiftWin@
+shiftWin :: XLike m => WorkspaceId -> Window -> m Any
+shiftWin tag w = do
+  ctag <- curTag
+  when' (tag /= ctag) $ do
+    modifyWindowSet' (W.shiftWin tag w)
+    return $ Any True
```
liskin, line 280: `W.shiftWin` can end up being a no-op if `tag` refers to a non-existent workspace or if `w` refers to a non-existent window. Not sure how big a deal this is, but the other functions in this module handle that correctly.

ss7m:
@slotThe I'll admit ignorance on how to do testing within xmonad. I'll have to look into that later today.

XMonad/Util/PureX.hs
ss7m: Latest commits fix this, and a separate logic error (the old implementation assumed that the window `w` was on the current workspace)

slotThe:
I just rebased and squashed everything down and will merge this for now.  For better or worse, none of the functions in X.U.PureX have proper tests, so it would be unfair to expect that of this one.  This is something we should definitely change though, especially for these kinds of functions that really only interact with the `StackSet`.  So if you still want to think about this I very much welcome you to!

PR #564, opened 2021-06-23
X.U.NamedScratchpad: Add logHook to auto-hide named scratchpads on focus loss

slotThe:
### Description

Adds a new logHook, nsHideOnFocusLoss, that hides the given
scratchpads when they lose focus.  This is akin to the functionality
provided by a lot of dropdown terminals.

Relevant commits:

- X.H.RefocusLast: Export withRecentsIn

  This exports the `withRecentsIn` function, as it's quite useful when
  using X.H.RefocusLast in other modules as a library.

  It is already possible (RecentsMap is fully exported) to completely
  define this function outside of the module, so we are not exposing any
  more internals than we were before.

- X.U.NamedScratchpad: Add nsHideOnFocusLoss 

### Checklist

        manually, ...) and concluded: I don't think we have any proper setup for testing logHooks do we?  As far as I can tell, this is only using the reader/state part of `X`, so it should theoretically be possible...  

### DIFF
--- a/XMonad/Hooks/RefocusLast.hs
+++ b/XMonad/Hooks/RefocusLast.hs
@@ -41,7 +41,9 @@ module XMonad.Hooks.RefocusLast (
   RecentWins(..),
   RecentsMap(..),
   RefocusLastLayoutHook(..),
-  RefocusLastToggle(..)
+  RefocusLastToggle(..),
+  -- * Library functions
+  withRecentsIn,
 ) where
 
 import XMonad
@@ -262,7 +264,7 @@ updateRecentsOn tag = withWindowSet $ \ws ->
 
 -- }}}
 
--- --< Private Utilities >-- {{{
+-- --< Utilities >-- {{{
 
 -- | Focuses the first window in the list it can find on the current workspace.
 tryFocus :: [Window] -> WindowSet -> WindowSet
--- a/XMonad/Util/NamedScratchpad.hs
+++ b/XMonad/Util/NamedScratchpad.hs
@@ -27,15 +27,17 @@ module XMonad.Util.NamedScratchpad (
   allNamedScratchpadAction,
   namedScratchpadManageHook,
   namedScratchpadFilterOutWorkspace,
-  namedScratchpadFilterOutWorkspacePP
+  namedScratchpadFilterOutWorkspacePP,
+  nsHideOnFocusLoss,
   ) where
 
 import XMonad
-import XMonad.Prelude (filterM, find, unless)
-import XMonad.Hooks.ManageHelpers (doRectFloat)
 import XMonad.Actions.DynamicWorkspaces (addHiddenWorkspace)
-import XMonad.Hooks.DynamicLog (PP, ppSort)
 import XMonad.Actions.SpawnOn (spawnHere)
+import XMonad.Hooks.DynamicLog (PP, ppSort)
+import XMonad.Hooks.ManageHelpers (doRectFloat)
+import XMonad.Hooks.RefocusLast (withRecentsIn)
+import XMonad.Prelude (filterM, find, unless, when)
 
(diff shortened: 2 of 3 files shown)


XMonad/Hooks/RefocusLast.hs
```diff
@@ -41,7 +41,9 @@ module XMonad.Hooks.RefocusLast (
   RecentWins(..),
   RecentsMap(..),
   RefocusLastLayoutHook(..),
-  RefocusLastToggle(..)
+  RefocusLastToggle(..),
+  -- * Library functions
+  withRecentsIn,
```
slotThe, line 46: I wonder whether it would be better to instead expose every utility here (they can all be recreated by what's already exported) or if we should do this on a case-by-case basis, depending on when they are needed

XMonad/Util/NamedScratchpad.hs
```diff
@@ -152,6 +154,31 @@ allNamedScratchpadAction :: NamedScratchpads
...
+nsHideOnFocusLoss :: NamedScratchpads -> X ()
+nsHideOnFocusLoss scratches = withWindowSet $ \winSet -> do
+    let cur = W.currentTag winSet
+    withRecentsIn cur () $ \lastFocus _ ->
+        when (lastFocus `elem` W.index winSet && cur /= scratchpadWorkspaceTag) $
+            whenX (isNS lastFocus) $
+                shiftToNSP (W.workspaces winSet) ($ lastFocus)
+  where
+    isNS :: Window -> X Bool
+    isNS w = or <$> traverse ((`runQuery` w) . query) scratches
```
TheMC47, line 180: Hmm, can this be useful somewhere else? It looks handy

TheMC47 (approved):
This is a cool addition, thanks! I haven't tried it out (yet) but it looks fine
LGTM

XMonad/Util/NamedScratchpad.hs
geekosaur: There's also https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/src/XMonad.Util.Loggers.NamedScratchpad.html#isSp, but it's not quite the same thing and arguably misnamed as it does more than just identifying named scratchpads.

XMonad/Hooks/RefocusLast.hs
TheMC47: I'm not sure why `RecentsMap` was exported in the first place, but your argument makes sense (although if I could, I'd un-export `RecentsMap`)

slotThe: Yes that would be the optimal solution but, alas, it has already made its way to a release.  We could think about deprecating the export (is that possible?  If not I guess a simple deprecation of the function that informs users about this would also work) but that's not something I necessarily want to do in this pr

XMonad/Util/NamedScratchpad.hs
slotThe: I mean, we _could_ make this a top-level function (I haven't actually checked if this could be useful anywhere else, but I imagine we'd find places).  Then, however I propose we go the extra mile and rewrite this as a proper 

``` haskell
  isNS scratches w = anyM ((`runQuery` w) . query) scratches
``` 

for a proper definition of `anyM`.  Reason being that `or <$> traverse` (as opposed to `or $ map`) does not actually short-circuit when a match has been found (yay for applicative effects).  There are "correct" definitions of `orM` and `anyM` that do this, but they have to be defined via some custom fold and are thus more than one line.

TheMC47: Interesting. You don't have to delay merging this though, we can take care of it later

slotThe:
Okay I just added a few more top-module docs.  Let's see whether we actually remember to think about both of these points later :)

PR #568, opened 2021-06-29
Change EWMH windows ordering to be closer to the spec.

### LINKED ISSUE #567: EWMH windows ordering
Currently xmonad sets ewmh windows with focused window first, and sometimes it's not very convenient. For instance, it causes taffybar to redraw windows icons in its workspaces widget each time focus is changed, and focused window will be displayed first. It's quite confusing because you can't follow focus visually (taffybar highlights focused window).

What was the reasoning of this behavior? Could it be changed to just use `W.integrate'` [here](https://github.com/xmonad/xmonad-contrib/blob/5fffe03e063aef25881230918cc249f9143f7d80/XMonad/Hooks/EwmhDesktops.hs#L175)?

kurnevsky:
### Description

Fixes #567

### Checklist

        manually, ...) and concluded: tested manually, have no idea how it can be covered with tests.

  - [x] I updated the `XMonad.Doc.Extending` file (not appropriate)

### DIFF
--- a/XMonad/Hooks/EwmhDesktops.hs
+++ b/XMonad/Hooks/EwmhDesktops.hs
@@ -124,6 +124,14 @@ instance ExtensionClass ClientList where
     initialValue = ClientList [none]
 
 -- |
+-- Cached stacking client list (e.g. @_NET_CLIENT_LIST_STACKING@).
+newtype ClientListStacking = ClientListStacking [Window]
+                           deriving Eq
+
+instance ExtensionClass ClientListStacking where
+    initialValue = ClientListStacking [none]
+
+-- |
 -- Cached current desktop (e.g. @_NET_CURRENT_DESKTOP@).
 newtype CurrentDesktop = CurrentDesktop Int
                        deriving Eq
@@ -171,9 +179,16 @@ ewmhDesktopsLogHookCustom t = withWindowSet $ \s -> do
         setNumberOfDesktops (length desktopNames)
         setDesktopNames desktopNames
 
-    -- Set client list; all windows, with focused windows last
-    let clientList = nub . concatMap (maybe [] (\(W.Stack x l r) -> reverse l ++ r ++ [x]) . W.stack) $ ws
+    -- Set client list which should be sorted by window age. We just
+    -- guess that StackSet contains windows list in this order which
+    -- isn't true but at least gives consistency with windows cycling
+    let clientList = nub . concatMap (W.integrate' . W.stack) $ ws
     whenChanged (ClientList clientList) $ setClientList clientList
+
+    -- Set stacking client list which should have bottom-to-top
+    -- stacking order, i.e. focused window should be last
+    let clientListStacking = nub . concatMap (maybe [] (\(W.Stack x l r) -> reverse l ++ r ++ [x]) . W.stack) $ ws
+    whenChanged (ClientListStacking clientListStacking) $ setClientListStacking clientListStacking
 
     -- Remap the current workspace to handle any renames that f might be doing.
     let maybeCurrent' = W.tag <$> listToMaybe (t [W.workspace $ W.current s])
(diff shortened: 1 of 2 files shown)


XMonad/Hooks/EwmhDesktops.hs
```diff
@@ -171,10 +179,17 @@ ewmhDesktopsLogHookCustom t = withWindowSet $ \s -> do
         setNumberOfDesktops (length desktopNames)
         setDesktopNames desktopNames
 
-    -- Set client list; all windows, with focused windows last
-    let clientList = nub . concatMap (maybe [] (\(W.Stack x l r) -> reverse l ++ r ++ [x]) . W.stack) $ ws
+    -- Set client list which should be sorted by window age. We just
+    -- guess that StackSet contains windows list in this order which
+    -- isn't true but at least gives consistency with windows cycling
+    let clientList = nub . concatMap (reverse . W.integrate' . W.stack) $ ws
```
slotThe, line 185: Why is there a `reverse` here?

XMonad/Hooks/EwmhDesktops.hs
```diff
@@ -171,10 +179,17 @@ ewmhDesktopsLogHookCustom t = withWindowSet $ \s -> do
...
     whenChanged (ClientList clientList) $ setClientList clientList
 
+    -- Set stacking client list which should have bottom-to-top
+    -- stacking order, i.e. focused window should be last
+    let clientListStacking = nub . concatMap (maybe [] (\(W.Stack x l r) -> l ++ reverse r ++ [x]) . W.stack) $ ws
```
slotThe, line 190: What's the rationale for reversing the `r` here?

slotThe (commented):
I have no objections to the /per se/, but would like others to chime in before I do anything hasty :)

XMonad/Hooks/EwmhDesktops.hs
kurnevsky: Currently we have `reverse l ++ r ++ [x]` but it seemed to me incorrect. We should put there windows in ascending stacking and if we have for instance windows `1 2 [3] 4 5`, currently they will be reordered to `2 1 4 5 3` but it doesn't seem correct because 5 doesn't follow 3. After my fix they will be ordered as `1 2 5 4 3` - here after 3 we will have windows `4 5` which are next windows that we get with `focusDown`, and then windows `2 1` which we will get with `focusUp`.

kurnevsky: Hm, I think I misunderstood it a bit. StackSet stores `left` part in a reversed order so initially this reverse just restored windows ordering. So to follow my logic I should reverse left part as well. Or I can just leave it as it was before :)

XMonad/Hooks/EwmhDesktops.hs
kurnevsky: If we don't have reverse here taffybar will display windows icons in a reversed order, but I'm not sure why. I'm fine if we remove reverse here because ordering can be easily adjusted on taffybar side,

kurnevsky: Removed reverse as it seems this way it'll be closer to specs.

XMonad/Hooks/EwmhDesktops.hs
kurnevsky: Changed it to what it was before - nobody complained about it and I don't have good explanation why it should be changed :)

XMonad/Hooks/EwmhDesktops.hs
```diff
@@ -171,10 +179,17 @@ ewmhDesktopsLogHookCustom t = withWindowSet $ \s -> do
...
-    -- Set client list; all windows, with focused windows last
-    let clientList = nub . concatMap (maybe [] (\(W.Stack x l r) -> reverse l ++ r ++ [x]) . W.stack) $ ws
+    -- Set client list which should be sorted by window age. We just
+    -- guess that StackSet contains windows list in this order which
+    -- isn't true but at least gives consistency with windows cycling
+    let clientList = nub . concatMap (W.integrate' . W.stack) $ ws
     whenChanged (ClientList clientList) $ setClientList clientList
 
+    -- Set stacking client list which should have bottom-to-top
+    -- stacking order, i.e. focused window should be last
```
liskin, line 189: Note that focused window being last is not universally true. It's done by LayoutModifiers such as Magnifier that need it, and also by some layouts like MultiColumns (for reasons unknown), but most other layouts simply put the master window to the top.

kurnevsky: Yeah, but this comment means that in the result list focused window should be the last. And we assume here that focused window is the window that's selected in a `Stack` struct. It might be not always true, but it's what we have now, and I don't have a good example how it breaks something. So I left it as is because fixing it would require much more effort with no proper goal to reach :)

liskin: Yeah, no action required, just wanted to note that in case it brings any forgotten memories out of anyone. :-)

slotThe: @liskin so I assume we can merge this then?  More time to see whether people complain :)

liskin: Yeah, let's do it.

slotThe:
Thanks!

PR #571, opened 2021-07-15
fix compilation warning in DebugEvents.hs

Rogach:
### Description

This fixes the compilation warning:

```
XMonad/Hooks/DebugEvents.hs:35:1: warning: [-Wunused-imports]
    The import of ‘Control.Monad.Fail’ is redundant
      except perhaps to import instances from ‘Control.Monad.Fail’
    To import instances alone, use: import Control.Monad.Fail()
   |
35 | import           Control.Monad.Fail
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Context for the warning: https://wiki.haskell.org/MonadFail_Proposal

### Checklist

  - [x] I've considered how to best test these changes (property, unit, manually, ...) and concluded: best to test this manually. We could enable `-Werror`, but that's too big of a change.

  - [x] I **didn't** update the `CHANGES.md` file because it's a very small internal change.

  - [x] I **didn't** update the `XMonad.Doc.Extending` file.

### DIFF
--- a/XMonad/Hooks/DebugEvents.hs
+++ b/XMonad/Hooks/DebugEvents.hs
@@ -32,7 +32,9 @@ import           XMonad.Util.DebugWindow                     (debugWindow)
 -- import           Graphics.X11.Xlib.Extras.GetAtomName        (getAtomName)
 
 import           Control.Exception                    as E
+#if __GLASGOW_HASKELL__ <= 806
 import           Control.Monad.Fail
+#endif
 import           Control.Monad.State
 import           Control.Monad.Reader
 import           Codec.Binary.UTF8.String


XMonad/Hooks/DebugEvents.hs
```diff
@@ -32,7 +32,9 @@ import           XMonad.Util.DebugWindow                     (debugWindow)
 -- import           Graphics.X11.Xlib.Extras.GetAtomName        (getAtomName)
 
 import           Control.Exception                    as E
+#if __GLASGOW_HASKELL__ <= 806
 import           Control.Monad.Fail
+#endif
```
slotThe, line 37: I was going to say enabling `CPP` just for this is definitely not worth it imo, but the module apparently already has this enabled...  However, we already run [(parts of) the CI](https://github.com/xmonad/xmonad-contrib/blob/master/.github/workflows/haskell-ci.yml#L156) with `-Werror`, so I'm still not convinced it's worth it here.

On a completely different note, I'm not sure `Decoder` needs that CPP at all and so it coul probably be removed completely from DebugEvents.

Rogach: It's my understanding that if we enable `-Werror` then the build will start failing on GHC 8.8 and 8.10, so we need some way of sliencing the warning (anyway, keeping lots of warnings around is usually suboptimal).

What is the danger of enabling CPP to get rid of the warning?

liskin: We do enable `-Werror` already but we explicitly disable it for `unused-imports` via `-Wwarn=unused-imports`:
https://github.com/xmonad/xmonad-contrib/blob/cfc793e94f6df8d3702a902833ec2ff1c2206c89/xmonad-contrib.cabal#L66-L73
because there are differences in exports between versions of base fairly often, and it's a maintenance burden to `#if` around all of them all the time.

Rogach: I understand the rationale, but currently this warning stays alone as a sore foot in a sea of flawless compilation :)

What's the verdict - should I close the PR?

slotThe: > What's the verdict - should I close the PR?

I would say that CPP is ugly enough to not make me personally want to merge this.  If you still want to use this PR, then I'd say removing CPP completely from DebugEvents sounds like a good idea!

Rogach: Re-using this PR for entirely different purpose will be a bit untidy, it's better to close this one then.

slotThe: Well, I think what counts in the end are the commits that make it into the git history and those don't really care which GitHub PR they originally came from.  But fair enough

PR #573, opened 2021-07-15
rewrite GridSelect.stringToRatio to use randomR (fixes #572)

### LINKED ISSUE #572: XMonad.Actions.GridSelect: colorRangeFromClassName gives incorrect results
### Problem Description

Due to how `stringToRatio` is implemented `colorRangeFromClassName` is giving incorrect results (tile colors are wrong).

### Steps to Reproduce

Compile xmonad with configuration file below, restart xmonad, press "win-shift-G".

Observe that all inactive window tiles are black:
selector-bad

Compare with how it should actually look:
selector-good

### Configuration File

```haskell
{-# OPTIONS_GHC -fno-warn-type-defaults #-}
import XMonad
import XMonad.Actions.GridSelect
import XMonad.Config.Xfce

import Text.Printf
import Data.Word (Word8)
import System.Random (mkStdGen, genRange, next, uniformR)


myKeys :: XConfig Layout -> M.Map (ButtonMask, KeySym) (X ())
myKeys (XConfig {modMask = modm}) =
  let greenColorizer =
        colorRangeFromClassName (0x00, 0x88, 0x00) (0x00, 0xff, 0x00) (0xff, 0x00, 0x00) maxBound maxBound
  in M.fromList $
     [ ((modm .|. shiftMask, xK_g), goToSelected $ buildDefaultGSConfig greenColorizer) ]

main :: IO ()
main = do
  xmonad $ xfceConfig
              { modMask = mod4Mask
              , keys = myKeys <+> keys xfceConfig
              }
```

### Checklist

  - I tested my configuration
    - [x] With `xmonad` commit af354f7
    - [x] With `xmonad-contrib` commit da2fb360
    - [x] `random` package version `1.2.0`

Rogach:
### Description

Due to differences between random-1.1 and random-1.2, on newer systems stringToRatio returns numbers outside [0, 1] range, which breaks colorRangeFromClassName colorizers.

This commit fixes the issue by using randomR to directly generate the random number.

Also this fixes the compilation warning (genRange and next are deprecated in random-1.2).

Instead of using sum of charactes I've elected use a simple hash - this will result in a bit less collisions (simple sums of characters collide quite often).

  - [x] I've considered how to best test these changes (property, unit, manually, ...) and added a test for `stringToRatio` function.

  - [x] I updated the `CHANGES.md` file.

  - [x] I *didn't* update the `XMonad.Doc.Extending` file (not appropriate)

### DIFF
--- a/XMonad/Actions/GridSelect.hs
+++ b/XMonad/Actions/GridSelect.hs
@@ -47,6 +47,7 @@ module XMonad.Actions.GridSelect (
     fromClassName,
     stringColorizer,
     colorRangeFromClassName,
+    stringToRatio,
 
     -- * Navigation Mode assembly
     TwoD,
@@ -93,7 +94,7 @@ import XMonad.Layout.Decoration
 import XMonad.Util.NamedWindows
 import XMonad.Actions.WindowBringer (bringWindow)
 import Text.Printf
-import System.Random (mkStdGen, genRange, next)
+import System.Random (mkStdGen, randomR)
 import Data.Word (Word8)
 
 -- $usage
@@ -628,15 +629,12 @@ mix (r1, g1, b1) (r2, g2, b2) r = (mix' r1 r2, mix' g1 g2, mix' b1 b2)
 
 -- | Generates a Double from a string, trying to
 -- achieve a random distribution.
--- We create a random seed from the sum of all characters
+-- We create a random seed from the hash of all characters
 -- in the string, and use it to generate a ratio between 0 and 1
 stringToRatio :: String -> Double
 stringToRatio "" = 0
-stringToRatio s = let gen = mkStdGen $ sum $ map fromEnum s
-                      range = (\(a, b) -> b - a) $ genRange gen
-                      randomInt = foldr1 combine $ replicate 20 next
-                      combine f1 f2 g = let (_, g') = f1 g in f2 g'
-                  in fi (fst $ randomInt gen) / fi range
+stringToRatio s = let gen = mkStdGen $ foldl' (\t c -> t * 31 + fromEnum c) 0 s
+                  in fst $ randomR (0, 1) gen
 
 -- | Brings up a 2D grid of elements in the center of the screen, and one can
 -- select an element with cursors keys. The selected element is returned.
--- a/tests/GridSelect.hs
+++ b/tests/GridSelect.hs
@@ -1,0 +1,15 @@
+module GridSelect where
+
+import Test.Hspec
+import Test.Hspec.QuickCheck
+
+import XMonad.Actions.GridSelect
+
+spec :: Spec
+spec = do
+  prop "prop_stringToRatio_valuesInRange"  prop_stringToRatio_valuesInRange
+
+prop_stringToRatio_valuesInRange :: String -> Bool
+prop_stringToRatio_valuesInRange s =
+  let r = stringToRatio s
+  in r >= 0 && r <= 1
--- a/xmonad-contrib.cabal
+++ b/xmonad-contrib.cabal
@@ -375,6 +375,8 @@ test-suite tests
   type:           exitcode-stdio-1.0
   main-is:        Main.hs
   other-modules:  CycleRecentWS
+                  ExtensibleConf
+                  GridSelect
                   Instances
                   ManageDocks
                   NoBorders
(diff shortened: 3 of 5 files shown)


TheMC47:
Thank you for your contribution :tada: 

> Also this fixes the compilation warning (genRange and next are deprecated in random-1.2).

Where can you see the logs, exactly? We tried to fix any warnings, and as far as I'm aware, we did (at least I don't see any while compiling locally or in the CI)

Rogach:
@TheMC47 When compiling locally on my system (Arch Linux). Here's the link to the deprecated method: https://hackage.haskell.org/package/random-1.2.0/docs/System-Random.html#v:next

And here's the error message as seen on my system (building using `runhaskell Setup.lhs build`):

```
XMonad/Actions/GridSelect.hs:636:52: warning: [-Wdeprecations]
    In the use of ‘genRange’
    (imported from System.Random, but defined in System.Random.Internal):
    Deprecated: "No longer used"
    |
636 |                       range = (\(a, b) -> b - a) $ genRange gen
    |                                                    ^^^^^^^^

XMonad/Actions/GridSelect.hs:637:65: warning: [-Wdeprecations]
    In the use of ‘next’
    (imported from System.Random, but defined in System.Random.Internal):
    Deprecated: "No longer used"
    |
637 |                       randomInt = foldr1 combine $ replicate 20 next
    |                                                                 ^^^^
```

xmonad-contrib.cabal
```diff
@@ -383,31 +385,41 @@ test-suite tests
...
                   XMonad.Actions.FocusNth
+                  XMonad.Actions.GridSelect
                   XMonad.Actions.PhysicalScreens
                   XMonad.Actions.RotateSome
                   XMonad.Actions.SwapWorkspaces
                   XMonad.Actions.TagWindows
+                  XMonad.Actions.WindowBringer
                   XMonad.Hooks.ManageDocks
+                  XMonad.Hooks.ManageHelpers
+                  XMonad.Hooks.UrgencyHook
```
TheMC47, line 399: Any reason all these modules were added?

TheMC47 (commented):
LGTM, I don't understand the changes to the xmonad-contrib.cabal though. Is this because of warnings? If so, it might be better to make them a separate commit

xmonad-contrib.cabal
Rogach: That's to silence the stack warnings - I simply copied the list it suggested.

Rogach:
@TheMC47 Separate commit in a separate pull request, or just another commit in this PR?

xmonad-contrib.cabal
TheMC47: Hmmm, I can't see the warnings in the CI though. I'm not familiar with the build though, so maybe wait for someone else with more experience to come along

Rogach: Yes, there are no warnings in the CI - because stack pulls the older version of `random` package (1.1 specifically).
I think this happens because stack is configured to use LTS Haskell 17.14 (`resolver: lts-17.14` line in stack-master.yaml), and this LTS includes random version 1.1, not 1.2.

TheMC47:
> @TheMC47 Separate commit in a separate pull request, or just another commit in this PR?

I'd say another PR, but I couldn't reproduce the warnings with a simple `stack test`. So maybe (again) wait for someone to chime in.

Ping @slotThe you seemed to have talked about this in the mailing list, so your feedback is appreciated! (I can't see your mails for some reason, they're also not in the archives)

Rogach:
> I'd say another PR, but I couldn't reproduce the warnings with a simple stack test

Did you run the test on this PR, but with the cabal changes reverted? If you just run this on the master branch then there shouldn't be any warnings - the warnings appeared due to the fact that I added an additional test file which pulled in lots of other modules (that's why I included the changes into this PR, by the way - the change seems conceptually related, to fix the warning in the same commit that would have introduced it).

xmonad-contrib.cabal
TheMC47: Nevermind, I understand now

TheMC47:
Yes I can confirm that's the case. Nevermind my comment then. Awesome!

liskin (approved):

tests/GridSelect.hs
```diff
@@ -0,0 +1,12 @@
...
+import Test.Hspec
+
+import XMonad.Actions.GridSelect
+
+spec :: Spec
+spec = do
+  describe "stringToRatio" $ do
+    it "returns values in (0, 1) range" $ do
+      stringToRatio "deadbeef" `shouldSatisfy` (>= 0)
+      stringToRatio "deadbeef" `shouldSatisfy` (<= 1)
```
slotThe, line 12: Since there is something like `arbitraryPrintableChar` in QuickCheck it might be nice to property test this instead?

XMonad/Actions/GridSelect.hs
```diff
@@ -628,15 +629,12 @@ mix (r1, g1, b1) (r2, g2, b2) r = (mix' r1 r2, mix' g1 g2, mix' b1 b2)
...
+-- We create a random seed from the hash of all characters
 -- in the string, and use it to generate a ratio between 0 and 1
 stringToRatio :: String -> Double
 stringToRatio "" = 0
-stringToRatio s = let gen = mkStdGen $ sum $ map fromEnum s
-                      range = (\(a, b) -> b - a) $ genRange gen
-                      randomInt = foldr1 combine $ replicate 20 next
-                      combine f1 f2 g = let (_, g') = f1 g in f2 g'
-                  in fi (fst $ randomInt gen) / fi range
+stringToRatio s = let gen = mkStdGen $ foldl (\t c -> t * 31 + fromEnum c) 0 s
```
slotThe, line 636:
```suggestion
stringToRatio s = let gen = mkStdGen $ foldl' (\t c -> t * 31 + fromEnum c) 0 s
```

slotThe (approved):

tests/GridSelect.hs
Rogach: I used default String instance, not `arbitraryPrintableChar` - is that okay?

slotThe: That's totally okay as well

slotThe:
Thanks!

PR #574, opened 2021-07-16
X.L.NoBorders: Add OnlyFloat

slotThe:
### Description

This adds an `OnlyFloat` data constructor to `Ambiguity`; I've found this curiously missing with the current setup of things. 

### Checklist

        manually, ...) and concluded: A property test is included; we sadly can't test for arbitrary `WindowSet`s because there's no arbitrary instance for this.  However, the hand-crafted dual-head setup seems decent enough for this purpose.

### DIFF
--- a/XMonad/Layout/NoBorders.hs
+++ b/XMonad/Layout/NoBorders.hs
@@ -279,6 +279,8 @@ instance SetsAmbiguous Ambiguity where
                         in lr == wr1 && (not . or) vu
                     OnlyLayoutFloat ->
                         lr == wr1
+                    OnlyFloat ->
+                        True
                     _ ->
                         wr1 `R.supersetOf` sr
                 return w1
@@ -288,6 +290,7 @@ instance SetsAmbiguous Ambiguity where
               | Screen <- amb = [w]
               | OnlyScreenFloat <- amb = []
               | OnlyLayoutFloat <- amb = []
+              | OnlyFloat <- amb = []
               | OnlyLayoutFloatBelow <- amb = []
               | OtherIndicated <- amb
               , let nonF = map integrate $ W.current wset : W.visible wset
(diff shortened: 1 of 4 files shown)


XMonad/Layout/NoBorders.hs
```diff
@@ -326,6 +329,9 @@ data Ambiguity
         -- ^ Focus in an empty screen does not count as ambiguous.
     | OtherIndicated
         -- ^ No borders on full when all other screens have borders.
+    | OnlyFloat
```
slotThe, line 332: I'm not sure how far down this should be in terms of ambiguity, but it feels rather ambiguous and so it was placed here.  Other opinions welcome though

tests/Instances.hs
```diff
@@ -122,6 +122,8 @@ instance Arbitrary RectC where
 instance Arbitrary Rectangle where
   arbitrary = Rectangle <$> arbitrary <*> arbitrary <*> arbitrary <*> arbitrary
 
+instance Arbitrary RationalRect where
+  arbitrary = RationalRect <$> arbitrary <*> arbitrary <*> arbitrary <*> arbitrary
```
slotThe, line 126: This seems general enough to be in the Instances module (frankly, I'm surprised no one else has needed it up to this point) and so it is there :)

geekosaur: Does that somehow limit the values it can take, or are we also testing out-of-range `RationalRect`s? Because I'm pretty sure the type doesn't and can't enforce the range limits.

slotThe: We're also testing out ot range arguments here.  Since the type doesn't enforce these limits, I've seen users not care much about them either (stretching a floating window over three monitors is easily done with the appropriate `RationalRect`, for example). 

It would be an easy fix to only accept value between 0 and 1 here; not sure what's best

slotThe: Okay I decided to restrict `RationalRects` to be at least between 0 and 1.  There is no coherence between these values, so this still leaves room for windows to go outside of the rectangle.  However, that's something that can (and will) happen quite regularly so I don't think it's that much of an issue

slotThe:
I will merge this for now; I think that restricting all of the fields to be in `[0, 1]` seems like the most sensible option and the rest doesn't seem to be very contentious

PR #575, opened 2021-07-21
trap `(0,0)` keybinding function and use default binding

geekosaur:
`(0,0)` typically comes from using `def` as the keybinding function,
which is resolved as `const (0,0)` via the function and tuple instances
for `def`.

In older versions of xmonad, this would bind every unbound key to
toggle struts. This has been fixed, but we should still try to
honor the user's intent in using `def` there.

A user who intends no keybinding function should use `withSB`
instead, especially if xmonad doesn't have aforementioned fix yet.

### Checklist

        manually, ...) and concluded: this involves xmonad core and the X
        server, so probably manually

### DIFF
--- a/XMonad/Hooks/DynamicLog.hs
+++ b/XMonad/Hooks/DynamicLog.hs
@@ -156,7 +156,8 @@ import XMonad.Hooks.StatusBar
 -- You should use this function only when the default 'dzen' function does not
 -- serve your purpose.
 --
-dzenWithFlags :: LayoutClass l Window
+dzenWithFlags :: (LayoutClass l Window
+                 ,Read (l Window))
               => String     -- ^ Flags to give to @dzen@
               -> XConfig l  -- ^ The base config
               -> IO (XConfig (ModifiedLayout AvoidStruts l))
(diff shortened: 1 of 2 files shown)


geekosaur:
I requested a second set of eyes on this because of the extra constraints induced by wrapping the `layoutHook` in the existential, required to be able to run the keybinding function. I don't know if this will break anything else, or if we care since this whole thing is effectively new.

TheMC47 (commented):
Good catch! It looks fine at a first glance, I'll try to test it tomorrow (I don't think anything would break though)

geekosaur:
This has been on my to-do list since I tracked down what was up with https://github.com/xmonad/xmonad/commit/383ffb7, as the other half of the fix for the (only submitted in IRC) issue where someone was seeing that behavior with unbound keys. It's kind of an ugly hack, but so is defining an overlapping instance for our keybinding function.

TheMC47:
One could say it's the lesser of two evils... As you said, they should use `withSB` instead.

slotThe (approved):
Users shouldn't be able to compile their configs without every layout having a proper `Read (l Window)` instance, so I don't see where this would cause any problems

TheMC47 (approved):

TheMC47:
Nothing is likely to break, I took it out for a spin and it works fine. Thanks!

PR #576, opened 2021-07-22
X.H.StatusBar: Simplify the fix for `def` as `k`

liskin:
Would this also work?

Fixes: 58e2b803a478 ("fix: handle `const (0,0)` from passing `def` as keybinding action")


geekosaur:
Yes, that looks like it should also work.

geekosaur (approved):
Looks good to me!

PR #580, opened 2021-07-31
Add a flake.nix file

IvanMalison:
### Description

behind them.

### Checklist

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,25 @@
+name: Nix
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Nix Flake - Linux
+    permissions:
+      contents: read
+    steps:
+      - name: Install Nix
+        uses: cachix/install-nix-action@v13
+        with:
+          install_url: https://nixos-nix-install-tests.cachix.org/serve/i6laym9jw3wg9mw6ncyrk6gjx4l34vvx/install
+          install_options: '--tarball-url-prefix https://nixos-nix-install-tests.cachix.org/serve'
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+            access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
+      - name: Clone project
+        uses: actions/checkout@v2
+      - name: Build
+        run: nix build --print-build-logs
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -22,4 +22,7 @@ jobs:
       - name: Clone project
         uses: actions/checkout@v2
       - name: Build
-        run: nix build --print-build-logs
+        # "nix build" builds with full optimization and includes a profiling
+        # build, so just the build of xmonad-contrib itself takes 3 minutes.
+        # As a workaround, we invoke cabal manually here.
+        run: nix develop -c cabal v2-build -O0 -j
--- a/flake.nix
+++ b/flake.nix
@@ -23,6 +23,7 @@
   rec {
     devShell = pkgs.haskellPackages.shellFor {
       packages = p: [ p.xmonad-contrib ];
+      nativeBuildInputs = [ pkgs.cabal-install ];
     };
     defaultPackage = pkgs.haskellPackages.xmonad-contrib;
   }) // { inherit overlay overlays; } ;
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,29 @@
+# This file is maintained by @IvanMalison (github)
+{
+  inputs = {
+    flake-utils.url = github:numtide/flake-utils;
+    git-ignore-nix.url = github:IvanMalison/gitignore.nix/master;
+    xmonad.url = github:xmonad/xmonad;
+  };
+  outputs = { self, flake-utils, nixpkgs, git-ignore-nix, xmonad }:
+  let
+    overlay = final: prev: {
+      haskellPackages = prev.haskellPackages.override (old: {
+        overrides = prev.lib.composeExtensions (old.overrides or (_: _: {}))
+        (hself: hsuper: {
+          xmonad-contrib =
+            hself.callCabal2nix "xmonad-contrib" (git-ignore-nix.gitIgnoreSource ./.) { };
+        });
+      });
+    };
+    overlays = xmonad.overlays ++ [ overlay ];
+  in flake-utils.lib.eachDefaultSystem (system:
+  let pkgs = import nixpkgs { inherit system overlays; };
+  in
+  rec {
+    devShell = pkgs.haskellPackages.shellFor {
+      packages = p: [ p.xmonad-contrib ];
+    };
+    defaultPackage = pkgs.haskellPackages.xmonad-contrib;
+  }) // { inherit overlay overlays; } ;
+}


IvanMalison:
This does need https://github.com/xmonad/xmonad/pull/316 in order to technically work properly, but I tested by pointing it at my branch.

IvanMalison:
The build will fail until xmonad/xmonad#316 is merged

liskin:
Hm, even though you [guarantee that the build would be 10× faster](https://github.com/xmonad/xmonad/pull/316#issuecomment-890308172) (😃), it's not really the case and the Nix workflow in xmonad-contrib is the slowest of the three. That's probably because xmonad-contrib is huge and benefits a lot from building modules in parallel by passing `-j` to GHC, and we do that in the two other workflows:
https://github.com/xmonad/xmonad-contrib/blob/master/cabal.haskell-ci#L15
https://github.com/xmonad/xmonad-contrib/blob/master/stack-master.yaml#L26
Can we do that in the Nix build as well? How?

IvanMalison:
> it's not really the case and the Nix workflow in xmonad-contrib is the slowest of the three.

To be fair, that's because you cache dependencies as one giant blob for the other 2 systems and load them wholesale. Not really a fair comparison at all.

If you compared someone just running `stack build` or `cabal new-build` vs `nix build` on a freshly cloned repository it would indeed be 10x faster.

Another reason for it being slower is that the nix build does multiple versions of the build.

As for parallel building, it seems that this was enabled and then disabled:

https://github.com/NixOS/nixpkgs/pull/4554

If we really really want it, it should be possible to revert the revert and use a custom version of nixpkgs.

IvanMalison:
Actually if you look at the build log, it looks like we are already using the -j flag:
```
xmonad> setupCompileFlags: -package-db=/build/setup-package.conf.d -j2 +RTS -A64M -RTS -threaded -rtsopts
```

IvanMalison:
> To be fair, that's because you cache dependencies as one giant blob for the other 2 systems and load them wholesale. Not really a fair comparison at all.

Likely that caching /nix/store would help a lot, especially if we also add a flake.lock.

liskin:
> Another reason for it being slower is that the nix build does multiple versions of the build.

Oh, yeah, looking at the logs it seems the nix build enables profiling so it compiles xmonad-contrib twice:

```
Sun, 01 Aug 2021 21:27:27 GMT
xmonad-contrib> [  1 of 295] Compiling XMonad.Actions.AfterDrag ( XMonad/Actions/AfterDrag.hs, dist/build/XMonad/Actions/AfterDrag.o, dist/build/XMonad/Actions/AfterDrag.dyn_o )
Sun, 01 Aug 2021 21:28:53 GMT
xmonad-contrib> [295 of 295] Compiling XMonad.Actions.EasyMotion ( XMonad/Actions/EasyMotion.hs, dist/build/XMonad/Actions/EasyMotion.o, dist/build/XMonad/Actions/EasyMotion.dyn_o )
```

```
Sun, 01 Aug 2021 21:28:55 GMT
xmonad-contrib> [  1 of 295] Compiling XMonad.Actions.AfterDrag ( XMonad/Actions/AfterDrag.hs, dist/build/XMonad/Actions/AfterDrag.p_o )
Sun, 01 Aug 2021 21:30:36 GMT
xmonad-contrib> [295 of 295] Compiling XMonad.Actions.EasyMotion ( XMonad/Actions/EasyMotion.hs, dist/build/XMonad/Actions/EasyMotion.p_o )
```

> it looks like we are already using the -j flag

Oh, okay. Still, the stack build is faster:

```
Sun, 01 Aug 2021 21:26:40 GMT
[  1 of 295] Compiling XMonad.Actions.AfterDrag
Sun, 01 Aug 2021 21:27:07 GMT
[295 of 295] Compiling XMonad.Actions.EasyMotion
```

~30 seconds instead of ~1 min 30 sec for the builds nix does. Any way to check if using `-j` instead of `-j2` helps?

Note that we only enable `-j` for xmonad-contrib, not for all the other packages, because those are built in parallel by stack/cabal. This was discussed here: https://github.com/xmonad/xmonad-contrib/pull/431, and I guess it might be one of the reasons it's not a good idea to enable it globally in nix either. So if we could just change the flake to add `-j` to ghc-options for xmonad-contrib, like we do for cabal and stack, that could be best?

I'm not sure caching the entire nix store is the correct solution here. I guess that might cut the first minute (the one where nix logs nothing and just does I don't what) off nix build, and then probably also the xmonad core build would be cached, but xmonad-contrib would still often be built from scratch, and that takes at least 3 minutes now. Oh and caching the flake.lock… we probably don't want that do we? That's not a good tradeoff. :-)

IvanMalison:
@liskin One option would be to use nix develop and then run `cabal new-build` inside of that instead of using nix-build. Then we could customize the build flags in exactly the same way as we do with the other cabal builds.

IvanMalison:
> Oh and caching the flake.lock… we probably don't want that do we? That's not a good tradeoff. :-)

All that flake.lock would do in this case is freeze the nixpkgs and xmonad depedencies at a specific version. If we ever needed to bump those we would do so explicitly.

IvanMalison:
> ~30 seconds instead of ~1 min 30 sec for the builds nix does. Any way to check if using `-j` instead of `-j2` helps?

Looks like this flag gets set here:
https://github.com/NixOS/nixpkgs/blob/aaf9676fbb7fb4570216ca1e189a3dc769d62c45/pkgs/development/haskell-modules/generic-builder.nix#L185

seems like its being set to the number of cores which seems sensible to me.

IvanMalison:
We could also disable profiling with: https://github.com/NixOS/nixpkgs/blob/aaf9676fbb7fb4570216ca1e189a3dc769d62c45/pkgs/development/haskell-modules/generic-builder.nix#L31

liskin:
> seems like its being set to the number of cores which seems sensible to me.

I'll try stack build with `-j2` to see if it slows down to 1 min 30 sec or if that's something else.

> We could also disable profiling

Would it still use cached dependencies?

liskin:
> I'll try stack build with `-j2` to see if it slows down to 1 min 30 sec or if that's something else.

Hm, it still builds it in 30 seconds. :-/

IvanMalison:
> Would it still use cached dependencies?

Yes, we would only disable it for the xmonad-contrib package.

liskin:
I'll go to bed now, so let's wrap it up somehow: ideally we'd figure out why ghc in nix build is three times slower than in stack/cabal builds, and fix that. This should turn 3 minutes into 1, and pull request checks won't get slower by merging this. If we fail to do that, caching nix store and possibly also disabling profiling should also cut it down to acceptable levels.

(I hope you don't mind me obsessing about these details. But I did get praise from contributors for speeding up the PR checks, so I'd love to keep them fast.)

IvanMalison:
> I hope you don't mind me obsessing about these details

No, makes sense. quick ci is a good goal.

IvanMalison:
@liskin The nix develop approach did seem to reduce the time it took somewhat (mostly bc the profile build no longer happens)

The Nix build is still about a minute and a half longer than the other builds. I think we would have to cache the nix store to fix this.

IvanMalison:
Not that this is the best comparison, but on my local machine I get:
``
nix develop -c cabal v2-build -j4  55.33s user 1.69s system 99% cpu 57.065 total
```

liskin:
> ideally we'd figure out why ghc in nix build is three times slower than in stack/cabal builds, and fix that

Turns out the haskell-ci (cabal-install) workflow was also a 1min30s, because only the stack build had `--fast` (`-O0`) — [fixed here](https://github.com/xmonad/xmonad-contrib/commit/8de415743d2765fe121c492109fa107c30e0d371). So that answers that bit. Can we tell Nix to disable GHC optimizations in the CI workflow without disabling it for all users of the flake?

> The nix develop approach did seem to reduce the time it took somewhat

Seems a bit ugly to invoke `cabal v2-build` manually. 😞 Are we really testing the flake then? 🤔

IvanMalison:
> Seems a bit ugly to invoke `cabal v2-build` manually.  Are we really testing the flake then?

I would argue we're still getting most of what we need, since we are using nixDevelop. We are still using nix to process the cabal file, fetch all depdencies and build the environment in which the package is built.

The thing is that the alternative is to change what happens with defaultPackage, but I don't think we actually want those flags on by default right?

liskin:
> but I don't think we actually want those flags on by default right?

Yeah, definitely not by default, only to make CI faster. :-/

liskin:
Anyway, if you believe `nix develop … cabal v2-build` is good enough, I'm not going to challenge that as I really have no idea about the details there (same as the `nix flake check` thing, it looks desirable but I had no idea what it actually does ☺). So let's just add `-O0` to `cabal v2-build` and we should be good to go.

PR #583, opened 2021-08-05
X.L.Magnifier: Expose types

IvanMalison:
### Description

It is important for these types to be exposed to access the underlying LayoutModifier. It is not possible to interact with it without these.

In particular I needed these changes for compatibility with this PR https://github.com/xmonad/xmonad-contrib/pull/582

### Checklist

### DIFF
--- a/XMonad/Layout/Magnifier.hs
+++ b/XMonad/Layout/Magnifier.hs
@@ -42,7 +42,8 @@ module XMonad.Layout.Magnifier
       -- * Messages and Types
       MagnifyMsg (..),
       MagnifyThis(..),
-      Magnifier,
+      Magnifier(..),
+      Toggle(..)
     ) where
 
 import Numeric.Natural (Natural)


liskin:
Is this useful without [the layout inspection](https://github.com/xmonad/xmonad-contrib/pull/119#issuecomment-744111659) bits?

IvanMalison:
Yes -- I actually don't see how this would be related to layout inspection at all.

This is needed for use with my conditional layout modifier module. Atm, Magnify only provides functions that return functions  Modified Payouts, not the layout modifiers themselves. My conditionallayout modifier needs to interact directly with the layout modifier obejcts so it cannot use these functions.

liskin:
Oh, mea culpa, now I see. For some reason I though this is because you wanted to make a condition that looks into Magnifier and completely overlooked the “little” issue that ConditionalLayoutModifier doesn't work with a `l a -> ModifiedLayout Magnifier l a`. Sorry. 🙇

What about trying to deal with that issue directly instead of exposing internals of every LayoutModifier someone wants to use with the ConditionalLayoutModifier? Should be easy to create a function

```haskell
conditional ∷ … ⇒ c → ModifiedLayout m l a → ModifiedLayout (ConditionalLayoutModifier m c) l a
```

so then you just do `conditional c $ magnifier $ Tall … ||| …`.
Or perhaps

```haskell
conditional ∷ … ⇒ c → (l a → ModifiedLayout m l a) → (l a → ModifiedLayout (ConditionalLayoutModifier m c) l a)
```

and then the usage is `conditional c magnifier $ Tall … ||| …`.

(Also, a random thought for #582 before I get to replying in detail to your other questions: if you implement it as an `instance LayoutClass (ConditionalModifiedLayout c m l) a` instead of `instance LayoutModifier (ConditionalLayoutModifier m c) a`, you won't need to implement as many functions and the reasoning may be easier, too. Or not. Just a random thought, really.)

PR #586, opened 2021-08-08
X.A.Submap, X.U.Ungrab: Sync after ungrab (fixes issues with runProcessWithInput)

liskin:
### Description

When `runProcessWithInput` is invoked immediately after `ungrabPointer`/`ungrabKeyboard`, we don't actually ungrab at all because `runProcessWithInput` blocks and the ungrab requests wait in Xlib's queue for a request that needs a queue flush.

Common uses of `unGrab` (before `spawn`) aren't followed by a blocking action, so the ungrab requests are flushed by xmonad's main loop, and this is merely a timing issue—fork/exec takes a while and xmonad probably manages to get back to its main loop in time. Uses of `runProcessWithInput` in ordinary non-submap key bindings happen to work because key bindings are passive grabs—the grab is released by the user's fingers releasing the key itself, even if xmonad's ungrab requests are stuck in a blocked queue. Submap key bindings, however, take an active grab and therefore need to ungrab explicitly.

Easy fix—explicit `sync`.

Fixes: https://github.com/xmonad/xmonad/issues/313

### Checklist

        manually, ...) and concluded:

    Appears to have fixed the issue and I'll ask the reporter of https://github.com/xmonad/xmonad/issues/313 to verify.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -681,6 +681,11 @@
 
     - Added `swapWithCurrent` and `swapOrder` to the list of exported names.
 
+  - `XMonad.Actions.Submap`, `XMonad.Util.Ungrab`:
+
+    - Fixed issue with keyboard/pointer staying grabbed when a blocking action
+      like `runProcessWithInput` was invoked.
+
 ## 0.16
 
 ### Breaking Changes
--- a/XMonad/Actions/Submap.hs
+++ b/XMonad/Actions/Submap.hs
@@ -92,5 +92,6 @@ submapDefaultWithKey defAction keys = do
 
     io $ ungrabPointer d currentTime
     io $ ungrabKeyboard d currentTime
+    io $ sync d False
 
     fromMaybe (defAction (m', s)) (M.lookup (m', s) keys)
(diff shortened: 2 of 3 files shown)


geekosaur (approved):
Looks good to me.

liskin:
It just occurred to me that we might also (or perhaps only) want `sync` in [`runProcessWithInput`](https://github.com/liskin/xmonad-contrib/blob/91995df5592f0d718f53d59f8ef2cc111dd7485b/XMonad/Util/Run.hs#L58) to always flush pending requests, even if `runProcessWithInput` is used elsewhere. Probably won't harm anything, will it?

geekosaur:
I don't think it'll hurt anything, but I'm also not sure it's `runProcessWithInput`'s job.

liskin:
It's not its job, but it's probably the only thing in xmonad(-contrib) that blocks, so it's understandable that everything else assumes explicit `sync` is unnecessary. It'd be a hack, but then the entire Xlib kind of is. :-))

liskin:
On the other hand it might be completely unnecessary. We won't know until someone files another bug, and that may never happen (even if there is one!).

liskin:
Let's just merge this as it is; should it ever break again, I'll add that other `sync`.

PR #589, opened 2021-08-12
Provide a mechanism by which users can customize the nix devShell

IvanMalison:
This allows users to use a file develop.nix to customize the devShell provided by the flake. This is useful if e.g. the user wants to have the haskell-language-server or other developer tools in the shell properly configured (correct ghc versions etc).

Here is an example develop.nix for haskell-language-server:
```
pkgs: devInputs: devInputs // {
  nativeBuildInputs = with pkgs.haskellPackages;
    [ cabal-install hlint ghcid ormolu implicit-hie haskell-language-server ];
}
```

### DIFF
--- a/xmonad-contrib.cabal
+++ b/xmonad-contrib.cabal
@@ -23,7 +23,7 @@ license:            BSD3
 license-file:       LICENSE
 author:             Spencer Janssen & others
 maintainer:         xmonad@haskell.org
-extra-source-files: README.md CHANGES.md scripts/generate-configs scripts/run-xmonad.sh
+extra-source-files: README.md CHANGES.md NIX.md scripts/generate-configs scripts/run-xmonad.sh
                     scripts/window-properties.sh
                     scripts/xinitrc scripts/xmonad-acpi.c
                     scripts/xmonad-clock.c
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ tags
 
 stack.yaml.lock
 
+flake.lock
--- a/NIX.md
+++ b/NIX.md
@@ -1,0 +1,17 @@
+# `nix` integration for XMonad
+
+## Customizing the `nix-shell`
+
+It's possible to use a file `develop.nix` to customize the `devShell`
+provided by the flake.  This is useful if e.g. you want to have the
+`haskell-language-server` or other developer tools in the shell properly
+configured (correct GHC versions, and the like).
+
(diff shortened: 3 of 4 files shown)


slotThe (commented):
I see that—in the nix spirit—all of this is undocumented.

I liked the example you gave, any way that may be incorporated into the flake (or be in an additional readme, that part does not really matter)?

IvanMalison:
> I liked the example you gave, any way that may be incorporated into the flake (or be in an additional readme, that part does not really matter)?

Sorry I'm not sure I understand what you are saying here.

> I see that—in the nix spirit—all of this is undocumented.

Are you saying that you would like me to document this somehow? I'm not really sure where I would put such documentation.

slotThe:
> Are you saying that you would like me to document this somehow?

Exactly, yes. 

> Sorry I'm not sure I understand what you are saying here.

I'm saying that something really close to what you wrote in the PR description 

> Here is an example develop.nix for haskell-language-server:
>
> ```
> pkgs: devInputs: devInputs // {
>   nativeBuildInputs = with pkgs.haskellPackages;
>     [ cabal-install hlint ghcid ormolu implicit-hie haskell-language-server ];
> }
> ```

would be really nice to have as documentation.

> I'm not really sure where I would put such documentation.

It's a bit iffy, but that's why I mentioned the extra README (perhaps we could introdce a `nix` directory, where all of this stuff goes).  Alternatively, I don't see any reasons this could not be the top-level documentation of `modifyDevShell` itself

slotThe:
I just created a `NIX.md` with the information from the OP for now.  This also seems like a good place to document bits like https://github.com/xmonad/xmonad/issues/332 should anyone feel motivated to do so

xmonad-contrib.cabal
```diff
@@ -23,7 +23,7 @@ license:            BSD3
 license-file:       LICENSE
 author:             Spencer Janssen & others
 maintainer:         xmonad@haskell.org
-extra-source-files: README.md CHANGES.md scripts/generate-configs scripts/run-xmonad.sh
+extra-source-files: README.md CHANGES.md NIX.md scripts/generate-configs scripts/run-xmonad.sh
```
liskin, line 26: We aren't shipping the nix flake in the tarball, so does it make sense to ship a doc for it?

slotThe: Hm, you're probably right.  I'll push a commit to remove that at some point later today

PR #594, opened 2021-08-14
X.A.WindowBringer: Add a filter function to config

IvanMalison:
### Description

A new field `windowFilter` was added to the config, which allows the user to provide a function which will decide whether each window should be included in the window bringer menu.

### Checklist

        manually, ...) and concluded: I've been using this in my config for a couple of years with no issues. No testing was done with xmonad-testing, but I don't think it is necessary.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -419,8 +419,12 @@
 
   * `XMonad.Prompt.WindowBringer`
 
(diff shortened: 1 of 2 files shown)


slotThe:
I added docs to the `windowFilter` field, but other than that this looks good to me; thanks!

PR #596, opened 2021-08-17
X.U.Loggers: Add `logTitlesOnScreen`

slotThe:
### Description

This works like `logTitles`, but gets an explicit screen to log the window
titles on.  This may be useful when having status bars for each screen
that show all windows on their respective visible workspaces.

### Checklist

        manually, ...) and concluded: It's a trivial extension of `logTitles` and I've confirmed manually that this work for me.  It might be conceivable to property test this, as we are only accessing internal state, but I did not deem this necessary in this case.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -501,7 +501,9 @@
     - Added `logConst` to log a constant `String`, and `logDefault` (infix: `.|`)
       to combine loggers.
 
-    - Added `logTitles` to log all window titles (focused and unfocused ones).
+    - Added `logTitles` to log all window titles (focused and unfocused
+      ones) on the focused workspace, as well as `logTitlesOnScreen` as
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -503,6 +503,8 @@
 
     - Added `logTitles` to log all window titles (focused and unfocused ones).
 
+    - Added `logTitlesOnScreen` as a screen-specific variant of `logTitles`.
```
TheMC47, line 506: Should we merge these two entries together?

TheMC47 (approved):
LGTM, I don't think it's necessary to add property tests

CHANGES.md
slotThe: Oh yeah we definitely should, thanks

PR #600, opened 2021-09-02
xmonad-contrib part of "Make extensibleState primarily keyed by TypeRep instead of type names"

liskin:
### Description

We've been using the String we get out of `show . typeOf` as key in `extensibleState`, but that has a somewhat serious bug: it shows unqualified type names, so if two modules use the same type name, their extensible states will be stored in one place and get overwritten all the time.

To fix this, the `extensibleState` map is now primarily keyed by the TypeRep themselves, with fallback to String for not yet deserialized data. XMonad.Core now exports `showExtType` which serializes type names qualified, and this is used in `writeStateToFile`.

A simpler fix would be to just change the serialization of type names in `XMonad.Util.ExtensibleState`, but I'm afraid that might slows things down: Most types used here will start with "XMonad.", and that's a lot of useless linked-list pointer jumping.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/94
Related: https://github.com/xmonad/xmonad/pull/326

---

I must admit I don't really like the code, especially the xmonad-contrib part. There's the assumption that a Right key maps to a Right value, and Left key to Left value, but enforcing that in the types probably means adding a another field to the XState.

### Checklist

        manually, ...) and concluded:

    (TODO)

### DIFF
--- a/XMonad/Util/ExtensibleState.hs
+++ b/XMonad/Util/ExtensibleState.hs
@@ -1,5 +1,10 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeOperators #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Util.ExtensibleState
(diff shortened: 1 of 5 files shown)


stack-master.yaml
```diff
@@ -9,8 +9,8 @@ packages:
 extra-deps:
 - github: xmonad/X11
   commit: master@{today}
-- github: xmonad/xmonad
-  commit: master@{today}
+- github: liskin/xmonad
```
geekosaur, line 12: Don't forget to change this!

XMonad/Util/ExtensibleState.hs
```diff
@@ -24,14 +29,18 @@ module XMonad.Util.ExtensibleState (
                               , gets
                               , modified
                               , modifiedM
+
+#ifdef TESTING
```
geekosaur, line 33: Does this really want to be conditional? Looks to me like you need it any time you're upgrading from an xmonad without this change.

geekosaur (approved):
TBH I don't see why you don't like the code here; it seems clean enough and in fact cleans up some existing code. In short, LGTM.

Issue #527, opened 2021-04-23
New Feature: Add killOthers to XMonad.Actions.WithAll (place not decided yet)


oogeek:
### Note

This issue is intended for new contributors to xmonad :tada: . Feel free to ask for help when working on this by commenting below or saying Hi in the IRC channel smiley

The place to put the function is not decided yet.

Hints:
You can check the code for `KillAll` from `X.A.WithAll` and `Integrate'` from `XMonad.StackSet`

### Feature Description

When someone needs to close all other windows in a workspace and leave the focused one alone, multiple `kill1`s from `XMonad.Actions.CopyWindow` are just too much. It will be nice to have a `killOthers` function to do so.

### (Potential) Usage example
```
killOthers :: X ()
myKeys = [
...
 , ("M-S-a", killOthers)
...
```

### Checklist

  - [X] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

ssbothwell:
how about this?

```
-- | Apply an 'X' operation to all unfocused windows, if there are any.
withUnfocused :: (Window -> X ()) -> X ()
withUnfocused f = withWindowSet $ \ws ->
    case peek ws of
        Just w ->
            let unfocusedWindows = XMonad.Prelude.filter (/= w) $ index ws
            in mapM_ f unfocusedWindows
        Nothing -> pure ()

-- | Kill all the unfocused windows on the current workspace.
killOthers :: X ()
killOthers = withUnfocused killWindow
```

It would make sense to me to put `withUnfocused` in `XMonad.Operations` and `killOthers` in `XMonad.Actions.WithAll`

slotThe:
> It would make sense to me to put `withUnfocused` in `XMonad.Operations` and `killOthers` in `XMonad.Actions.WithAll`

X.Operations sounds good to me, but I'm not sure X.A.WithAll really fits the bill of not doing something to all windows.  That said, it's probably the closest module we have and it has to live inside _someting_ so...  Feel free to open the respective PRs :)

ssbothwell:
I agree about `X.A.WithAll`. If we weren't worried about breaking changes we could rename the module and description to something like:

```
-- Provides functions for performing a given action on sets of windows of
-- the current workspace.
-----------------------------------------------------------------------------

module XMonad.Actions.With (
```

[...]

PR #602, opened 2021-09-16
Adds killOthers function to Xmonad.Actions.WithAll

ssbothwell:
### Description

Resolves https://github.com/xmonad/xmonad-contrib/issues/527

Adds `killOthers` function to `XMonad.Actions.WithAll`. Depends on https://github.com/xmonad/xmonad/pull/327

### Checklist

### DIFF
--- a/XMonad/Actions/WithAll.hs
+++ b/XMonad/Actions/WithAll.hs
@@ -5,15 +5,16 @@
 -- Stability    : unstable
 -- Portability  : unportable
 --
--- Provides functions for performing a given action on all windows of
--- the current workspace.
+-- Provides functions for performing a given action on all or certain
+-- groups of windows on the current workspace.
 -----------------------------------------------------------------------------
 
 module XMonad.Actions.WithAll (
     -- * Usage
     -- $usage
     sinkAll, withAll,
-    withAll', killAll) where
+    withAll', killAll,
+    killOthers) where
 
 import XMonad.Prelude hiding (foldr)
 
@@ -50,3 +51,7 @@ withAll f = withWindowSet $ \ws -> let all' = integrate' . stack . workspace . c
 -- | Kill all the windows on the current workspace.
 killAll :: X()
 killAll = withAll killWindow
+
+-- | Kill all the unfocused windows on the current workspace.
+killOthers :: X ()
+killOthers = withUnfocused killWindow
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -302,6 +302,11 @@
 
   * Add support for GHC 9.0.1.
 
+  * `XMonad.Actions.WithAll`
+
+    - Added `killOthers`, which kills all unfocused windows on the
+      current workspace.
+
   * `XMonad.Prompt.XMonad`
 
     - Added `xmonadPromptCT`, which allows you to create an XMonad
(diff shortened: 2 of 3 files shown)


XMonad/Actions/WithAll.hs
```diff
@@ -50,3 +51,7 @@ withAll f = withWindowSet $ \ws -> let all' = integrate' . stack . workspace . c
 -- | Kill all the windows on the current workspace.
 killAll :: X()
 killAll = withAll killWindow
+
+-- | Kill all the unfocused windows on the current workspace.
```
slotThe, line 55: Now that `WithAll` contains functions that do not act on all windows, it might be nice to mention this somewhere in the documentation of the module, as well as X.D.Extending

ssbothwell: What would you like in the module documentation?

```
-- Provides functions for performing a given action on all windows and
-- all unfocused windows of the current workspace.
```

ssbothwell: We could also now add:

```
sinkOthers :: X ()
sinkOthers = withUnfocused sink
```

slotThe: > What would you like in the module documentation?

Perhaps something about "all or only certain groups of windows" 

> rename this module something like XMonad.Actions.Grouped?

It would probably be for the best, but renaming modules and keeping their old version around is always a bit of a pain :/

ssbothwell: > Perhaps something about "all or only certain groups of windows"

how does this look?
  
> It would probably be for the best, but renaming modules and keeping their old version around is always a bit of a pain :/

We could have `Xmonad.Actions.Grouped` with everything and then this module can re-export the old functions?

slotThe: > how does this look?

Looks fine to me; thanks!

> We could have `Xmonad.Actions.Grouped` with everything and then this module can re-export the old functions?

Indeed; but we'd probably still want to deprecate `WithAll` in that case.  I'm not sure it's quite warranted right now with just one additional function, but maybe as things evolve we'll eventually be forced to do this

ssbothwell: That  makes sense. Maybe we can put together a list of additional grouped actions we would like? `sinkOthers` is the first obvious one I can think of, but it will require a variation of `withUnfocused` like `withAll'`

slotThe: I have to say I've never felt the need for acting on so many windows at once, so naturally I can't think of any :)  This may or may not be the PR "feature creeping", so perhaps it's better to get this small change merged now and think of adding an X.A.Grouped module in another step; what do you think?

ssbothwell: Yeah that makes sense to me. Lets get this one merged and I'll write a new issue to discuss changing the module

slotThe:
> Yeah that makes sense to me. Lets get this one merged and I'll write a new issue to discuss changing the module

Neat; thanks!

I pushed a few minor nits regarding the wording in CHANGES.md and the commit message—we have a semi-standard way of doing this (sorry for force-pushing over your `master` branch; better create a new one next time :)).

ssbothwell:
> > Yeah that makes sense to me. Lets get this one merged and I'll write a new issue to discuss changing the module
> 
> Neat; thanks!
> 
> I pushed a few minor nits regarding the wording in CHANGES.md and the commit message—we have a semi-standard way of doing this (sorry for force-pushing over your `master` branch; better create a new one next time :)).

no problem! I forgot to make a branch this time.

PR #577, opened 2021-07-26
X.A.UpdateFocus: Add focusUnderPointer

liskin:
### Description

Some people like their mouse pointer to move when changing focus with the keyboard, other people like their pointer to stay and focus to follow. xmonad(-contrib) supports both preferences, but imperfectly: The former requires using the XMonad.Actions.UpdatePointer contrib module, the latter (focusFollowsMouse) only reacts to CrossingEvent; the focus isn't updated after changing workspaces or layouts.

This adds an inverse of XMonad.Actions.UpdatePointer.updatePointer that immediately updates the focus instead.

Fixes: https://github.com/xmonad/xmonad/issues/108

### Checklist

  - [X] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [X] I've considered how to best test these changes (property, unit,
        manually, ...) and concluded:

    I tested this a bit with a minimal config file and I hope the reporter and commenters of https://github.com/xmonad/xmonad/issues/108 will test it some more. I do not intend to use this myself.

  - [X] I updated the `CHANGES.md` file

  - n/a I updated the `XMonad.Doc.Extending` file (if appropriate)

PR #605, opened 2021-09-19
`ppPrinters` for custom workspace types, `copiesPP` fix

### LINKED ISSUE #557: copiesPP not working with my configuration
### Problem Description
how do I add ``copiesPP`` to this configuration?

### Configuration File
```haskell
myXmobarPP :: PP
myXmobarPP = def
  { ppSep           = gray " │ "
  , ppTitleSanitize = xmobarStrip
  , ppCurrent       = blue . wrap "" "" . xmobarBorder "Bottom" "#8be9fd" 4
  , ppHidden        = lowWhite . wrap "" ""
  -- , ppHiddenNoWindows = lowWhite . wrap " " ""
  , ppWsSep         = "  "
  , ppTitle = magenta . xmobarAction "xdotool key Super+shift+c" "2" . shorten
                50
  , ppOrder         = \[ws, l, t, ex] -> [ws, l, ex, t]
  , ppExtras        = [xmobarColorL "#ff6c6b" "" $ windowCount]
  , ppLayout = green . xmobarAction "xdotool key Super+space" "1" . xmobarAction
                 "xdotool key Super+shift+space"
                 "3"
  }
 where
  blue, lowWhite, magenta, red, white, yellow, purple, green
    :: String -> String
  magenta  = xmobarColor "#b48ead" ""
  blue     = xmobarColor "#51afef" ""
  purple   = xmobarColor "#bd93f9" ""
  lowBlue  = xmobarColor "#8be9fd" ""
  white    = xmobarColor "#f8f8f2" ""
  yellow   = xmobarColor "#f1fa8c" ""
  red      = xmobarColor "#ff6c6b" ""
  lowWhite = xmobarColor "#a6aebf" ""
  gray     = xmobarColor "#434c5e" ""
  green    = xmobarColor "#c3e88d" ""

mySB :: StatusBarConfig
mySB = statusBarProp
  "xmobar ~/.config/xmonad/xmobar/xmobarrc"
  (clickablePP (filterOutWsPP [scratchpadWorkspaceTag] myXmobarPP))
```

### Checklist

  - I tested my configuration
    - [x] With `xmonad` commit 6d184e859ced72b8266d5d5b37004119da8f6dd9
    - [x] With `xmonad-contrib` commit 6a7eb85e84ddc2091706fbce5ff293e859481e51

TheMC47:
### Description

Using `ppPrinters` wiht `WorkspacePredicate` and `WorkspaceFormatter`
allows users to define custom workspace types, beyond the ones
integrated in PP (i.e. urgent, current, visible, visible with no
windows, hidden, and hidden with no windows). `WorkspacePredicate`s are
added for these predicates (`isType`) with unsafe versions that assume
that predicates with a higher precedence already faield `isType'`.
`WorkspacePredicate`s can also be combined and modified with `notWP`,
`andWP`, and `orWP`.

This also fixes #557 

### Checklist

  - [x] I've considered how to best test these changes manually and concluded: #577 is indeed fixed, and I can imagine this being useful in some cases


XMonad/Hooks/StatusBar/PP.hs
```diff
@@ -200,17 +210,105 @@ pprWindowSet :: WorkspaceSort -> [Window] -> PP -> WindowSet -> String
...
+        -> WindowSet
+        -> WindowSpace
+        -> Bool
+
+-- | Predicate for urgent workspaces.
+isUrgent :: WorkspacePredicate
+isUrgent urgents s w = any (\x -> (== Just (S.tag w)) (S.findTag x s)) urgents
+
+-- | Predicate for the current workspace. Caution: assumes default
+-- precedence is respected.
```
slotThe, line 246: It may not be that clear _which_ default precedence it meant here

XMonad/Hooks/StatusBar/PP.hs
```diff
@@ -200,17 +210,105 @@ pprWindowSet :: WorkspaceSort -> [Window] -> PP -> WindowSet -> String
...
+        `andWP` isHidden'
+
+-- | Negates a 'WorkspacePredicate'.
+notWP :: WorkspacePredicate -> WorkspacePredicate
+notWP p = not .:. p
+  where  (.:.) = (.) . (.:)
+
+-- | '&&' for 'WorkspacePredicate's.
+andWP :: WorkspacePredicate -> WorkspacePredicate  -> WorkspacePredicate
+andWP = liftM2x3 (&&)
```
slotThe, line 298: This is kind of fun because a stacking of `liftA2`s seems completely over the top for this, and yet it seems much nicer to read than

``` haskell
wPred `andWP` wPred' = \wins winSet winSpace -> wPred  wins winSet winSpace 
                                             && wPred' wins winSet winSpace
```

Although I question which one is clearer a bit :)

slotThe (commented):
Adding yet another field to `PP` seems unfortunate, but it is probably necessary and unavoidable in this case; eth approach overall looks quite decent to me.

A small typo I noticed in the commit message:

> Using `ppPrinters` **wiht** `WorkspacePredicate` and `WorkspaceFormatter`

TheMC47:
> Adding yet another field to `PP` seems unfortunate, but it is probably necessary and unavoidable in this case; eth approach overall looks quite decent to me.

I think this field should extend `PP`s life by quite a large time... But it makes me wonder what would be a better approach

XMonad/Hooks/StatusBar/PP.hs
TheMC47: https://github.com/xmonad/xmonad-contrib/pull/605/commits/55adca6f430dc85d189ced2bb6e91b4a1b00279f ?

XMonad/Hooks/StatusBar/PP.hs
TheMC47: Either way is good for me, I was just practicing some Haskell wizardry

XMonad/Hooks/StatusBar/PP.hs
slotThe: :+1:

XMonad/Hooks/StatusBar/PP.hs
slotThe: Eh, probably not that important; just go with your gut I suppose :)

slotThe:
I like this approach, especially the fact that we don't have to define custom operators!

---

``` haskell
{-# LANGUAGE RecordWildCards #-}
```

I don't know if this is just me, but I always find `RecordWildcards` to be terribly unreadable; perhaps `NamedFieldPuns` would be nicer?

---

> What I don't like is the need for ReaderT. As explained in one of the comments, there is not and there cannot be an Alternative instance for `r -> Maybe a`, so I use ReaderT for `ppPrinters` and a simple `WS → Bool` reader monad for the conditions, and it's all a bit confusing. Probably not really Haskell-beginner friendly as it requires knowledge of all these instances. :-/

``` haskell
-- XXX: ReaderT instead of -> because there is no
--   instance (Alternative m) => Alternative (Λa. r -> Maybe a)
-- (there cannot be in Haskell, ReaderT is essentially (r ->) ∘ Maybe)
-- But we don't need Alternative for WS -> Bool, so we use the simple
-- function-based reader there. This may be a bit confusing. :-/
```

I think people reading xmonad source code already need to be familiar with monad transformers anyways, so there's no real harm in doing this.  Especially with documentation outlining why this was done the way it is.  (I find the implicit `Reader` usage in `ppRename` more confusing than this usage of `ReaderT`, personally)

Completely overthinking this, while it's true that there can't be an `Alternative` instance for `∀ a. r -> Maybe a` in Haskell, we can certainly just write down the composition (`Compose ((->) r) Maybe a`), for which there is an instance because it's also a `newtype`.  Dressing this up as a Kleisli arrow we could define

``` haskell
type (~>) r a = Compose ((->) r) Maybe a
```

and be on our way.  Although since `Compose` is a `newtype` as well it wouldn't necessarily shorten the implementation.  I'm also not sure that we wouldn't lose even more people with this :)

TheMC47:
I really like your approach @liskin! And I don't see using `ReaderT` as a problem, I agree with @slotThe  that `ppRename` is more confusing in that regard. I'm not that familiar with the concepts you used, but I understood what's going on quickly with the `ReaderT` docs open and just reading through the comments. I think it's fine, and quite elegant. Kudos :tada:

liskin:
> I don't know if this is just me, but I always find `RecordWildcards` to be terribly unreadable; perhaps `NamedFieldPuns` would be nicer?

Sure, no strong opinion on this one.

> I think people reading xmonad source code already need to be familiar with monad transformers anyways, so there's no real harm in doing this. Especially with documentation outlining why this was done the way it is.

Okay. :-)

> Completely overthinking this, while it's true that there can't be an `Alternative` instance for `∀ a. r -> Maybe a` in Haskell, we can certainly just write down the composition (`Compose ((->) r) Maybe a`), for which there is an instance because it's also a `newtype`. Dressing this up as a Kleisli arrow we could define
> 
> ```haskell
> type (~>) r a = Compose ((->) r) Maybe a
> ```
> 
> and be on our way. Although since `Compose` is a `newtype` as well it wouldn't necessarily shorten the implementation. I'm also not sure that we wouldn't lose even more people with this :)

Oh, I didn't know this one! And there even is an `Alternative` instance for `Compose`. The `newtype` wouldn't be a problem if there was a `MonadReader` instance, but I don't think there is, and `ReaderT` is probably still easier to understand anyway. But thanks a lot for teaching me something new. :-)

---

I just pushed https://github.com/liskin/xmonad-contrib/commit/089d80f3d8f6258602846b994c756afd368c610b which makes it possible to easily use the default/fallback `ppPrinters` and then maybe do something with their output. Like for example we can now have a `copiesPP'` that instead of using a different printer for workspaces with copies just adds a `+` at the end. (To add `+` inside of the parentheses/brackets and colors set by `ppCurrent` et al, `ppRename` is probably sufficient.) Surely there are other interesting uses as well.

So… can I now leave it up to @TheMC47 to clean this all up and fix the docs? Thanks!

TheMC47:
> So… can I now leave it up to @TheMC47 to clean this all up and fix the docs? Thanks!

All good :+1:

TheMC47:
In the end, I didn't expose `?->` since it might be confusing (as @geekosaur mentioned in the IRC channel), and the gain from it is minimal (right now). We can change it if we see the need for that in the future, but it's not a must to figure this out before the release.

Regarding the name, nothing else came to mind.

PS: I'm gonna squash my last commit and @liskin's commits later

TheMC47:
> Surely there are other interesting uses as well.

I think the interface is really flexible, but I fear that `fallbackPrinters` wouldn't quite do it since we need `PP` as an argument (maybe I'm missing something). Defining `type WSPP = WSPP' (PP -> WorkspaceId -> String)` would solve this though

geekosaur:
Just to document my "confusing" comment, `X.H.ManageHelpers` already uses `-?>` and the similarity is a bit uncomfortable.

PR #606, opened 2021-09-22
Remove codesearch, isount and openstreetmap engines

alternateved:
### Description

This is mostly a cosmetic change. Removed search engines point to sites that do not exist anymore or their functionality has changed. 

### Checklist

### DIFF
--- a/XMonad/Actions/Search.hs
+++ b/XMonad/Actions/Search.hs
@@ -250,7 +250,7 @@ search browser site query = safeSpawn browser [site query]
    appends it to the base. You can easily define a new engine locally using
    exported functions without needing to modify "XMonad.Actions.Search":
 
-> myNewEngine = searchEngine "site" "http://site.com/search="
+> myNewEngine = searchEngine "site" "https://site.com/search="
 
    The important thing is that the site has a interface which accepts the escaped query
    string as part of the URL. Alas, the exact URL to feed searchEngine varies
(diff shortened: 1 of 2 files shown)


XMonad/Actions/Search.hs
```diff
@@ -300,11 +290,9 @@ hackage       = searchEngine "hackage"       "http://hackage.haskell.org/package
 hoogle        = searchEngine "hoogle"        "http://hoogle.haskell.org/?hoogle="
 images        = searchEngine "images"        "http://images.google.fr/images?q="
 imdb          = searchEngine "imdb"          "http://www.imdb.com/find?s=all&q="
-isohunt       = searchEngine "isohunt"       "http://isohunt.com/torrents/?ihq="
 lucky         = searchEngine "lucky"         "http://www.google.com/search?btnI&q="
 maps          = searchEngine "maps"          "http://maps.google.com/maps?q="
 mathworld     = searchEngine "mathworld"     "http://mathworld.wolfram.com/search/?query="
-openstreetmap = searchEngine "openstreetmap" "http://gazetteer.openstreetmap.org/namefinder/?find="
```
slotThe, line 307: I'm not sure about the other two, but name finder was effectively replaced by [nominatim](https://nominatim.openstreetmap.org/ui/search.html), which offers much of the same functionality.  It would seem prudent to just replace the URLs, instead of removing the functionality entirely

alternateved: Yes, you are right. I have checked how those URLs changed and there is a replacement for those. I updated them accordingly. 

I also added Github search engine as it may be a good addition. 

I am not sure about isohunt; although there is a couple of pages with that URLs their status may not be legal and/or their URLs might change a lot.

slotThe: Yeah, probably fine leaving isoHunt out of there; not like anyone has complained about it not working up to this point :)

slotThe:
Thank you for the contribution! :tada:

While squashing things down, I noticed one more engine that was faulty (`thesaurus`) and have now replaced that as well.  Also I took the liberty of (in a separate commit) changing all of the URLs from an `http` prefix to `https` because your browser will do that anyways nowadays :)

alternateved:
Oh, yes, I was also thinking about changing `http` to `https` prefix, that's a really good idea. Thanks!

PR #608, opened 2021-09-26
X.A.Prefix: Add easyPrefixArgument

slotThe:
### Description

Adds a function to utilise EZConfig's Emacs style syntax for entering the prefix key.

The syntax has been well-established for use in configuration files; especially in this case it can be a bit awkward having to specify a function that takes a config just for a single keybinding.

Cc: @Fuco1 

### Checklist

        manually, ...) and concluded:
        - I tested it manually in my config and it works.  If anything, we should write some property tests for the EZConfig parser

### DIFF
--- a/XMonad/Actions/Prefix.hs
+++ b/XMonad/Actions/Prefix.hs
@@ -39,6 +39,7 @@ import XMonad
 import XMonad.Util.ExtensibleState as XS
 import XMonad.Util.Paste (sendKey)
 import XMonad.Actions.Submap (submapDefaultWithKey)
+import XMonad.Util.EZConfig (readKeySequence)
 
 {- $usage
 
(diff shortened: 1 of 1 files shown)


XMonad/Actions/Prefix.hs
```diff
@@ -136,6 +143,13 @@ usePrefixArgument prefix conf = conf {
           let binding = prefix conf'
           in M.singleton binding (handlePrefixArg [binding])
 
+-- | Like 'usePrefixArgument', but allow the key to be entered in Emacs
+-- style syntax.  Defaults to @C-u@ when given an invalid key.
+easyPrefixArgument :: LayoutClass l Window => String -> XConfig l -> XConfig l
+easyPrefixArgument prefix conf = case readKeySequence conf prefix of
+  Just [key] -> usePrefixArgument (const key) conf
+  _          -> useDefaultPrefixArgument conf
+
```
slotThe, line 152: I actually think the better change would be to just use this as the implementation of `usePrefixArgument`—using the EZConfig parser seems like a straight upgrade to having to manually extract the mod key from an `XConfig` argument

slotThe: On second thought, I really do think this is the right decision and I've made the appropriate changes

Issue #592, opened 2021-08-13
Add `Description` string to modules


slotThe:
### Problem Description

Right now, only two modules have a Haddock-visible `Description` set: https://xmonad.github.io/xmonad-docs/

It would be nice to change this before 0.17 hits, in order to aid discoverability.  Most of the information needed is already available in appropriately concise form in [X.D.Extending](https://xmonad.github.io/xmonad-docs/xmonad-contrib-0.16.999/XMonad-Doc-Extending.html) and would only need to be copied over.

At the same time, it would be nice to add a template for new modules to CONTRIBUTING that includes this and perhaps drops some fields that no one ever looks at anymore (namely, Portability and Stability).  I would suggest something like 

``` haskell
--------------------------------------------------------------------------------
-- |
-- Module      :  XMonad.Module.Name
-- Description :  Short one-line description
-- Copyright   :  (c) YEAR  person <email>
-- License     :  BSD3-style (see LICENSE)
--
-- Maintainer  :  person
--
-- Summary of the module 
--------------------------------------------------------------------------------
```

PR #609, opened 2021-09-26
Add Description string to modules

slotThe:
### Description

Sigh.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/592

### Checklist

        manually, ...) and concluded: It's docs

### DIFF
--- a/XMonad/Actions/AfterDrag.hs
+++ b/XMonad/Actions/AfterDrag.hs
@@ -1,6 +1,7 @@
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Actions.AfterDrag
+-- Description :  Allows you to add actions dependent on the current mouse drag.
 -- Copyright   :  (c) 2014 Anders Engstrom <ankaan@gmail.com>
 -- License     :  BSD3-style (see LICENSE)
 --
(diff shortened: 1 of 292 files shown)


TheMC47 (approved):
Kudos! I'm curious, was this manual work?

slotThe:
> Kudos! I'm curious, was this manual work?

Less than I was initially anticipating, actually.  The work flow was basically as follows:

  1. For every file that's in X.D.Extending, copy the first sentence in its description, `find-file-at-point`, insert a description after the `Module` keyword; this was done in a single macro and then looped over the file.
  2. Build the Haddocks, find all modules that don't have descriptions (and are hence not in X.D.Extending), copy the names to a buffer, do essentially 1. again, only use the first line in the usage section; this also neatly fit into a macro.
  3. Insert the modules that were missing into X.D.Extending; the format was macroable (using the usage section again for most modules), but inserting the modules into the right place was done manually because I couldn't think of a good way to script it (though it would probably be possible with a little bit more effort).
  4. Fine-tuning—i.e., actually see whether the descriptions accurately describe the modules and are not too long for the Haddocks; this was manually because I get obsessive about that kind of stuff :)

liskin:
Wow, thanks! :tada:

Issue #616, opened 2021-10-05
XMonad.Prompt.RunOrRaise always opens a directory that has the same name as an executable.


yusi1:
### Problem Description
Hello everyone again, I've been messing around with XMonad.Prompt.RunOrRaise and since it can open both directories and executables (in /usr/bin); I've been having an issue where a directory that has the same name as an executable causes the directory to _always_ be shown in a file manager, with the executable _never_ being executed; An example of this is in this scenario: I type `vmware` into the prompt, the `vmware` program has a directory in my home dir that has the same name as the executable in `/usr/bin`, (the prompt shows 2 matches: `vmware` and `vmware/`), if I select `vmware` thunar file manager launches and shows the `vmware` directory, if I select `vmware/` thunar file manager launches and shows the `vmware` directory aswell. Shouldn't `vmware` be the executable and `vmware/` be the dir?? It should not launch in both cases...

#### Example:
- Setup XMonad.Prompt.RunOrRaise...
- Create a directory in the users home directory that has the same name as an executable in `/usr/bin` (for example: "vmware")
- The prompt shows 2 matches: `vmware` and `vmware/` both do the same thing, but they shouldn't
- The directory is *always* opened, with the executable *never* being executed...

### Configuration File

Please include the smallest _full_ configuration file that reproduces
the problem you are experiencing:

```haskell
module Main (main) where

import XMonad
import XMonad.Prompt
import XMonad.Prompt.RunOrRaise

-- Basic XPConfig
myXPConfig :: XPConfig
myXPConfig = def { font = "xft:PragmataPro Mono Liga:Bold:size=18"
                                , bgColor = "black"  
                                , fgColor = "green"
                                , bgHLight = "green"
                                , fgHLight = "black"
                                , borderColor = "#262626"
                                , promptBorderWidth = 1
                                , position = Top    
                                , height = 24   
                                , searchPredicate = fuzzyMatch
                                , sorter = fuzzySort
                  }

-- Main ...
-- Shortcuts for RunOrRaisePrompt ...
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.15
    - [x] With `xmonad-contrib` version 0.16

yusi1:
[...]

PR #617, opened 2021-10-09
X.P.RunOrRaise: Disambiguate directories and executables

slotThe:
##### X.P.RunOrRaise: Disambiguate directories and executables

When there is a directory and an executable of the same name `foo`, we
display `foo`, as well as `foo/` in the prompt, but selecting either one
of them will invariably run or raise the directory, as this test is done
beforehand.

Since directories already carry a trailing forward slash from the shell
prompt's `getShellCompl`, this is easily disambiguated by incorporating
an additional check.

It is duly noted that the same problem also exists for files and
executables.  This may be fixed in a similar way (adding a unique suffix
to files and checking for that), but since this would involve changing
`getShellCompl` and no-one has complained about this so far, it was
deemed "not worth it" for now.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/616

### Checklist

        manually, ...) and concluded: I tested it manually, waiting for confirmation of @yusi1 as well.

    TODO

### DIFF
--- a/XMonad/Prompt/RunOrRaise.hs
+++ b/XMonad/Prompt/RunOrRaise.hs
@@ -22,14 +22,14 @@ module XMonad.Prompt.RunOrRaise
     ) where
 
 import XMonad hiding (config)
-import XMonad.Prelude (liftA2)
+import XMonad.Prelude (isNothing, isSuffixOf, liftA2)
 import XMonad.Prompt
 import XMonad.Prompt.Shell
 import XMonad.Actions.WindowGo (runOrRaise)
 import XMonad.Util.Run (runProcessWithInput)
 
 import Control.Exception as E
-import System.Directory (doesDirectoryExist, doesFileExist, executable, getPermissions)
+import System.Directory (doesDirectoryExist, doesFileExist, executable, findExecutable, getPermissions)
 
 econst :: Monad m => a -> IOException -> m a
 econst = const . return
@@ -60,9 +60,14 @@ open path = io (isNormalFile path) >>= \b ->
             then spawn $ "xdg-open \"" ++ path ++ "\""
             else uncurry runOrRaise . getTarget $ path
     where
-      isNormalFile f = exists f >>= \e -> if e then notExecutable f else return False
-      exists f = or <$> sequence [doesFileExist f,doesDirectoryExist f]
+      isNormalFile f = do
+          notCommand <- isNothing <$> findExecutable f -- not a command (executable in $PATH)
+          exists <- or <$> sequence [doesDirExist f, doesFileExist f]
+          case (notCommand, exists) of
+              (True, True) -> notExecutable f -- not executable as a file in current dir
+              _            -> pure False
       notExecutable = fmap (not . executable) . getPermissions
+      doesDirExist f = ("/" `isSuffixOf` f &&) <$> doesDirectoryExist f
       getTarget x = (x,isApp x)
 
 isApp :: String -> Query Bool


liskin:
> Isn't that files vs. directories vs. executables (of the same name)? I'm a bit confused, but do feel free to suggest ways to make this clearer

Not really, no. The problem is in mixing executable names in $PATH and relative file/dir names in the same list of completions.

In bash, there's no such confusion: https://www.gnu.org/software/bash/manual/html_node/Command-Search-and-Execution.html
When you type "xterm<kbd>↵</kbd>", it always looks for "xterm" in $PATH (after trying aliases, functions and builtins), even if there's an "xterm" executable in the current directory. When you type "x/xterm<kbd>↵</kbd>", there's a slash, so it doesn't search for it and executes it directly.

But in here, we don't have that slash distinction, so we need to disambiguate. But we're not disambiguating between "directories and files and executables", because "executables" are just files, we're disambiguating between "relative paths (files/dirs) and executables/commands in $PATH". It's probably just a wording nitpick, as all you need to do is change "executables" to "executables in $PATH" or "commands in $PATH" (or perhaps just "commands", although that's almost as unclear as "executables").

Anyway, as to whether this is worth dealing with, I'd think that preferring commands in $PATH to both files/dirs in the current directory would be the most natural thing as it's closer in behaviour to a shell. Adding a suffix like the attached patch would probably work, but it's a bit ugly. Can we perhaps refactor the check in RunOrRaise to try https://hackage.haskell.org/package/directory-1.3.6.2/docs/System-Directory.html#v:findExecutable first (you'll need to also check for a slash not being present as findExecutable doesn't do that)? That seems more like what the shell does and what I think might be the expected behaviour. Haven't really thought about what might break, though.

(That being said, absolutely do feel free to merge the fix you have so far, with improved commit message wording, and leave the rest for later.)

slotThe:
> Not really, no. The problem is in mixing executable names in $PATH and relative file/dir names in the same list of completions.

Ah I see; thanks for the explanation.

> Anyway, as to whether this is worth dealing with, I'd think that preferring commands in $PATH to both files/dirs in the current directory would be the most natural thing as it's closer in behaviour to a shell. Adding a suffix like the attached patch would probably work, but it's a bit ugly. Can we perhaps refactor the check in RunOrRaise to try https://hackage.haskell.org/package/directory-1.3.6.2/docs/System-Directory.html#v:findExecutable first (you'll need to also check for a slash not being present as findExecutable doesn't do that)? That seems more like what the shell does and what I think might be the expected behaviour. Haven't really thought about what might break, though.

Since the old code definitely executes the file when it's an executable, I think this check can actually be done up-front without breaking anything; I did that in c83a46a92fe8e77567a97ba87016f71862ccd08f.

The problem with files in the current directory having the same name as something in `$PATH` is that right now they won't even show up in the prompt since we don't show the user paths, but only names of files/directories.  We would need to at least mark these "duplicates" somehow, which would involve messing with X.P.Shell internals and I'm really not up for that right now :)

yusi1:
> > Not really, no. The problem is in mixing executable names in $PATH and relative file/dir names in the same list of completions.
> 
> Ah I see; thanks for the explanation.
> 
> > Anyway, as to whether this is worth dealing with, I'd think that preferring commands in $PATH to both files/dirs in the current directory would be the most natural thing as it's closer in behaviour to a shell. Adding a suffix like the attached patch would probably work, but it's a bit ugly. Can we perhaps refactor the check in RunOrRaise to try https://hackage.haskell.org/package/directory-1.3.6.2/docs/System-Directory.html#v:findExecutable first (you'll need to also check for a slash not being present as findExecutable doesn't do that)? That seems more like what the shell does and what I think might be the expected behaviour. Haven't really thought about what might break, though.
> 
> Since the old code definitely executes the file when it's an executable, I think this check can actually be done up-front without breaking anything; I did that in [c83a46a](https://github.com/xmonad/xmonad-contrib/commit/c83a46a92fe8e77567a97ba87016f71862ccd08f).
> 
> The problem with files in the current directory having the same name as something in `$PATH` is that right now they won't even show up in the prompt since we don't show the user paths, but only names of files/directories. We would need to at least mark these "duplicates" somehow, which would involve messing with X.P.Shell internals and I'm really not up for that right now :)
> 
> EDIT: As for breaking current functionality I'm hoping that @yusi1 will confirm that at least their use-case still works

@slotThe, just tested your fix in a VM and it does exactly what I wanted and fits my usecase, thanks for working on this everyone involved! :smile:

liskin:
> Since the old code definitely executes the file when it's an executable, I think this check can actually be done up-front without breaking anything; I did that in [c83a46a](https://github.com/xmonad/xmonad-contrib/commit/c83a46a92fe8e77567a97ba87016f71862ccd08f).

The old code does, but the new code doesn't always. If you try to run e.g. `bin/something`, the old code would decide it's not a normal file and execute it (but not raise it! `pidof bin/something` won't find the pid). The new code doesn't do the executable check, so it'll invoke `xdg-open bin/something` instead. I have absolutely no idea if this breaks any meaningful workflows, though: stuff in `~/bin` will likely be on the $PATH, and also raising won't work, so…

slotThe:
I'm willing to take that risk :)

liskin:
There was a comment here from @yusi1 saying that this breaks inputs like /usr/bin/firefox, which it definitely does, and since it's a full path, `pidof` accepts that and raising would work before as well. That comment is now lost, or at least I don't see it now. Reintroducing the `getPermissions` check would fix this.

Anyway, I'm happy to merge this as it is, but I do need to say that the commit message is still the same as it was before we fixed the files vs. commands ambiguity, so it now claims that a problem still exists which it no longer does. :-)

slotThe:
> I do need to say that the commit message is still the same as it was before we fixed the files vs. commands ambiguity, so it now claims that a problem still exists which it no longer does. :-)

The problem still exists insofar as what I said above about displaying files in path and files in the current directory:

> The problem with files in the current directory having the same name as something in $PATH is that right now they won't even show up in the prompt since we don't show the user paths, but only names of files/directories. We would need to at least mark these "duplicates" somehow

slotThe:
> > I do need to say that the commit message is still the same as it was before we fixed the files vs. commands ambiguity, so it now claims that a problem still exists which it no longer does. :-)
> 
> The problem still exists insofar as what I said above about displaying files in path and files in the current directory:
> 
> > The problem with files in the current directory having the same name as something in $PATH is that right now they won't even show up in the prompt since we don't show the user paths, but only names of files/directories. We would need to at least mark these "duplicates" somehow

so I figured the commit message still applied

PR #618, opened 2021-10-09
EwmhDesktops: Advertise _NET_WM_STATE_DEMANDS_ATTENTION.

aartamonau:
### Description

XMonad.Hooks.UrgencyHook knows how to understand this hint, but it's
not advertised as supported via _NET_SUPPORTED. This prevents certain
applications (e.g. kitty terminal emulator) from using it.

### Checklist

        manually, ...) and concluded: I ran `stack test` in xmonad-contrib and verified that with the change kitty terminal emulator sets the hint on bell character.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,6 +52,11 @@
       was changed to be closer to the spec. From now these two lists will have
       differently sorted windows.
 
+    - `_NET_WM_STATE_DEMANDS_ATTENTION` was added to the list of supported
+      hints (as per `_NET_SUPPORTED`). This hint has long been understood by
+      `UrgencyHook`. This enables certain applications (e.g. kitty terminal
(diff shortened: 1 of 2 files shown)


geekosaur:
Hm. @liskin, do we perhaps want this to be conditional on use of `UrgencyHook` via your new framework?

liskin:
> Hm. @liskin, do we perhaps want this to be conditional on use of `UrgencyHook` via your new framework?

Definitely, yes. The first step in my refactor is going to be moving all the _NET_SUPPORTED stuff to XMonad.Util.EWMH and exporting an API that individual modules can use to advertise their part of EWMH support.

Actually, we already kind of have that a bit with `addSupported` in EwmhDesktops, it's just not exported and doesn't compose with `setSupported` reliably.

Issue #619, opened 2021-10-11
Delete/Auto-generate `XMonad.Doc.Extending` sections


TheMC47:
### Problem Description

Currently, `XMonad.Doc.Extending` lists every module with its description. The descriptions were moved to the corresponding Haddock field in each module's file in #609 (thanks @slotThe!). The problem: now there are two places that need to be updated when the description is tweaked (I'm looking at you, #613). 

#### Proposal
Let's just remove the list of modules `XMonad.Doc.Extending`. Haddock beautifully generates a grouped list of all modules (https://xmonad.github.io/xmonad-docs/xmonad-contrib-0.16.999/), so having the list in `X.D.Extending` isn't really adding anything.

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

slotThe:
Agreed, let's kill that list with fire

slotThe:
Closed by https://github.com/xmonad/xmonad-contrib/pull/620 (839302533bed6c6ece709a9d8508cc9b773b3975)

PR #613, opened 2021-10-04
Update `X.H.DynamicLog` references

TheMC47:
### Description

Since `X.H.StatusBar` is the newer way to interact with status bars, the `X.H.DynamicLog` references need to be updated.

This is a WIP, as there are many modules to patch

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I've considered how to best test these changes manually and concluded: WIP

  - [ ] I updated the `CHANGES.md` file

  - [ ] I updated the `XMonad.Doc.Extending` file (if appropriate)

PR #620, opened 2021-10-12
`X.D.Extending`: Remove module list

TheMC47:
### Description
As of 38c11c1e3cfa1b91d9af796872e3d011895b4d50, all modules now include
a `Description` string that can be rendered using Haddock. This makes
the list in `XMonad.Doc.Extending` redundant.

Related: https://github.com/xmonad/xmonad-contrib/issues/619

The module still needs a bit of reworks, I'll take care of the DynamicLog/StatusBar bits in #613 

### Checklist

  - [x] I've considered how to best test these changes (Haddock). Yep they're gone alright

  - (N/A) I updated the `CHANGES.md` file 

  - [X] I updated the `XMonad.Doc.Extending` file (you bet I did)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,5 +11,3 @@ behind them.
         manually, ...) and concluded: XXX
 
   - [ ] I updated the `CHANGES.md` file
-
-  - [ ] I updated the `XMonad.Doc.Extending` file (if appropriate)


slotThe:
> We just need to keep checking that new modules set Description.

I suppose the easiest thing would be do change the "I have updated `X.D.Extending`" nit into an `I have added a `Description` field into the module header" nit.

Alternatively we could add a template for this in CONTRIBUTING (though I suppose explaining how to write contrib modules does not really fit into core's CONTRIBUTING...)

Issue #565, opened 2021-06-23
X.A.DynamicWorkspaces freezes system with W.copy and removeWorkspace


ixzh:
### Problem Description
X.A.DynamicWorkspaces freezes system  

### Steps to Reproduce
1.  W.copy a window to a new workspace, say helix.
2.  remove helix workspace
3.  memory surge to 98%, system freeze, require a restart

### Configuration File
```haskell
-- Dynamic workspaces
   , ((modMask .|. shiftMask, xK_BackSpace), removeWorkspace)
   , ((modMask .|. shiftMask, xK_v ), selectWorkspace         myXPConfig)
   , ((modMask .|. shiftMask, xK_n ), withWorkspace           myXPConfig (windows . W.shift))
   , ((modMask .|. shiftMask, xK_m ), withWorkspace           myXPConfig (windows . copy))
   , ((modMask .|. shiftMask, xK_r ), renameWorkspace         myXPConfig)
```

### Checklist

 xmonad 0.16.9999 and xmonad-contrib are git version to date.

liskin:
Can you be more specific about how you "delete" a workspace? There doesn't seem to be workspace deletion in X.A.CycleWS, so I'm a bit confused about the role of CycleWS and what deletion means in this context.

ixzh:
> Can you be more specific about how you "delete" a workspace?

my mystake, it is X.A.DynamicWorkspaces:
```
  , ((modMask .|. shiftMask, xK_BackSpace),                 removeWorkspace)
   , ((modMask .|. shiftMask, xK_v ), selectWorkspace        myXPConfig)
   , ((modMask .|. shiftMask, xK_n ), withWorkspace          myXPConfig (windows . W.shift))
   , ((modMask .|. shiftMask, xK_m ), withWorkspace          myXPConfig (windows . copy))
   , ((modMask .|. shiftMask, xK_r ), renameWorkspace        myXPConfig)
```

liskin:
Oh, right, can you try adding `nub` here: https://github.com/xmonad/xmonad-contrib/blob/6d184e859ced72b8266d5d5b37004119da8f6dd9/XMonad/Actions/DynamicWorkspaces.hs#L255? That should remove the duplicates and fix it.

ixzh:
> Oh, right, can you try adding `nub` here:
> 
[...]

PR #621, opened 2021-10-13
X.A.DynamicWorkspaces: Remove duplicates when melding

slotThe:
Lest we forget to include this in the new release, I took the liberty of packing @liskin's fix up as a PR.  We should probably still the root cause of it all, but I see us not having the time for that in the new few days (mid-October, folks!).  Hence this only refers to the issue as `Related:` and not `Fixes:` (even though it does *fix* the problem, but I figured an open issue would give us some more motivation down the road than a closed one),

### Description

When removing a workspace and distributing its windows, it's important
to remove any duplicates that may be there due to, for example, usage of
X.A.CopyWindow.  X will only draw one of these window, leading to some
artifacts.

Related: https://github.com/xmonad/xmonad-contrib/issues/565

Co-authored-by: Tomas Janousek <tomi@nomi.cz>

### Checklist

        manually, ...) and concluded: @ixzh confirmed that this fixes their issue in https://github.com/xmonad/xmonad-contrib/issues/565

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -737,6 +737,11 @@
     - Removed deprecations for `spacing`, `spacingWithEdge`,
       `smartSpacing`, and `smartSpacingWithEdge`.
 
+  * `XMonad.Actions.DynamicWorkspaces`
+
+    - Fixed a system freeze when using `X.A.CopyWindow.copy` in
+      combination with `removeWorkspace`.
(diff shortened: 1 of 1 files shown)


liskin:
Thanks! I removed #565 from the 0.17 release project for now.

PR #626, opened 2021-10-18
EWMH: Improve interface for custom sorting, filtering, renaming and window activation

### LINKED ISSUE #396: Breaking change in EWMH window activation

liskin:
### Description

Work in progress on #396. “#399 done right but quick.”

Yesterday's #625 is the glorious future, but it's too big a leap at this point in the release cycle. #399 is the ugly past with an ugly interface and no docs. This one is a pragmatic compromise: (mostly) backwards compatible interface extended with just enough ExtensibleConf bits to fix the problems at hand.

Supersedes #399.
Fixes #396.

#### TODO:

  - [x] documentation updates for workspace filtering (scratchpads, etc.)
  - [x] ExtensibleConf configuration for `activateHook`
  - [x] documentation updates for window activation ([examples](https://github.com/xmonad/xmonad-contrib/pull/399#issuecomment-788042938))
  - [x] documentation updates for X.H.Focus
  - [x] (optional) mark individual hooks in EwmhDesktops deprecated but don't trigger deprecation warnings during build of xmonad-contrib itself
  - [x] CHANGES.md
  - [x] review
  - 🚀 merge

And that's it! [Fullscreen hooks are not critical](https://github.com/xmonad/xmonad-contrib/pull/399#issuecomment-807755361) and will be done post-release in #625 or a followup thereof.

### Checklist

  - [x] I've considered how to best test these changes (property, unit, manually, ...) and concluded: the plan is to try to migrate my xmonad config that makes use of most relevant features and hope any bugs are apparent; I'm not expecting any, this should primarily be a change in interface, not in behaviour; the only risk is in the X.H.Focus module, which I'm afraid will need to be accepted

  - [X] I updated the `CHANGES.md` file: existing commits update CHANGES.md, so this part is ready for review

### DIFF
--- a/XMonad/Config/Bluetile.hs
+++ b/XMonad/Config/Bluetile.hs
@@ -197,15 +197,13 @@ bluetileLayoutHook = avoidStruts $ minimize $ boringWindows $
             floatingDeco l = buttonDeco shrinkText defaultThemeWithButtons l
 
 bluetileConfig =
-    docks $
+    docks . ewmhFullscreen . ewmh $
     def
         { modMask = mod4Mask,   -- logo key
           manageHook = bluetileManageHook,
           layoutHook = bluetileLayoutHook,
-          logHook = currentWorkspaceOnTop >> ewmhDesktopsLogHook,
-          handleEventHook = ewmhDesktopsEventHook
-                                `mappend` fullscreenEventHook
-                                `mappend` minimizeEventHook
+          logHook = currentWorkspaceOnTop,
+          handleEventHook = minimizeEventHook
                                 `mappend` serverModeEventHook' bluetileCommands
                                 `mappend` positionStoreEventHook,
           workspaces = bluetileWorkspaces,
--- a/XMonad/Hooks/EwmhDesktops.hs
+++ b/XMonad/Hooks/EwmhDesktops.hs
@@ -40,7 +40,7 @@ module XMonad.Hooks.EwmhDesktops (
     -- $customActivate
     setEwmhActivateHook,
 
-    -- * Standalone hooks (to be deprecated)
+    -- * Standalone hooks (deprecated)
     ewmhDesktopsStartup,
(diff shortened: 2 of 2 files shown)


liskin:
Just one last missing TODO: X.H.Focus docs, so marking as ready to review and will hopefully finish that tomorrow morning.

PR #637, opened 2021-10-30
Start modernising X.D.Extending

slotThe:
### Description

We've discussed this a bit on IRC, but here it goes again: since we have the tutorial, both X.D.Configuring and X.D.Extending lose a bit of their relevance.  In my opinion, the former is completely redundant now (and should perhaps be removed some other time) while the latter could now be used for more "advanced" configuration that still isn't quite writing ones own module.  This is an attempt to kickstart this.

##### X.D.Extending: Link to tutorial instead of the wiki

The configurations on the wiki are quite old and mostly non-functional at this point.  Instead, we have a shiny new tutorial we can refer to. It contains links to actual maintainers configurations (complex they may be, at least they compile).

##### X.D.Extending: Rephrase introduction, update

Since the tutorial is really the better place to start learning how to use xmonad, X.D.Extending can be the place for more "advanced" usage examples.  These would be things like writing small functions to scratch an itch (rather than entire modules).  Rewrite the introduction accordingly.

On the way, remove some small inconsistencies/dead links/remarks about how things are different from version 0.5 onwards—these times are long gone now.

##### X.D.Extending: Explain how to write additionalKeys

Instead of telling the user how to add custom keybindings to xmonad—something which is already done in the tutorial—instead explain instead how one would go about writing a version of X.U.EZConfig.additionalKeys.  This mostly involves looking at the type signatures and sticking some standard functions together, so it's quite a decent way to learn about some of xmonad's internals.

### Checklist

        manually, ...) and concluded: it's docs!

### DIFF
--- a/XMonad/Doc/Extending.hs
+++ b/XMonad/Doc/Extending.hs
@@ -65,11 +65,8 @@ module XMonad.Doc.Extending
     -- * Extending xmonad
     -- $extending
 
-    -- ** Editing key bindings
+    -- ** Adding key bindings
     -- $keys
-
-    -- *** Adding key bindings
-    -- $keyAdding
 
     -- *** Removing key bindings
     -- $keyDel
@@ -141,9 +138,6 @@ In the @XMonad.Actions@ namespace you can find modules exporting
 various functions that are usually intended to be bound to key
 combinations or mouse actions, in order to provide functionality
 beyond the standard keybindings provided by xmonad.
-
-See "XMonad.Doc.Extending#Editing_key_bindings" for instructions on how to
-edit your key bindings.
 
 -}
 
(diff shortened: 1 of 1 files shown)


TheMC47 (approved):
This is great, especially the keybindings part! LGTM :tada:

XMonad/Doc/Extending.hs
```diff
@@ -154,23 +147,23 @@ In the @XMonad.Hooks@ namespace you can find modules exporting
...
   since it lets us examine the new window's properties and act
   accordingly. For instance, we can configure xmonad to put windows
   belonging to a given application in the float layer, not to manage
   dock applications, or open them in a given workspace. See
-  "XMonad.Doc.Extending#Editing_the_manage_hook" for more information on
-  customizing 'XMonad.Core.manageHook'.
+  "XMonad.Doc.Extending#Editing_the_manage_hook" for more information
+  on customizing 'XMonad.Core.manageHook'.
 
 * 'XMonad.Core.logHook': this hook is called when the stack of windows
```
geekosaur, line 159: There is no `XMonad.Core`; wouldn't this be confusing?

XMonad/Doc/Extending.hs
```diff
@@ -154,23 +147,23 @@ In the @XMonad.Hooks@ namespace you can find modules exporting
...
-  managed by xmonad has been changed, by calling the
-  'XMonad.Operations.windows' function. For instance
+  managed by xmonad has been changed; for example, this is invoked at
+  the end of the 'XMonad.Operations.windows' function. For instance
   "XMonad.Hooks.DynamicLog" will produce a string (whose format can be
```
geekosaur, line 162: Do we want to use XMonad.Hooks.StatusBar here?

XMonad/Doc/Extending.hs
```diff
@@ -237,124 +230,161 @@ yourself.
...
+>                  -> XConfig l
+
+looks exactly like what we want!  Now that we know what kind of data
+structure—that is, maps—we are dealing with, the implementation of this
+function simply merges the two together, preferring the user config to
+xmonad's defaults in case of any conflicts.  Thankfully, someone else
+has already done the hard work and written the merging function for us;
+it's called 'Data.Map.Strict.union'.
+
+What's left is essentially playing type tetris:
```
geekosaur, line 325: It's all very well to call it that informally, but it strikes me that (a) it's a bad idea in more formal documentation (b) it may run into copyright / licensing issues.

XMonad/Doc/Extending.hs
```diff
@@ -237,124 +230,161 @@ yourself.
...
+
+The function @mergeKeyList@ take some user config, transforms the custom
+keybindings into a map (@Map.fromList keyList@), gets the keys from the
+base config (remember @keys baseConf@ is again a function, morally of
+type @UserConfig -> Map KeyPress Action@, and so we have to apply @conf@
+to it in order to get a map!), and then merges these two maps together.
+Since @mergeKeylist@ now has exactly the right type signature, we can
+just set that as the keys.
+
+If you like operators, 'Data.Monoid.<>' (or, xmonad's alias for it,
```
geekosaur, line 339: Pedantry: too many commas, I think. "or xmonad's alias for it,".

XMonad/Doc/Extending.hs
```diff
@@ -237,124 +230,161 @@ yourself.
...
+> main :: IO ()
+> main = xmonad $ def
+>   `myAdditionalKeys`
+>     [ ((mod1Mask, xK_m        ), spawn "echo 'Hi, mom!' | dzen2 -p 4")
+>     , ((mod1Mask, xK_BackSpace), spawn "xterm")
+>     ]
+
+Lastly, if you want you can also emulate the automatic modifier
+detection by 'XMonad.Util.EZConfig.additionalKeysP' by defining the bulk
+of your config as separate function
```
geekosaur, line 359: "as a separate"

XMonad/Doc/Extending.hs
```diff
@@ -468,9 +498,8 @@ advanced feature of the Haskell programming language: type classes.
 This allows us to very easily write new layouts, combine or modify
 existing layouts, create layouts with internal state, etc. See
 "XMonad.Doc.Extending#The_LayoutClass" for more information. This
-means that we cannot simply have a list of layouts as we used to have
-before the 0.5 release: a list requires every member to belong to the
-same type!
+means that we cannot simply have a list of layouts; a list requires
```
geekosaur, line 501: ":" instead of ";".

geekosaur (changes requested):
Most of the above are nits, but the copyright issue worries me a bit.

XMonad/Doc/Extending.hs
liskin: This is only broken on Hackage. When the docs are generated together in one `stack` invocation, like they're for https://xmonad.github.io/xmonad-docs/, this link works fine.

XMonad/Doc/Extending.hs
```diff
@@ -349,13 +347,13 @@ provide useful functions for editing your key bindings; "XMonad.Util.EZConfig" e
 {- $keyDel
 #Removing_key_bindings#
 
-Removing key bindings requires modifying the 'Data.Map.Map' which
-stores the key bindings.  This can be done with 'Data.Map.difference'
-or with 'Data.Map.delete'.
+Removing key bindings requires modifying the 'Data.Map.Strict.Map' which
+stores the key bindings.  This can be done with 'Data.Map.difference' or
+with 'Data.Map.Strict.delete'.
```
liskin, line 383: Hm, this is really unfortunate. :-/

> The Map type is shared between the lazy and strict modules, meaning that the same Map value can be passed to functions in both modules. This means that the Functor, Traversable and Data instances are the same as for the Data.Map.Lazy module, so if they are used the resulting maps may contain suspended values (thunks).

(https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Map-Strict.html)

So we really need to link to `Data.Map.Strict.Map` for haddock to produce a good link, but we also don't really want to confuse users into importing `Data.Map.Strict` instead of `Data.Map`. :-/

And then these links don't work with xmonad-docs anyway, as the base haddocks aren't there. :-/

There's probably no good way to do this.

XMonad/Doc/Extending.hs
```diff
@@ -666,11 +663,11 @@ And then we can add @myManageHook@ to the default one to create
 One more thing to note about this system is that if
 a window matches multiple rules in a 'XMonad.Config.manageHook', /all/
 of the corresponding actions will be run (in the order in which they
-are defined).  This is a change from versions before 0.5, when only
-the first rule that matched was run.
+are defined).  An alternative version where only the first rule that
+matches is run is available as 'XMonad.Hooks.ManageHelpers.composeOne'.
```
liskin, line 698: :+1:!

XMonad/Doc/Extending.hs
liskin: This whole section is somewhat informal though, so "type tetris" doesn't stand out too much.
Could you elaborate a bit about the copyright/licensing issue? Are you suggesting that one cannot use the word tetris in a documentation? \[citation needed\] :-)

XMonad/Doc/Extending.hs
```diff
@@ -236,111 +230,148 @@ yourself.
...
+has already done the hard work and written the merging function for us;
+it's called 'Data.Map.Strict.union'.
+
+What's left is essentially playing type tetris:
+
+> myAdditionalKeys baseConf keyList =
+>   let mergeKeylist conf = Map.fromList keyList `Map.union` (keys baseConf) conf
+>    in baseConf { keys = mergeKeylist }
+
+The function @mergeKeyList@ take some user config, transforms the custom
```
liskin, line 331: But why does it take a user config at all? I think this question should be answered as soon as the `myKeys :: UserConfig -> Map KeyPress Action` example. I mean, I know the answer, yet I'm still scratching my head here! :-)

XMonad/Doc/Extending.hs
slotThe: Users importing `Data.Map.Strict` isn't so bad, actually.  The base data structures (`Data.Map.Lazy.Map` and `Data.Map.Strict.Map`, that is) are actually the same; the functions in the stricd module just make sure to force things before handling them.  I don't think we're really relying on the lazyness of `Data.Map` anywhere in xmonad.  The fact that the links don't work, however, is not good :/

XMonad/Doc/Extending.hs
geekosaur: Not sure I can point to specific law, but trademarks and in particular registered trademarks have to be enforced to not be lost in the U.S. and probably other countries. See "Kleenex" for one that's in danger of being turned over, since hardly anyone calls them "tissues". Which means we risk being contacted by a lawyer if we use it.

XMonad/Doc/Extending.hs
slotThe: We probably want to rewrite the whole `logHook` section to instead refer to X.H.StatusBar, but just switching the names won't do (because the functions referred to in these sections are specifically functions from X.H.DynamicLog)

XMonad/Doc/Extending.hs
liskin: `f conf = … (modMask conf) …` is shorter but not sure if good for what you want to say. No strong opinions here, I'll leave it up to you to decide which sounds better.

XMonad/Doc/Extending.hs
liskin: True, it's not bad. I did find it confusing as the code uses `Data.Map` everywhere. Maybe it doesn't matter.

Fixing the links may involve some `sed` hackery and can be somewhat brittle, so I'd deprioritize it somehow (I do need a break, after all).

XMonad/Doc/Extending.hs
liskin: According to https://desiree47.wordpress.com/2008/12/07/tetris-and-intellectual-property-rights-trademark/amp/:

> Trademark rights prevent outside vendors from using identical or “confusingly similar” marks on their merchandise to sell their products (USPTO).
> […]
> The Tetris Company has obtained registered trademarks for the use of the word TETRIS and the Tetris logo in the industries of:
>
> “games and playthings; namely action figures, bagatelle games; video games and video game cartridges; electronic game equipment for playing video games; electronic game programs; electronic toys with a clock or timepiece feature; hand held electronic toys; non-electronic hand held games; player operated electronic controllers for electronic video game machines; toy figures mountable on pencils and pens,”
>
> “video game cartridges and software for playing video games,”
>
> “clothing, namely shirts, hats and t-shirts,”and “hand held video games”.

We're not using it for marketing/sales purposes, and we're not in any of these industries, so I'd say it should be fine, but IANAL.

XMonad/Doc/Extending.hs
slotThe: It is shorter indeed, though I like the fact that `XConfig{modMask=modMask}` makes it quite clear we are _only_ interested in that field (I just realized that we can even drop the `conf@` in these examples)).  It probably does not matter much in the end.  The core message that it would be nice if the keybindings knew things about the config will probably come across either way

XMonad/Doc/Extending.hs
slotThe: I mean, there's nothing stopping us from rewriting the docs (and the code, eventually) to use `Data.Map.Strict` instead of `Data.Map`

liskin: Which won't fix the links. ;-)

XMonad/Doc/Extending.hs
slotThe: c8417409fcdd6ef3dd02a1d92e7b17edd13b0fc6?

XMonad/Doc/Extending.hs
```diff
@@ -237,124 +230,171 @@ yourself.
...
+>   myKeys :: XConfig l -> Map (ButtonMask, KeySym) (X ())
+>   myKeys conf = Map.fromList
+>     [ ((mod1Mask    , xK_m        ), spawn "echo 'Hi, mom!' | dzen2 -p 4")
+>     , ((modMask conf, xK_BackSpace), spawn "xterm")
+>     ]
+
+Now, obviously we don't want to do that; we only want to add to existing
+bindings (or, perhaps, override some of them with our own).  Let's break
+@myKeys@ down a little.  You can think of the type signature of @myKeys@
+(and hence also of @key@) like this:
```
liskin, line 279:
```suggestion
(and hence also of @keys@) like this:
```

liskin (approved):

XMonad/Doc/Extending.hs
slotThe: So, at this point, what's the general feeling here: keep/delete?

liskin: My vote is "keep". You wouldn't even be the first to use "type tetris".

XMonad/Doc/Extending.hs
slotThe: I created https://github.com/xmonad/xmonad-contrib/issues/645 to keep track of this

PR #639, opened 2021-11-01
Test against GHC 9.2.1; fix new warnings

liskin:
A lot of ugliness. We may want to disable the warnings instead. :-/

Related: https://github.com/xmonad/xmonad/pull/346

### Checklist

  - [ ] I've considered how to best test these changes (property, unit, manually, ...) and concluded: I'll think about it later, but I admit some places may need some testing as I did change the code in non-trivial manner

### DIFF
--- a/XMonad/Actions/CycleWindows.hs
+++ b/XMonad/Actions/CycleWindows.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE ViewPatterns #-}
+
 --------------------------------------------------------------------------------
 -- |
 -- Module       : XMonad.Actions.CycleWindows
@@ -53,7 +55,9 @@ module XMonad.Actions.CycleWindows (
 ) where
 
 import XMonad
+import XMonad.Prelude
 import qualified XMonad.StackSet as W
+import qualified Data.List.NonEmpty as NE
 import XMonad.Actions.RotSlaves
 
 import Control.Arrow (second)
@@ -179,7 +183,7 @@ rotOpposite' :: W.Stack a -> W.Stack a
 rotOpposite' (W.Stack t l r) = W.Stack t' l' r'
   where rrvl = r ++ reverse l
         part = (length rrvl + 1) `div` 2
-        (l',t':r') =  second reverse . splitAt (length l) $
+        (l', notEmpty -> t' :| r') = second reverse . splitAt (length l) $
                                 reverse (take part rrvl ++ t : drop part rrvl)
 
 
@@ -205,7 +209,7 @@ rotFocusedDown = windows . W.modify' $ rotFocused' rotDown
 rotFocused' :: ([a] -> [a]) -> W.Stack a -> W.Stack a
 rotFocused' _ s@(W.Stack _ [] []) = s
 rotFocused' f   (W.Stack t [] (r:rs)) = W.Stack t' [] (r:rs') -- Master has focus
-    where (t':rs') = f (t:rs)
+    where (notEmpty -> t' :| rs') = f (t:rs)
 rotFocused' f s@W.Stack{} = rotSlaves' f s                    -- otherwise
 
 
@@ -223,7 +227,7 @@ rotUnfocused' :: ([a] -> [a]) -> W.Stack a -> W.Stack a
 rotUnfocused' _ s@(W.Stack _ [] []) = s
 rotUnfocused' f s@(W.Stack _ [] _ ) = rotSlaves' f s                 -- Master has focus
 rotUnfocused' f   (W.Stack t ls rs) = W.Stack t (reverse revls') rs' -- otherwise
-    where  (master:revls)  = reverse ls
+    where  (master :| revls) = NE.reverse (let l:ll = ls in l :| ll)
            (revls',rs') = splitAt (length ls) (f $ master:revls ++ rs)
 
 -- $generic
--- a/XMonad/Actions/FocusNth.hs
+++ b/XMonad/Actions/FocusNth.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE ViewPatterns #-}
+
 -----------------------------------------------------------------------------
 -- |
 -- Module       : XMonad.Actions.FocusNth
@@ -18,8 +20,9 @@ module XMonad.Actions.FocusNth (
                  focusNth,focusNth',
                  swapNth,swapNth') where
 
-import XMonad.StackSet
 import XMonad
+import XMonad.Prelude
+import XMonad.StackSet
 
 -- $usage
 -- Add the import to your @~\/.xmonad\/xmonad.hs@:
@@ -40,8 +43,8 @@ focusNth :: Int -> X ()
 focusNth = windows . modify' . focusNth'
 
 focusNth' :: Int -> Stack a -> Stack a
-focusNth' n s@(Stack _ ls rs) | (n < 0) || (n > length ls + length rs) = s
-                              | otherwise = listToStack n (integrate s)
+focusNth' n s | n >= 0, (ls, t:rs) <- splitAt n (integrate s) = Stack t (reverse ls) rs
+              | otherwise = s
 
 -- | Swap current window with nth. Focus stays in the same position
 swapNth :: Int -> X ()
@@ -50,11 +53,5 @@ swapNth = windows . modify' . swapNth'
 swapNth' :: Int -> Stack a -> Stack a
 swapNth' n s@(Stack c l r)
   | (n < 0) || (n > length l + length r) || (n == length l) = s
-  | n < length l = let (nl, nc:nr) = splitAt (length l - n - 1) l in Stack nc (nl ++ c : nr) r
-  | otherwise    = let (nl, nc:nr) = splitAt (n - length l - 1) r in Stack nc l (nl ++ c : nr)
-
-listToStack :: Int -> [a] -> Stack a
-listToStack n l = Stack t ls rs
- where
-    (t:rs)    = drop n l
-    ls        = reverse (take n l)
+  | n < length l = let (nl, notEmpty -> nc :| nr) = splitAt (length l - n - 1) l in Stack nc (nl ++ c : nr) r
+  | otherwise    = let (nl, notEmpty -> nc :| nr) = splitAt (n - length l - 1) r in Stack nc l (nl ++ c : nr)
--- a/XMonad/Actions/Navigation2D.hs
+++ b/XMonad/Actions/Navigation2D.hs
@@ -824,8 +824,7 @@ doSideNavigationWithBias bias dir (cur, rect)
     rHalfPiCC r = SideRect (-y2 r) (-y1 r) (x1 r) (x2 r)
 
     -- Apply the above function until d becomes synonymous with R (wolog).
-    rotateToR d = let (_, _:l) = break (d ==) [U, L, D, R]
-                  in  foldr (const $ (.) rHalfPiCC) id l
+    rotateToR d = fromJust . lookup d . zip [R, D, L, U] . iterate rHalfPiCC
 
     transform = rotateToR dir . translate . toSR
 
--- a/XMonad/Actions/RotSlaves.hs
+++ b/XMonad/Actions/RotSlaves.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE ViewPatterns #-}
+
 -----------------------------------------------------------------------------
 -- |
 -- Module       : XMonad.Actions.RotSlaves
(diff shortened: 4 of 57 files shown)


liskin:
9e30631 updates hlint, so I'll need to fix the new hints as well.

TheMC47:
Out of curiosity, what's `-Woperator-whitespace-ext-conflict`? I googled it but nothing came up

liskin:
> Out of curiosity, what's `-Woperator-whitespace-ext-conflict`? I googled it but nothing came up

https://downloads.haskell.org/~ghc/9.2.1/docs/html/users_guide/using-warnings.html#ghc-flag--Woperator-whitespace-ext-conflict

XMonad/Layout/BinarySpacePartition.hs
```diff
@@ -425,13 +425,14 @@ resizeSplit :: Direction2D -> (Rational,Rational) -> Zipper Split -> Maybe (Zipp
...
   Nothing -> Just z
-  Just (t, crumb) -> Just $ case dir of
+  Just (t@Node{}, crumb) -> Just $ case dir of
     R -> (t{value=sp{ratio=scaleRatio (ratio sp) xsc}}, crumb)
     D -> (t{value=sp{ratio=scaleRatio (ratio sp) ysc}}, crumb)
     L -> (t{value=sp{ratio=1-scaleRatio (1-ratio sp) xsc}}, crumb)
     U -> (t{value=sp{ratio=1-scaleRatio (1-ratio sp) ysc}}, crumb)
     where sp = value t
           scaleRatio r fac = min 0.9 $ max 0.1 $ r*fac
+  Just (Leaf{}, _) -> undefined
```
slotThe, line 435: Is there any harm in just returning `Nothing` here?  If the `undefined` stays, I think it would be worth adding the commit message as a comment there, so people will know this is intentional

XMonad/Prelude.hs
```diff
@@ -55,3 +60,11 @@ chunksOf i xs = chunk : chunksOf i rest
 -- > f .: g ≡ (f .) . g ≡ \c d -> f (g c d)
 (.:) :: (a -> b) -> (c -> d -> a) -> c -> d -> b
 (.:) = (.) . (.)
+
+-- | 'Data.List.NonEmpty.fromList' with a better error message. Useful to
+-- silence GHC's Pattern match(es) are non-exhaustive warning in places where
+-- the programmer knows it's always non-empty, but it's infeasible to express
+-- that in the type system.
+nonEmpty :: HasCallStack => [a] -> NonEmpty a
```
slotThe, line 68: This confused me when reading through the code top-to-bottom, so I'll mention it: `Data.List.NonEmpty.nonEmpty` already exists and returns a `Maybe`, making some of these view patterns really confusing to read if one has that function in mind

XMonad/Actions/FocusNth.hs
```diff
@@ -50,11 +53,11 @@ swapNth = windows . modify' . swapNth'
...
+  | n < length l = let (nl, nonEmpty -> nc :| nr) = splitAt (length l - n - 1) l in Stack nc (nl ++ c : nr) r
+  | otherwise    = let (nl, nonEmpty -> nc :| nr) = splitAt (n - length l - 1) r in Stack nc l (nl ++ c : nr)
 
 listToStack :: Int -> [a] -> Stack a
 listToStack n l = Stack t ls rs
  where
-    (t:rs)    = drop n l
-    ls        = reverse (take n l)
+    (nonEmpty -> t :| rs) = drop n l
+    ls                    = reverse (take n l)
```
slotThe, line 63: Some of these make me pause for a little bit.  In the docs for `nonEmpty` we write 

> places where the programmer knows it's always non-empty, but it's infeasible to express that in the type system

but this is not really the case here, right?  There's nothing stopping me from calling `listToStack 42 [0, 1]` and have this blow up, it's just an invariant the author hopefully kept in mind while writing the module.  I feel like these kind of occurences should be rewritten with some sort of failure case in mind instead

slotThe (approved):
I like this.  I don't think we should silence `-Wincomplete-uni-patterns` in so many modules; it's really quite useful most of the time.  If that means we'll have to live with some additional glue code that is kind of ugly sometimes then so be it

XMonad/Layout/BinarySpacePartition.hs
liskin: The point of the undefined is to silence the warning and to produce an error message annotated with source code location if the surrounding code changes in a way that this can match. Returning Nothing may have consequences depending on how the surrounding code changes, and it's hard to foresee what consequences (although maybe if I looked more closely, which I won't, because my hand hurts terribly today). A comment would be appropriate indeed.

XMonad/Prelude.hs
liskin: Good point!

I wanted to change this to `notEmpty`, which looks a bit more like an assertion, which it is, but then forgot. Suggestions for better name welcome.

XMonad/Actions/FocusNth.hs
liskin: Or maybe just turn it into a local binding instead (less likely to be used in an unintended way). Or use some `X.U.Stack` functions. I admit I was quite tired when I got to this module and decided to just take a shortcut. :-/

slotThe: Alright, I'll go through all of these view patterns again today (hopefully, anyways) to see which ones we might need to rewrite/make into local functions

XMonad/Actions/CycleWindows.hs
```diff
@@ -205,7 +209,7 @@ rotFocusedDown = windows . W.modify' $ rotFocused' rotDown
 rotFocused' :: ([a] -> [a]) -> W.Stack a -> W.Stack a
 rotFocused' _ s@(W.Stack _ [] []) = s
 rotFocused' f   (W.Stack t [] (r:rs)) = W.Stack t' [] (r:rs') -- Master has focus
-    where (t':rs') = f (t:rs)
+    where (nonEmpty -> t' :| rs') = f (t:rs)
```
slotThe, line 212: `f` is an arbitrary function, so I guess this technically isn't safe.  Since the function is exported we should probably just return the original stack in case of an empty list here

XMonad/Actions/Navigation2D.hs
```diff
@@ -824,7 +824,7 @@ doSideNavigationWithBias bias dir (cur, rect)
     rHalfPiCC r = SideRect (-y2 r) (-y1 r) (x1 r) (x2 r)
 
     -- Apply the above function until d becomes synonymous with R (wolog).
-    rotateToR d = let (_, _:l) = break (d ==) [U, L, D, R]
+    rotateToR d = let (_, nonEmpty -> _ :| l) = break (d ==) [U, L, D, R]
```
slotThe, line 827: This works, but perhaps we could just do `drop 1 $ dropWhile (d /=) [U, L, D, R]` to get rid of `nonEmpty` entirely here

XMonad/Actions/RotSlaves.hs
```diff
@@ -49,8 +52,8 @@ rotSlaves' :: ([a] -> [a]) -> Stack a -> Stack a
 rotSlaves' _ s@(Stack _ [] []) = s
 rotSlaves' f   (Stack t [] rs) = Stack t [] (f rs)                -- Master has focus
 rotSlaves' f s@(Stack _ ls _ ) = Stack t' (reverse revls') rs'    -- otherwise
-    where  (master:ws)     = integrate s
-           (revls',t':rs') = splitAt (length ls) (master:f ws)
+    where  (nonEmpty -> master :| ws)      = integrate s
+           (revls', nonEmpty -> t' :| rs') = splitAt (length ls) (master:f ws)
```
slotThe, line 56: `f` is an argument so it could be anything; again since this is exported it may be worth just returning the original stack on error here

XMonad/Actions/RotSlaves.hs
```diff
@@ -60,4 +63,4 @@ rotAllDown = windows $ modify' (rotAll' (\l -> last l : init l))
 -- | The actual rotation, as a pure function on the window stack.
 rotAll' :: ([a] -> [a]) -> Stack a -> Stack a
 rotAll' f s = Stack r (reverse revls) rs
-    where (revls,r:rs) = splitAt (length (up s)) (f (integrate s))
+    where (revls, nonEmpty -> r :| rs) = splitAt (length (up s)) (f (integrate s))
```
slotThe, line 66: Same as above

XMonad/Actions/Workscreen.hs
```diff
@@ -90,7 +93,7 @@ viewWorkscreen wscrId = do (WorkscreenStorage c a) <- XS.get
                            let wscr = if wscrId == c
                                           then Workscreen wscrId $ shiftWs (workspaces $ a !! wscrId)
                                           else a !! wscrId
-                               (x,_:ys) = splitAt wscrId a
+                               (x, nonEmpty -> _ :| ys) = splitAt wscrId a
```
slotThe, line 96: Since `wscrId` is just some random Int, this is probably not safe either.  It's not any less unsafe than the rest of the function though (see the `if-then-else` above) so...

XMonad/Layout/TallMastersCombo.hs
```diff
@@ -322,12 +322,10 @@ swapWindow w s =
...
       downLst = down s
   in if w `elem` downLst
-     then let us   = takeWhile (/= w) downLst
-              d:ds = dropWhile (/= w) downLst
+     then let (us, nonEmpty -> d :| ds) = break (== w) downLst
               us'  = reverse us ++ d : upLst
           in  Stack foc us' ds
-     else let ds   = takeWhile (/= w) upLst
-              u:us = dropWhile (/= w) upLst
+     else let (ds, nonEmpty -> u :| us) = break (== w) upLst
```
slotThe, line 328: The first `nonEmpty` is safe but I don't think we know that `w` is in `upLst` if it's not in `downLst`

slotThe (commented):
I glossed over the `Config` files, but everything else should've been looked at

XMonad/Actions/CycleWindows.hs
liskin: `f` being arbitrary, this can blow up in other ways, like losing windows, for example, which can happen silently and there's nothing we can do about it even if we wanted to, because there's no `Eq`/`Ord` here. :-/
I'd personally prefer to be verbose at least in the case of `f` returning `[]`, though. If someone passes a function that does more than just a reordering, they should be told.

XMonad/Actions/Navigation2D.hs
liskin: I ended up rewriting it:
```haskell
rotateToR d = fromJust . lookup d . zip [R, D, L, U] . iterate rHalfPiCC
```
Like elsewhere, I think a verbose blowup is better than silent misbehaviour.

XMonad/Actions/Workscreen.hs
liskin: The intended usage of this is to bind it to a key, so it's completely acceptable that a keybinding raises an exception which is caught and logged, and then everything continues as if nothing happened.

XMonad/Layout/BinarySpacePartition.hs
liskin: a41f168de7063739d547aaa3e81c7b351056344b

XMonad/Actions/FocusNth.hs
liskin: 673ef6ef3807d8e369751d07ef66435c842d6003

XMonad/Actions/Navigation2D.hs
liskin: 3bfbd2239d543a6419227fe38ce79678ea442d2a

XMonad/Layout/TallMastersCombo.hs
liskin: Indeed!

427316b58dd423da049e675478ae8c0cf01b1816

XMonad/Prelude.hs
liskin: fe79488cf0af04cf1631145c5268901be6a003e4

XMonad/Actions/Workscreen.hs
slotThe: That's fair enough

XMonad/Actions/CycleWindows.hs
liskin: May I consider this and the other two "arbitrary `f`" comments dealt with or would you still like a silent fallback rather than an exception?

slotThe: It's probably fine.  Not like anyone has ever complained about these functions blowing up so it doesn't seem to be a problem (and if it does turn out to be you're right that exceptions will make this easier to debug)

Issue #316, opened 2019-10-02
X.L.NoBorders flashes border when switching workspaces


sporkm:
### Problem Description

Whenever I switch from one workspace to another, if the workspace I'm switching from is using a layout modified with `noBorders`, there will be a border flash (reverting to the border width set through XConfig) for a fraction of a second before it switches to the other workspace and applies its layout.

This happens even if I'm not using a compositor like compton with fading effects.

Expected behavior: no border flash, simply switch to the new layout and apply it.

### Configuration File

```haskell
module Main (main) where

import XMonad
import XMonad.Layout.NoBorders

main :: IO ()
main = xmonad $ def
        { borderWidth = 2
        , workspaces = ["one", "two"]
        , layoutHook = noBorders Full
        }
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [x] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

apvodney:
This happens to me as well, but when alt-tabbing between windows in the Full layout with noBorders.

slotThe:
I believe this is fixed by [X.L.VoidBorders](https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/XMonad-Layout-VoidBorders.html), so closing this for now

d3adb5:
> I believe this is fixed by [X.L.VoidBorders](https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/XMonad-Layout-VoidBorders.html), so closing this for now

Not quite. X.L.VoidBorders will completely disable borders on whatever window happens to go into a layout it modifies. I made it that way because the windows I would normally put in a workspace using a layout without borders will normally be kept there forever.

If you use, say, `voidBorders Full ||| Full` for your layout hook, any window you display using the first layout there will have no borders for the rest of its lifetime, even if you switch to a normal `Full` layout. Only a screen refresh (or a call to `setWindowBorderWidth`) will restore that window's borders. `voidBorders` doesn't take it upon itself to restore borders when a window "leaves" the layout, and XMonad isn't setting initial properties every time a layout is changed, which is a good thing.

I think this issue can be reopened, if there aren't any duplicates or similar issues still open.

viriuwu:
[...]

PR #640, opened 2021-11-01
Add X.U.ActionQueue And X.H.BorderPerWindow

subbyte:
### Description

Here's a functionality that no existing extension covers: default border width is 4; a user wants `firefox` and `chromium` to have 0 border and `feh` to have border width 16. The border width is permanent and no need to change when switching between layouts. And the user wants to avoid any unnecessary visual effects, e.g., [window jumping](https://github.com/xmonad/xmonad-contrib/issues/316).

12 years ago, users first discussed this need [here](https://xmonad.haskell.narkive.com/130DKJDZ/how-to-make-a-window-floating-above-all-the-other-windows-and-border-less). Unfortunately, no working solution was provided. At the end of the discussion, _Magicloud_ said "the toggleBorder part does not work".

Recently I developed the same need, and asked the question in stackoverflow [Config No Border For Specific Windows in Xmonad](https://stackoverflow.com/questions/69725874/config-no-border-for-specific-windows-in-xmonad). After a few days, I figured out why _Magicloud_ said `ManageHook` does not work (detailed in [my answer](https://stackoverflow.com/a/69774944/8418198)) and wrote this PR to help others solve the same problem.

I put _why I think this is the best way to realize the functionality_ and _why there is no existing extension that implements exactly this_ in the docstring of the source code---the `$design` and `$options` sections.

### Checklist

        manually, ...) and concluded: I tested the new module manually to verify its behaviors.

### DIFF
--- a/XMonad/Hooks/BorderPerWindow.hs
+++ b/XMonad/Hooks/BorderPerWindow.hs
@@ -1,0 +1,78 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Hooks.BorderPerWindow
+-- Description :  Set border width for each window in all layouts.
+-- Copyright   :  (c) 2021 Xiaokui Shu
+-- License     :  BSD-style (see LICENSE)
+--
+-- Maintainer  :  subbyte@gmail.com
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- Want to customize border width, for each window on all layouts? Want
+-- specific window have no border on all layouts? Try this.
+-----------------------------------------------------------------------------
+
+module XMonad.Hooks.BorderPerWindow ( -- * Usage
+                                      -- $usage
+                                      defineBorderWidth
+                                    , actionQueue
+
+                                      -- * Design Considerations
+                                      -- $design
+                                    ) where
+
+
+import Graphics.X11.Xlib (Dimension)
+import XMonad
+import XMonad.Util.ActionQueue (enqueue, actionQueue)
+
+-- $usage
+--
+-- To use this module, first import it
+--
+-- > import XMonad.Hooks.BorderPerWindow (defineBorderWidth, actionQueue)
+--
+-- Then specify which window to customize the border of in your
+-- @manageHook@:
+--
+-- > myManageHook :: ManageHook
+-- > myManageHook = composeAll
+-- >     [ className =? "firefox"  --> defineWindowWidth 0
+-- >     , className =? "Chromium" --> defineWindowWidth 0
+-- >     , isDialog                --> defineWindowWidth 8
+-- >     ]
+--
+-- Finally, add the 'actionQueue' combinator and @myManageHook@ to your
+-- config:
+--
+-- > main = xmonad $ actionQueue $ def
+-- >     { ...
+-- >     , manageHook = myManageHook
+-- >     , ...
+-- >     }
+--
+-- Note that this module is incompatible with other ways of changing
+-- borders, like "XMonad.Layout.NoBorders".  This is because we are
+-- changing the border exactly /once/ (when the window first appears)
+-- and not every time some condition is satisfied.
+
+-- $design
+--
+-- 1. Keep it simple. Since the extension does not aim to change border setting
+--    when layout changes, only execute the border setting function once to
+--    avoid potential window flashing/jumping/scaling.
+--
+-- 2. The 'ManageHook' eDSL is a nice language for specifying windows. Let's
+--    build on top of it and use it to specify window to define border.
+
+defineBorderWidth :: Dimension -> ManageHook
+defineBorderWidth bw = do
+    w <- ask
+    liftX . enqueue $ updateBorderWidth w bw
+    idHook
+
+updateBorderWidth :: Window -> Dimension -> X ()
+updateBorderWidth w bw = do
+    withDisplay $ \d -> io $ setWindowBorderWidth d w bw
+    refresh
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,11 @@
     Layout modifier that, if only a single window is on screen, places that window
     in the middle of the screen.
 
+  * `XMonad.Util.ActionQueue`
+
+    Put XMonad actions in the queue to be executed every time the
+    `logHook` (or, alternatively, a hook of your choice) runs.
+
 ### Bug Fixes and Minor Changes
 
   * `XMonad.Prompt`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,11 @@
     Put XMonad actions in the queue to be executed every time the
     `logHook` (or, alternatively, a hook of your choice) runs.
 
+  * `XMonad.Hooks.BorderPerWindow`
+
+    While XMonad provides config to set all window borders at the same
+    width, this extension defines and sets border width for each window.
+
 ### Bug Fixes and Minor Changes
 
   * `XMonad.Prompt`
--- a/XMonad/Util/ActionQueue.hs
+++ b/XMonad/Util/ActionQueue.hs
@@ -1,0 +1,99 @@
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Util.ActionQueue
+-- Description :  Queue of XMonad actions
+-- Copyright   :  (c) 2021 Xiaokui Shu
+-- License     :  BSD-style (see LICENSE)
+--
+-- Maintainer  :  subbyte@gmail.com
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- Put XMonad actions in the queue to be executed in either the
+-- @logHook@ or another hook of your choice.
+-----------------------------------------------------------------------------
+
+module XMonad.Util.ActionQueue ( -- * Usage
+                                 -- $usage
+                                 ActionQueue
+                               , actionQueue
+                               , enqueue
+                               , exequeue
+                               ) where
+
+import XMonad
+import qualified XMonad.Util.ExtensibleConf  as XC
+import qualified XMonad.Util.ExtensibleState as XS
+
+import Data.Sequence (Seq (..), ViewL (..), viewl, (|>))
+
+-- $usage
+--
+-- This module provides a queue that, by default, gets executed every
+-- time the @logHook@ runs.  To use this module
+--
+-- 1. Enqueue `X ()` actions at the place you need; e.g.:
+--
+-- > enqueue myAction
+--
+-- 2. Add the 'actionQueue' combinator to your configuration:
+--
+-- > main = xmonad $ actionQueue $ def
+-- >     { ... }
+--
+-- This will execute all of the actions in the queue (if any) every time
+-- the @logHook@ runs.  Developers of other extensions using this module
+-- should re-export 'actionQueue'.
+--
+-- Alternatively, you can directly add 'exequeue' to a hook of your choice.
+-- This is discouraged when writing user-facing modules, as (accidentally)
+-- adding 'exequeue' to two different hooks might lead to undesirable
+-- behaviour.  'actionQueue' uses the "XMonad.Util.ExtensibleConf" interface to
+-- circumvent this.
+--
+
+newtype ActionQueue = ActionQueue (Seq (X ()))
+  deriving newtype (Semigroup)
+
+-- NOTE: The 'Semigroup' instance is technically wrong for a queue
+-- (simply appending two queues together).  However, since these are
+-- only needed for essentially the unit element and multiplication with
+-- the unit, this does not seem like a big deal.
+--
+-- Indeed, since 'ActionQueue' is exported as an abstract type, there is
+-- no way to build up a queue outside os using 'enqueue', meaning users
+-- can't abuse these.
+
+instance ExtensionClass ActionQueue where
+    initialValue = emptyQueue
+
+emptyQueue :: ActionQueue
+emptyQueue = ActionQueue mempty
+
+-- | Every time the @logHook@ runs, execute all actions in the queue.
+actionQueue :: XConfig l -> XConfig l
+actionQueue = XC.once (\cfg -> cfg{ logHook = logHook cfg <> exequeue })
+                      emptyQueue
+
+-- | Enqueue an action.
+enqueue :: X () -> X ()
+enqueue = XS.modify . go
+  where
+    go :: X () -> ActionQueue -> ActionQueue
+    go a (ActionQueue as) = ActionQueue $ as |> a
+
+-- | Execute every action in the queue.
+exequeue :: X ()
+exequeue = do
+    -- Note that we are executing all actions one by one.  Otherwise, we may
+    -- not execute the actions in the right order.  Any of them may call
+    -- 'refresh' or 'windows', which triggers the logHook, which may trigger
+    -- 'exequeue' again if it is used in the logHook.
+    ActionQueue aas <- XS.get
+    case viewl aas of
+      EmptyL  -> pure ()
+      a :< as -> do XS.put (ActionQueue as)
+                    a `catchX` pure ()
+                    exequeue
--- a/XMonad/Util/ExtensibleConf.hs
+++ b/XMonad/Util/ExtensibleConf.hs
@@ -62,7 +62,7 @@ import qualified Data.Map as M
 -- > newtype MyConf = MyConf{ fromMyConf :: [Int] } deriving Semigroup
 -- >
 -- > customLogger :: Int -> XConfig l -> XConfig l
--- > customLogger i = XC.once (MyConf [i]) $ \c -> c{ logHook = logHook c <> lh }
+-- > customLogger i = XC.once (\c -> c{ logHook = logHook c <> lh }) (MyConf [i])
 -- >   where
 -- >     lh :: X ()
 -- >     lh = XC.with $ io . print . fromMyConf
--- a/xmonad-contrib.cabal
+++ b/xmonad-contrib.cabal
@@ -335,6 +335,7 @@ library
                         XMonad.Prompt.XMonad
                         XMonad.Prompt.Zsh
                         XMonad.Util.ActionCycle
+                        XMonad.Util.ActionQueue
                         XMonad.Util.ClickableWorkspaces
                         XMonad.Util.Cursor
                         XMonad.Util.CustomKeys
--- a/xmonad-contrib.cabal
+++ b/xmonad-contrib.cabal
@@ -168,6 +168,7 @@ library
                         XMonad.Doc.Configuring
                         XMonad.Doc.Developing
                         XMonad.Doc.Extending
+                        XMonad.Hooks.BorderPerWindow
                         XMonad.Hooks.CurrentWorkspaceOnTop
                         XMonad.Hooks.DebugEvents
                         XMonad.Hooks.DebugKeyEvents


XMonad/Hooks/BorderPerWindow.hs
```diff
@@ -0,0 +1,132 @@
...
+
+-- | The XMonad message that transfers 'WindowBorderSpec' from the window
+-- creation event to the layout handling function.
+-- Note that XMonad reads config and draws window border after 'ManageHook's
+-- are applied. It does not work if we set the border in a 'ManageHook'
+-- (hooking window creation). So we choose to send a message in
+-- 'defineWindowBorder' function to a later stage, e.g., layout handling, to
+-- set the window border.
+data WindowBorderSpecMessage
+    = WindowBorderSpecMessage WindowBorderSpec deriving (Read, Show)
```
slotThe, line 95: We can probably combine these into a single `WindowBorder[Spec]Message` since they are not used separately

XMonad/Hooks/BorderPerWindow.hs
```diff
@@ -0,0 +1,132 @@
...
+
+instance LayoutModifier BorderPerWindow a where
+    modifierDescription = const "BorderPerWindow"
+
+    redoLayout (BorderPerWindow Nothing) _ _ wrs =
+        return (wrs, Nothing)
+
+    redoLayout (BorderPerWindow (Just (WindowBorderSpec w bw))) _ _ wrs = do
+        withDisplay $ \d -> io $ setWindowBorderWidth d w bw
+        return (wrs, Nothing)
```
slotThe, line 122: I think this would be an ideal candidate for separating the `X` part and the layouting part.  Basically, instead of specifying `redoLayout` you could put the `withDisplay` call into `hook` and not have all of this noise about returning the window/rectangle pairs (this is then taken care of by the default implementation of `pureModifier` and `redoLayout`)

slotThe (commented):
I like this, thanks!  I wonder if we should combine it with VoidBoders, somehow[^1].  The ergonomics are slightly worse (having to specify a manageHook in addition to a layout modifier) but it might sitll be worth it.

[^1]: Which would mean deprecating a module that was quite literally just released—oh well.

subbyte:
> I like this, thanks! I wonder if we should combine it with VoidBoders, somehow. The ergonomics are slightly worse (having to specify a manageHook in addition to a layout modifier) but it might sitll be worth it.

Good suggestion! I had the same thought. However, after writing the code, I find they are two functionalities:
1. `VoidBorders` specifies a layout but not a window to set all windows borderless.
2. `BorderPerWindow` specifies a window but not a layout to define border width.

I was thinking to combine them from the beginning, but I haven't figured out a good way for users to tell the combined module what to do. The current last step for both are layoutmodifier. To achieve `VoidBorders`, the user understands any layout it modifies will have all window borderless. To achieve `BorderPerWindow`, the user understands not all window in a layout decorated by `BorderPerWindow` will have customized border width. We could use a parameter to the combined module to indicate all windows, but that is confusing since the user also specifies border for each window.

The two modules look similar because part of the implementation approach they share. One step back, the original goal of `BorderPerWindow` is to not bother any layout, but focus on which window has which border setting. The config will not change in different layouts. I involve a layout hook because this is the only post-managehook-procedure I know (sorry about my limited knowledge of XMonad). I need to run `setWindowBorderWidth` after manage hook since [XMonad do `setWindowBorderWidth` with default border config after applying all manage hooks](https://stackoverflow.com/a/69774944/8418198). This also goes to your second code comment. I tried to do `setWindowBorderWidth` inside the manage hook (so we even do not need the layout hook), which is the [Attempt A in my stackoverflow question](https://stackoverflow.com/questions/69725874/config-no-border-for-specific-windows-in-xmonad) but it does not work according to [my answer](https://stackoverflow.com/a/69774944/8418198).

To summarize, we may not want users to think this is a layout hook (named as `XMonad.Hooks.BorderPerWindow`) since this is per window config that applies to all layouts. The border setting is permanent, so it makes little sense to use it per layout (typical layout hook) like
```
myLayoutHook = Tall 1 (3/100) (1/2) ||| setBorderPerWindow Full ||| ...
```
The layout hook in the implementation is just an approach, not goal, and the meaningful way of using it is to apply it to all layouts since the window border setting will not be different between layouts.
```
myLayoutHook = setBorderPerWindow
             $ Tall 1 (3/100) (1/2) ||| Full ||| Grid ||| ...
```

geekosaur:
If all you're looking for is to apply it post-`manageHook`, you may be looking for `logHook` instead of the layout.

XMonad/Hooks/BorderPerWindow.hs
subbyte: The reason I introduced `WindowBorderSpec` is for
```
data BorderPerWindow a
    = BorderPerWindow (Maybe WindowBorderSpec) deriving (Read, Show)
```
I think we need a data type for `Maybe` other than two arguments. And since I introduce `WindowBorderSpec`, I just use it in `WindowBorderSpecMessage` to make it simpler.

I think the only chance to get rid of it is to eliminate it in both, which goes to your second comment. Please see my thoughts there.

XMonad/Hooks/BorderPerWindow.hs
subbyte: I need to make sure I fully understand your point. Please correct me if I misunderstand any part of it.

If we put `withDisplay` (basically `setWindowBorderWidth`) into `hook` (I assume you mean the manage hook `defineWindowBorder`), then we actually do not need a layout hook afterwards, so it goes to [Attempt A in my stackoverflow question](https://stackoverflow.com/questions/69725874/config-no-border-for-specific-windows-in-xmonad) but [people found it does not work 12 years ago](https://xmonad.haskell.narkive.com/130DKJDZ/how-to-make-a-window-floating-above-all-the-other-windows-and-border-less) and I checked `XMonad.operations` code and [confirmed it does not work](https://stackoverflow.com/a/69774944/8418198). As I detailed in [my comment below](https://github.com/xmonad/xmonad-contrib/pull/640#issuecomment-957847814), basically I need to execute `setWindowBorderWidth` outside and after a manage hook to make it work, so I create a layout hook to do it and try to carefully only execute it once (not to have any [windows flashing/jumping/scaling effect](https://github.com/xmonad/xmonad-contrib/issues/316)).

Please advice or correct me if I incorrectly understand your point.

subbyte:
> If all you're looking for is to apply it post-`manageHook`, you may be looking for `logHook` instead of the layout.

Thanks for the suggestion! Looks like a good one to check. Will try in a few days.

liskin:
> > If all you're looking for is to apply it post-`manageHook`, you may be looking for `logHook` instead of the layout.
> 
> Thanks for the suggestion! Looks like a good one to check. Will try in a few days.

I'm a bit afraid that if you want to avoid the borders being seen for a split-second, logHook might be too late.

But I like the idea nonetheless. A generic logHook that executes little tasks from a queue might be useful elsewhere too. We have a couple usecases that would benefit from an afterManageHook, for example, and this seems like an easy way to implement that.

XMonad/Hooks/BorderPerWindow.hs
slotThe: I meant it in the sense that we could just define

``` haskell
  data WindowBorderMessage
      = WindowBorderMessage Window Dimension deriving (Read, Show)

  instance Message WindowBorderMessage
```

and be on our way.  The definition of `BorderPerWindow` wouldn't need to change at all but it would safe us from having to wrap/unwrap twice a few times

XMonad/Hooks/BorderPerWindow.hs
slotThe: Sorry, I should've been more clear.  The `LayoutModifier` class actually has a [hook](https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/XMonad-Layout-LayoutModifier.html#v:hook) member function (perhaps a bit weirdly named) which gets executed by the default implementation of `redoLayout`: 

``` haskell
  redoLayout m r ms wrs = do hook m; return $ pureModifier m r ms wrs
```

Since we're not doing any layouty bits at all (i.e., we just return `(wrs, Nothing)`) and the only interesting thing is the call to `setWindowBorderWidth`, I was suggesting to implement `hook` instead of `redoLayout` here

subbyte: I see. Thank you for the information! Yes, this is a better place. Let me try this as well as `logHook` to figure out a good way to implement the functionality.

XMonad/Hooks/BorderPerWindow.hs
subbyte: I think I may get a good opportunity to better understand a little more on GHC. Could you explain a little more on wrap/unwarp twice? Or any article I can read? I really want to make it fast if possible, at least not adding too much overhead because of the module.

slotThe: Every time you create a new type with `data` you're essentially adding one layer of indirection (i.e., one more pointer to follow) to what you're handling.  Wrapping and unwrapping then _technically_ adds some costs because you need to follow the pointers to the actual object that you want to look at and doing that twice is more expensive than only doing it once :)

There is `newtype` for wrappers around a single field, which then has the same runtime representation as the thing that it wraps.  For example, you could just let `WindowBorderSpecMessage` be a newtype and that would be as "efficient" as what I'm suggesting.

However, I don't think overhead/performance is the primary goal here; I was thinking more along the lines of convenience.  I feel that in this case it simply reduces mental overhead (one less type that looks a lot like the others to worry about).  Plus, instead of something like

```haskell
liftX . broadcastMessage . WindowBorderSpecMessage $ WindowBorderSpec w bw
```

one gets to write

```haskell
liftX . broadcastMessage $ WindowBorderSpecMessage w bw
```

subbyte: I see. Thanks for the explanation!
So you suggest instead of
```Haskell
    pureMess (BorderPerWindow _) msg
        | Just (WindowBorderSpecMessage spec) <- fromMessage msg =
            Just . BorderPerWindow $ Just spec
```
I can revise it to
```Haskell
    pureMess (BorderPerWindow _) msg
        | Just (WindowBorderSpecMessage w bw) <- fromMessage msg =
            Just . BorderPerWindow . Just $ WindowBorderSpec w bw
```
When I wrote the old code. I guess `Just spec` will ease the compiler's job---the compiler directly passes the address of `spec` from `WindowBorderSpecMessage` to `BorderPerWindow` without checking what's inside. And until the `redoLayout` function the `WindowBorderSpec` data needs to be unpacked to expose the internal structure with guards:
```Haskell
    redoLayout (BorderPerWindow (Just (WindowBorderSpec w bw))) _ _ wrs = do
```
Is my guess correct?

slotThe: > I see. Thanks for the explanation! So you suggest instead of
>
> […]
>
> I can revise it to
>
> ```haskell
>     pureMess (BorderPerWindow _) msg
>         | Just (WindowBorderSpecMessage w bw) <- fromMessage msg =
>             Just . BorderPerWindow . Just $ WindowBorderSpec w bw
> ```

I believe it would just be

```haskell
    pureMess (BorderPerWindow _) msg
        | Just spec@WindowBorderSpecMessage{} <- fromMessage msg =
            Just . BorderPerWindow $ Just spec
```

But really, just rename `WindowBorderSpec` to `WindowBorderSpecMessage` (and delete the other type, of course) and follow the compiler—it won't disappoint!

> When I wrote the old code. I guess `Just spec` will ease the compiler's job---the compiler directly passes the address of `spec` from `WindowBorderSpecMessage` to `BorderPerWindow` without checking what's inside. And until the `redoLayout` function the `WindowBorderSpec` data needs to be unpacked to expose the internal structure with guards:
>
> […]
>
> Is my guess correct?

Sure, matching on only the outer constructor is probably cheaper because we don't have to actually look into the structure (this is also done by `spec@WindowBorderSpecMessage{}` as it only checks the outer shape of the data).  Though this again only conceptually matters—practically, the call to `setWindowBorderWidth` will take the vast majority of the time anyways

subbyte: Oh, I see. Thanks, @slotThe! Will do.

subbyte:
Thank you for your guidance, @slotThe ! I simplified the code in my second commit, and tested it well.

I would like to try `logHook` as well to see if the implementation has any lag. However, I haven't figured out how to obtain XMonad Message in a `logHook` if I keep the `manageHook` part of the module. I still want to use `manageHook` since it provides a good way to specify windows. Please advice how I can move forward @geekosaur .

Besides the two, the best implementation in my mind is to have a *post-window-management-hook* after the [`window` function](https://github.com/xmonad/xmonad/blob/master/src/XMonad/Operations.hs#L107), since the existing `manageHook` executes before it. Not sure if there are other functionalities sharing the same need to add the hook in XMonad.

liskin:
> I would like to try `logHook` as well to see if the implementation has any lag. However, I haven't figured out how to obtain XMonad Message in a `logHook` if I keep the `manageHook` part of the module.

My idea for the `logHook`-based implementation was that you'd simply make an ExtensibleState datatype holding a queue (a list that you then reverse before processing) of `X ()` actions and a `logHook` that executes these actions and drops them from the queue. The `manageHook` would simply push `withDisplay $ \d -> io $ setWindowBorderWidth d w bw` directly into that queue.

And if this causes flicker, you can implement the same as a LayoutModifier, which should then execute the actions before the window is mapped.

I'm quite surprised we don't have both of those already… :-)

geekosaur:
More or less what @liskin said, yes.

subbyte:
> My idea for the `logHook`-based implementation was that you'd simply make an ExtensibleState datatype holding a queue (a list that you then reverse before processing) of `X ()` actions and a `logHook` that executes these actions and drops them from the queue. The `manageHook` would simply push `withDisplay $ \d -> io $ setWindowBorderWidth d w bw` directly into that queue.

Thanks for the point! If we do not make the `X ()` action generic, I think we can just pass the data `WindowBorderSpec` in the queue, which I think I know how to write.

If we want to implement the `X ()` actions in the queue, which can be anything returning `X ()` like `withDisplay $ \d -> io $ setWindowBorderWidth d w bw`, I am thinking we may need a Monad like:
```Haskell
data WindowBorderSpec
    = WindowBorderSpec Window Dimension deriving (Read, Show)

newtype WindowBorderAction a = WindowBorderAction (ReaderT WindowBorderSpec X a)
    deriving (Functor, Applicative, Monad, MonadReader WindowBorderSpec, MonadIO)

runWindowBorderAction (WindowBorderAction m) spec = runReaderT m spec

type WindowBorderActionQueue = [WindowBorderAction]

instance ExtensionClass WindowBorderActionQueue where
    initialValue = []
```
I am not sure how to encapsulate `WindowBorderActionQueue` in a `manageHook` using `ExtensibleState.modify` (I see `Query` is a different monad based on `X`). Maybe I over-complicate the problem.

slotThe:
Mh, I don't think we need an extra monad for this (if I understood the idea correctly).  For a "queue" of actions we could define

```haskell
newtype ActionQueue = ActionQueue { getQueue :: [X ()] }

instance ExtensionClass ActionQueue where
    initialValue :: ActionQueue
    initialValue = ActionQueue []

appendQueue :: X () -> X ()
appendQueue a = XS.modify @ActionQueue $ coerce (a :) -- or just ActionQueue . ... . getQueue
```

and be good.  The `defineWindowBorder` function would then simply add some action to that queue

```haskell
defineWindowBorder :: Dimension -> ManageHook
defineWindowBorder bw = do
    w <- ask
    liftX $ appendQueue (withDisplay $ \d -> io $ setWindowBorderWidth d w bw)
    idHook
```

and users would have to add a logHook along the lines of

```haskell
queueLogHook :: X ()
queueLogHook = do
    as <- reverse . getQueue <$> XS.get @ActionQueue
    sequenceA_ as
    XS.put (ActionQueue [])
```

to their own logHook.

TheMC47:
Just coming by to say that I like the `logHook` idea better than a layout, but this is still an awesome feature @subbyte :tada:

liskin:
> and users would have to add a logHook along the lines of
> 
> ```haskell
> queueLogHook :: X ()
> queueLogHook = do
>     sequenceA_ . reverse . getQueue <$> XS.get 
>     XS.put (ActionQueue [])
> ```
> 
> to their own logHook.

Note that this isn't safe in general, as performing those actions can raise exceptions and the actions aren't removed from the queue then. So if we wanted to have something like this in `Hooks` or `Util`, we'd need to pop the actions before executing them, one by one. So this probably means using `Data.Sequence` instead of a list, unfortunately.

> Just coming by to say that I like the `logHook` idea better than a layout, but this is still an awesome feature

Yeah, it conceptually simpler and also safer. A general queue of `X ()` actions to be executed during the layout would be dangerous if anyone called `windows` in there. On the other hand, I don't see another way to avoid flicker. It's a silly tradeoff. :-/

(Note that there are many other places in xmonad where flicker could be avoided with a more complicated design, but it's just not worth it. Perhaps this is another one of those?)

subbyte:
Thanks for the guidance, @slotThe @liskin ! Will have a try to test it this week.

subbyte:
Thanks @slotThe for the code! I tried it and fixed an issue that the `sequenceA_` with `<$>` did not work (just unfolding the line makes it work yet I am not sure why). Here's the version that works finally:
```Haskell
newtype ActionQueue = ActionQueue { getQueue :: [X ()] }

instance ExtensionClass ActionQueue where
    initialValue = ActionQueue []

defineWindowBorder :: Dimension -> ManageHook
defineWindowBorder bw = do
    w <- ask
    liftX . XS.modify . append . withDisplay $ \d -> io $ setWindowBorderWidth d w bw
    idHook
  where
    append :: X () -> ActionQueue -> ActionQueue
    append x = coerce (x :)

queueLogHook :: X ()
queueLogHook = do
    as <- XS.get
    sequenceA_ . reverse $ getQueue as
    XS.put (ActionQueue [])
```
Before improving the realization using `Data.Sequence` and pop-n-execute each action as @liskin suggested, I tested the latency and visual effect of the `logHook` approach.
- The happy part is I don't feel any latency (or flicker of border removal) using `logHook`. I am very sensitive to flicker, and hopefully this is a strict human test.
- However, there is an issue: a new window created without border will not be re-rendered after the border is removed, which leaves blank (background) of double border width at the right and bottom of the window (see screenshot below for example). Next if I switch to another workspace and switch back, then the window is re-rendered and there is no blank space any more. Actually I experience the same when using `XMonad.Actions.NoBorders`. Not sure if this is my configuration issue or a logic issue without involving the XMonad layout procedure.

xmonad-640-loghook-approach-issue

subbyte:
Another thought, which is not related to the issue above but related to the improvement of realization @liskin mentioned: a window is created at a time, and there is a [`logHook` function in the window creation handler](https://github.com/xmonad/xmonad/blob/master/src/XMonad/Operations.hs#L212). For this extension only, we may not need to support a queue of actions, just
```
newtype BorderAction = BorderAction { getAction :: X () }
```
Since each border adjustment action will be processed for each [window creation](https://github.com/xmonad/xmonad/blob/master/src/XMonad/Operations.hs#L141).

A potential issue is when window creation is too fast, and the `windows` function hasn't finished for one window while another one is executed, so `BorderAction` may be overwritten or overcleared by each. Please comment if window creation events are handled in a multi-threading model with shared ExtensibleState and this is the reason for the queue.

liskin:
> * However, there is an issue: a new window created without border will not be re-rendered after the border is removed, which leaves blank (background) of double border width at the right and bottom of the window (see screenshot below for example). Next if I switch to another workspace and switch back, then the window is re-rendered and there is no blank space any more. Actually I experience the same when using `XMonad.Actions.NoBorders`. Not sure if this is my configuration issue or a logic issue without involving the XMonad layout procedure.

I'm afraid this is a logic issue. When the border width is changed, the window effectively becomes smaller, and it's only resized to the correct size on the next refresh:
https://github.com/xmonad/xmonad/blob/54df2e9acdbdd85b830865464bd89b36688b9963/src/XMonad/Operations.hs#L192
https://github.com/xmonad/xmonad/blob/54df2e9acdbdd85b830865464bd89b36688b9963/src/XMonad/Operations.hs#L310

So to make this work in a logHook, you might need to get the current border width using `withWindowAttributes` and `moveResizeWindow` to compensate for the border width change. Or just `refresh`, which might result in some flicker.

> For this extension only, we may not need to support a queue of actions, just…

The point of having a queue is that a) it's useful for other modules as well and perhaps even useful for user configs in general, b) it's conceptually simpler—you don't need to think hard whether you need to store one, zero-or-one, zero-or-more or one-or-more.

subbyte:
> So to make this work in a logHook, you might need to get the current border width using withWindowAttributes and moveResizeWindow to compensate for the border width change. Or just refresh, which might result in some flicker.

Now it is important to pop an action before executing it if I use `refresh`, which basically calls `windows`, at the end of which the `logHook` is invoked again.

To put action queue as a generic function others can use as well, we may want to create two modules: one for action queue, and one for the border customization. Here's the code I tested to work:
```Haskell
-- XMonad/Utils/ActionQueue.hs
-- exporting `executeActionQueue` and `appendActionQueue`
newtype ActionQueue = ActionQueue { getQueue :: Seq (X ()) }

instance ExtensionClass ActionQueue where
    initialValue = ActionQueue empty

appendActionQueue :: X () -> ActionQueue -> ActionQueue
appendActionQueue a (ActionQueue as) = ActionQueue (as |> a)

executeActionQueue :: X ()
executeActionQueue = do
    aq <- XS.get
    executeActionSeq $ getQueue aq

executeActionSeq :: Seq (X ()) -> X ()
executeActionSeq Empty = XS.put $ ActionQueue empty
executeActionSeq (a :<| as) = (XS.put $ ActionQueue as) >> a
```
```Haskell
-- XMonad/Hooks/BorderPerWindow.hs
-- exporting `defineWindowBorder`
defineWindowBorder :: Dimension -> ManageHook
defineWindowBorder bw = do
    w <- ask
    liftX . XS.modify . appendActionQueue $ (setBorder w bw >> refresh)
    idHook

setBorder :: Window -> Dimension -> X ()
setBorder w bw =
    withDisplay $ \d -> io $ setWindowBorderWidth d w bw
```
This code does not yield much flicker effect (sometimes a little bit) and I feel good with the testing results. If I do
```Haskell
    liftX . XS.modify . appendActionQueue $ setBorder w bw
    liftX . XS.modify . appendActionQueue $ refresh
```
instead of putting it in one action, I see obvious flicker every time.

Please advice if this is the way the community would like, especially whether splitting it into two modules is a good idea. Then I will rebase the PR to the logHook approach for further code review.

slotThe:
On Mon, Nov 15 2021 21:17, Xiaokui Shu wrote:
>     liftX . XS.modify . appendActionQueue $ (setBorder w bw >> refresh)

I wonder if we could put `refresh` as the `Empty` case of
`executeActionQueue` or if that would be too late (it would certainly
save us many calls to it)

> Please advice if this is the way the community would like, especially
> whether splitting it into two modules is a good idea. Then I will
> rebase the PR to the logHook approach for further code review.

Oh yeah, I think the original idea definitely was to have an action
queue as a separate module and let this module just use it via importing
it

subbyte:
> I wonder if we could put `refresh` as the `Empty` case of `executeActionQueue` or if that would be too late (it would certainly save us many calls to it)

I tried two ways to remove `refresh` into Seq pattern matching in `executeActionSeq`:

1. `executeActionSeq Empty = refresh`: this will trigger a logic bomb that XMonad uses 100% CPU on a single core and restarting XMonad cannot release since `refresh` has `logHook` (in `windows`), which will do `refresh` recursively on empty `ActionQueue`.

2. I think I should do:
```Haskell
executeActionSeq :: Seq (X ()) -> X ()
executeActionSeq Empty = pure ()
executeActionSeq (a :<| Empty) = (XS.put $ ActionQueue empty) >> a >> refresh
executeActionSeq (a :<| as) = (XS.put $ ActionQueue as) >> a
```
This works. In theory and in practice, I don't find this differs from my previous version: for the window border use case, only one window is opened at a time, and the queue is at most one action before it is cleared in `windows` function (`logHook` at the end). In my old code, the single action is actually two: border setup and refresh. In the new code, the single action is just border setup, plus the refresh when this single (last) action is cleared.

I think we may need to think from the generic ActionQueue module perspective whether we need `refresh` for actions (of any use cases) to be cleared. This may be not necessary or not the intention of the developer who is using ActionQueue.

liskin:
> ```haskell
> -- XMonad/Utils/ActionQueue.hs
> -- exporting `executeActionQueue` and `appendActionQueue`
> newtype ActionQueue = ActionQueue { getQueue :: Seq (X ()) }
> 
> instance ExtensionClass ActionQueue where
>     initialValue = ActionQueue empty
> 
> appendActionQueue :: X () -> ActionQueue -> ActionQueue
> appendActionQueue a (ActionQueue as) = ActionQueue (as |> a)
> 
> executeActionQueue :: X ()
> executeActionQueue = XS.get >>= executeActionSeq <$> getQueue
> 
> executeActionSeq :: Seq (X ()) -> X ()
> executeActionSeq Empty = XS.put $ ActionQueue empty
> executeActionSeq (a :<| as) = (XS.put $ ActionQueue as) >> a
> ```

This looks like `executeActionSeq` only executes the head of the queue, and leaves the rest for the next `logHook` invocation. Does that explain the flicker? (Actually I'd expect much more than that, a noticeable delay.)

liskin:
And I don't think the queue itself should be forcing the refresh. It's a logic bomb, and trying to hack it into not being a logic bomb is futile.

subbyte:
> This looks like `executeActionSeq` only executes the head of the queue, and leaves the rest for the next `logHook` invocation. Does that explain the flicker? (Actually I'd expect much more than that, a noticeable delay.)

Good catch! There is a logic error here.

However, I avoid the logic error by luck since the queue is at most one action for the window border use case as I explained earlier. For generic ActionQueue uses, I should execute all actions. Let me update.

subbyte:
Basic code submitted. Feel free to comment. I will add docs in the next days.

One strange thing: this `logHook` approach works without visible flicker for most windows such as `chromium` and `gvim`. However, `firefox` gives me a strange behavior: the window is first drawn with border, then the border is removed and the window is moved to the bottom left to the screen. This is different from [the phenomenon I encountered without refresh](https://github.com/xmonad/xmonad-contrib/pull/640#issuecomment-968196160), where
1. the window is moved to the top left (this time bottom left)
2. the desktop background is visible for the non-border area (this time just black)
3. all windows have the issue (this time only `firefox`)
4. switch to another workspace and back will rerender it correctly (this time not, the black part is always there until I open another window on the same workspace)

See attached screenshot for example. Any idea why?

2021-11-17-003211_2560x1440_scrot

subbyte:
Never mind the `firefox` issue. This traces back to a bug with `firefox` and `nvidia` driver. I configured `firefox` with `layers.offmainthreadcomposition.force-disabled:true` according to [this post](https://support.mozilla.org/en-US/questions/1152147), and it works now.

No flicker for any window with the `logHook` approach now.

XMonad/Hooks/BorderPerWindow.hs
```diff
@@ -0,0 +1,71 @@
...
+    w <- ask
+    liftX . XS.modify . enqueue $ updateBorderWidth w bw
+    idHook
+
+updateBorderWidth :: Window -> Dimension -> X ()
+updateBorderWidth w bw = do
+    withDisplay $ \d -> io $ setWindowBorderWidth d w bw
+    refresh
+
+setCustomWindowBorder = exequeue
```
slotThe, line 71: I think I would be against an alias like this.  If another module decides to use X.A.ActionQueue also does this you'd have the same logHook twice in there; that doesn't feel right

slotThe (approved):
LGTM!

XMonad/Hooks/BorderPerWindow.hs
liskin: Perhaps the ActionQueue module could export just an `XConfig a → XConfig a` combinator which configures the logHook exactly once (using extensibleConf)?

liskin: (I'll be happy to refactor this later, which I was going to do anyway as I think the queue handling needs some improvements but I can't look into it today and don't want to block the rest of you.)

slotThe: > Perhaps the ActionQueue module could export just an `XConfig a → XConfig a` combinator which configures the logHook exactly once (using extensibleConf)?

Oh, right, we have tools for that now!  In that case this sounds like the right thing to do

> (I'll be happy to refactor this later, which I was going to do anyway as I think the queue handling needs some improvements but I can't look into it today and don't want to block the rest of you.)

I don't suppose there is much of a hurry (except for @subbyte not having to run their own fork anymore :)); waiting a few days longer seems very much acceptable

subbyte:
Thanks guys @slotThe @liskin ! I pushed a new commit including the doc and a little code editing to make them more modular. Going forward, may I ask what I need to do to make it a branch here for you to upgrade?

slotThe:
I will merge this for now, since the largest issues (using ExtensibleConf and executing the actions one by one) are fixed now.  If we still see anything that needs to be improved we can always push another commit afterwards

ixzh:
the doc example seems to have a typo:
```
myManageHook = composeAll
    [ className =? "firefox"  --> defineWindowWidth 0
    , className =? "Chromium" --> defineWindowWidth 0
    , isDialog                --> defineWindowWidth 8
    ]
```
`defineWindowWidth` to be `defineBoorderWidth`?

slotThe:
Oh indeed, thanks!  Fixed by 95b37a9ab2cb3e5081b17f5f5161b66e0d9c69d5

PR #641, opened 2021-11-03
X.Prompt: Add transposeChars

slotThe:
### Description

This is an analogue to Emacs's `transpose-chars` function (expect that
it does not take a universal argument), bound to its default keybinding.

### Checklist

        manually, ...) and concluded: Manually, for now.  The transposing logic isn't that complicated, but the whole prompt machinery makes this kind of hard to unit/property test.

### DIFF
--- a/XMonad/Prompt.hs
+++ b/XMonad/Prompt.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ViewPatterns #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Prompt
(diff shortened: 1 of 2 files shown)


XMonad/Prompt.hs
```diff
@@ -1227,6 +1229,23 @@ changeWord p = join $ f <$> getInput <*> getOffset <*> pure p
...
+transposeChars :: XP ()
+transposeChars = do
+  off <- gets offset
+  cmd <- gets command
+  let (beforeCursor, afterCursor) = splitAt off cmd
+      (ncom, noff) = fromMaybe (cmd, off) (go beforeCursor afterCursor off)
+  modify $ \s -> setCommand ncom $ s{ offset = noff }
+ where
+  go :: [a] -> [a] -> Int -> Maybe ([a], Int)
+  go (reverse -> (b1 : b2 : bs)) [] offset =  -- end of line
```
liskin, line 1243: I wonder why hlint doesn't complain about the extra parens here.

liskin: (No need to remove them, it's just weird :-))

slotThe: hlint seems reluctant to suggest this in patterns, it seems.  The following are also completely fine (on `HLint v3.3.1`): 

``` haskell
f :: [a] -> [a]
f ((((a : as)))) = as
f _              = []

g :: [a] -> [a]
g = \case
  ((((a:as)))) -> as
  _            -> []
```

PR #643, opened 2021-11-06
add to hacks: rough draft trayer resize event hook

tulth:
### Description

Add a hook intended to be used with xmobar to auto reserve a space for trayer.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -103,6 +103,14 @@
     - Changed type signature of `keysMoveWindow` from `D -> Window -> X ()`
       to `ChangeDim -> Window -> X ()` to allow negative numbers without compiler warnings.
 
+  * `XMonad.Util.Hacks`
+
+    - Added `trayerQuery` for easily querying trayer and
+      `trayerPaddingXmobarEventHook` (as well as a variant to specify a
+      different X property) to communicate trayer resize events to
+      XMobar so that padding space may be reserved on xmobar for the
+      tray.  Requires `xmobar` version 0.40 or higher.
+
 ## 0.17.0 (October 27, 2021)
 
 ### Breaking Changes
--- a/XMonad/Util/Hacks.hs
+++ b/XMonad/Util/Hacks.hs
@@ -32,12 +32,20 @@ module XMonad.Util.Hacks (
 
   -- * Stacking trays (trayer) above panels (xmobar)
   -- $raiseTrayer
+  trayerQuery,
   trayerAboveXmobarEventHook,
   trayAbovePanelEventHook,
+
+  -- * Inform xmobar when trays (e.g., trayer) change width
+  -- $padTrayer
+  trayPaddingXmobarDefProp,
+  trayerPaddingXmobarEventHook,
+  trayerPaddingXmobarEventHook',
   ) where
 
 
 import XMonad
+import XMonad.Hooks.StatusBar (xmonadPropLog')
 import XMonad.Prelude (All (All), filterM, when)
 import System.Posix.Env (putEnv)
 
@@ -126,9 +134,13 @@ javaHack conf = conf
 --
 -- > handleEventHook = … <> Hacks.trayerAboveXmobarEventHook
 
--- | 'trayAbovePanelEventHook' for trayer/xmobar
+-- | A 'Query' to identify the trayer window.
+trayerQuery :: Query Bool
+trayerQuery = className =? "trayer"
+
+-- | Like 'trayAbovePanelEventHook', but specialised for trayer/xmobar.
 trayerAboveXmobarEventHook :: Event -> X All
-trayerAboveXmobarEventHook = trayAbovePanelEventHook (className =? "trayer") (appName =? "xmobar")
+trayerAboveXmobarEventHook = trayAbovePanelEventHook trayerQuery (appName =? "xmobar")
 
 -- | Whenever a tray window lowers itself to the bottom of the stack, look for
 -- any panels above it and lower these.
@@ -145,3 +157,75 @@ trayAbovePanelEventHook trayQ panelQ ConfigureEvent{ev_window = w, ev_above = a}
     mapM_ (io . lowerWindow dpy) panelWs
   mempty
 trayAbovePanelEventHook _ _ _ = mempty
+
+-- $padTrayer
+-- Communicating tray (e.g., trayer) resize events to xmobar so that
+-- padding space may be reserved on xmobar for the tray.
+--
+-- Basic Usage with trayer:
+--
+-- First, add the padding hook to your @handleEventHook@ as follows:
+--
+-- > main = xmonad $ def
+-- > { ...
+-- > , handleEventHook = handleEventHook def
+-- >                  <> Hacks.trayerPaddingXmobarEventHook
+-- > }
+--
+-- Then, assuming the tray is placed on the right, update your
+-- @xmobarrc@ as follows:
+--
+-- > Config { ...
+-- >        , commands = [ ...
+-- >                     , Run XPropertyLog "_XMONAD_TRAYPAD", ... ]
+-- >        , template = " ... %_XMONAD_TRAYPAD%"
+-- >        }
+--
+-- As an example of what happens in this basic usage, consider the
+-- case where trayer updates to a width of 53 pixels.
+-- The following property will appear on the root window:
+--
+-- > _XMONAD_TRAYPAD(UTF8_STRING) = "<hspace=53/>"
+
+-- | 'trayPaddingXmobarDefProp' is default property name,
+-- @"_XMONAD_TRAYPAD"@, to use with 'xmonadPropLog''
+trayPaddingXmobarDefProp :: String
+trayPaddingXmobarDefProp = "_XMONAD_TRAYPAD"
+
+-- | A simple trayer-specific event hook that watches for trayer window
+-- resize changes and update the value in the property specified by
+-- 'trayPaddingXmobarDefProp'.
+trayerPaddingXmobarEventHook :: Event -> X All -- ^ event hook
+trayerPaddingXmobarEventHook = trayerPaddingXmobarEventHook' trayPaddingXmobarDefProp
+
+-- | A more generic version of 'trayerPaddingXmobarEventHook' that
+-- allows the user to specify the property to use with 'xmonadPropLog''
+-- when 'trayPaddingXmobarDefProp' is not desired.  This is still a
+-- trayer-specific hook.
+trayerPaddingXmobarEventHook'
+  :: String         -- ^ 'xmonadPropLog'' string to use
+  -> Event -> X All -- ^ event hook result
+trayerPaddingXmobarEventHook' s = trayPaddingXmobarEventHook (trayDefaultAction s) trayerQuery
+
+-- | A fully generic tray resize hook.  This function is not
+-- trayer-specific; note the prefix is @tray@, not @trayer@.  Both the
+-- action to take and the tray identification query are given as
+-- arguments.
+trayPaddingXmobarEventHook
+  :: (Int -> X())   -- ^ action to take when query succeeds, pixels to action
+  -> Query Bool     -- ^ query to identify the tray window
+  -> Event -> X All -- ^ event hook result
+trayPaddingXmobarEventHook action trayQ ConfigureEvent{ ev_window = w, ev_width = wa } = do
+  whenX (runQuery trayQ w) $ action (fromIntegral wa)
+  return (All True)
+trayPaddingXmobarEventHook _ _ _ = return (All True)
+
+-- | The default tray action that is used by both
+-- 'trayerPaddingXmobarEventHook' and 'trayerPaddingXmobarEventHook''.
+-- This action places @\<hspace=pixels\/\>@ on the specified
+-- 'xmonadPropLog'' property.
+trayDefaultAction
+  :: String -- ^ 'xmonadPropLog'' property to use
+  -> Int    -- ^ new tray width in pixels
+  -> X ()   -- ^ resultant update
+trayDefaultAction xPropLog n = xmonadPropLog' xPropLog ("<hspace=" ++ show n ++ "/>")


XMonad/Util/Hacks.hs
```diff
@@ -32,14 +32,20 @@ module XMonad.Util.Hacks (
 
   -- * Stacking trays (trayer) above panels (xmobar)
   -- $raiseTrayer
+  trayerQuery,
   trayerAboveXmobarEventHook,
+  trayerResizeHook,
   trayAbovePanelEventHook,
+  trayResizeHook,
+  trayerResizeHook',
```
TheMC47, line 40: You should put the resize functions under their own section

TheMC47 (changes requested):
Awesome, good job :tada: I left you some comments

@liskin any other events we should catch?

TheMC47:
And one more thing: I'd personally like a `trayResizeHook' :: (Int -> X ()) -> Query Bool -> Event -> X All`, since it might be a bit more flexible (I need that to pass the width through STM to xmobar, since I plan to move away from  X Properties)

tulth:
I tried adding an event like this to handle the trayer being destroyed, but it does not work
```haskell
trayResizeHook action trayQ DestroyWindowEvent { ev_window = w} = do  -- DOES NOT WORK
  whenX (runQuery trayQ w) (  -- this query always fail but I am not clear why.  if i comment it out, the resize does happen
    action 0
    )
  return (All True)
```

XMonad/Util/Hacks.hs
```diff
@@ -145,3 +154,28 @@ trayAbovePanelEventHook trayQ panelQ ConfigureEvent{ev_window = w, ev_above = a}
...
+  :: String           -- ^ xmonadPropLog name
+  -> (Event -> X All) -- ^ event hook
+trayerResizeHook' = trayResizeHook trayerQuery
+
+trayResizeHook
+  :: Query Bool       -- ^ tray query
+  -> String           -- ^ xmonadPropLog name
+  -> (Event -> X All) -- ^ event hook
+trayResizeHook trayQ xPropLog ConfigureEvent { ev_window = w } = do
+    whenX (runQuery checkDock w <&&> runQuery trayQ w) (
```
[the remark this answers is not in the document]
tulth: I agree, and I just pushed an update that removes checkdock.

XMonad/Util/Hacks.hs
tulth: i moved them down in the latest rebase.

tulth:
> And one more thing: I'd personally like a `trayResizeHook' :: (Int -> X ()) -> Query Bool -> Event -> X All`, since it might be a bit more flexible (I need that to pass the width through STM to xmobar, since I plan to move away from X Properties)

I updated to handle a case like this in the latest rebase.  What do you think, does it meet the need now?

XMonad/Util/Hacks.hs
```diff
@@ -145,3 +158,49 @@ trayAbovePanelEventHook trayQ panelQ ConfigureEvent{ev_window = w, ev_above = a}
...
+trayerResizeHook
+  :: (Event -> X All) -- ^ event hook
+trayerResizeHook = trayerResizeHook' trayDefaultXpropLogName
+
+trayerResizeHook'
+  :: String           -- ^ xmonadPropLog string to use
+  -> (Event -> X All) -- ^ event hook result
+trayerResizeHook' s = trayResizeHook (trayDefaultAction s) trayerQuery
+
+trayResizeHook
```
TheMC47, line 191: For the sake of consistency, this should be `trayResizeHook'` since it's more general (so move the ')

XMonad/Util/Hacks.hs
```diff
@@ -145,3 +158,49 @@ trayAbovePanelEventHook trayQ panelQ ConfigureEvent{ev_window = w, ev_above = a}
...
+  :: (Int -> X())     -- ^ action to take when query succeeds
+  -> Query Bool       -- ^ tray query
+  -> (Event -> X All) -- ^ event hook result
+trayResizeHook action trayQ ConfigureEvent { ev_window = w, ev_width = wa } = do
+  whenX (runQuery trayQ w) (
+    action $ fromIntegral wa
+    )
```
TheMC47, line 198:
```suggestion
  whenX (runQuery trayQ w) (action $ fromIntegral wa)
```

XMonad/Util/Hacks.hs
```diff
@@ -145,3 +158,49 @@ trayAbovePanelEventHook trayQ panelQ ConfigureEvent{ev_window = w, ev_above = a}
...
 trayAbovePanelEventHook _ _ _ = mempty
+
+-- $resizeTrayer
+-- Communicate tray (trayer) resize events to xmobar so that space may
+-- be reserved on xmobar to house the tray.
+--
+-- Basic Usage, demonstrating Trayer:
+-- First, add to handleEventHook as follows:
+--
+-- > handleEventHook = handleEventHook def <+> Hacks.trayerResizeHook
```
TheMC47, line 169:
```suggestion
-- > handleEventHook = handleEventHook def <> Hacks.trayerResizeHook
```

TheMC47 (changes requested):
Great, can you add documentation to the functions you added?

TheMC47:
> I tried adding an event like this to handle the trayer being destroyed, but it does not work
> 
> ```haskell
> trayResizeHook action trayQ DestroyWindowEvent { ev_window = w} = do  -- DOES NOT WORK
>   whenX (runQuery trayQ w) (  -- this query always fail but I am not clear why.  if i comment it out, the action to 0 does happen
>     action 0
>     )
>   return (All True)
> ```

Let's wait for someone who knows more about this

geekosaur:
By the time you get the `DestroyWindowEvent`, the window is already gone and it's too late for operations against that window to do anything. In particular, the usual `title`, `className`, `appName` queries will throw exceptions because the window isn't there for the corresponding properties (`WM_NAME`, `WM_CLASS`) to be queried.

XMonad/Util/Hacks.hs
tulth: There is no trayResizeHook already, so why not use trayResizehook instead of trayResizeHook'?  Note the confusing prefix of tray vs trayer.  The prefix change is intended to indicate the move away from something trayer-specific.

Are you sure it should be changed?

XMonad/Util/Hacks.hs
tulth: updated!

XMonad/Util/Hacks.hs
tulth: Updated!

tulth:
I pushed a new update that adds docs and addresses comments except for ' on trayeventhook name.

tulth:
> By the time you get the `DestroyWindowEvent`, the window is already gone and it's too late for operations against that window to do anything. In particular, the usual `title`, `className`, `appName` queries will throw exceptions because the window isn't there for the corresponding properties (`WM_NAME`, `WM_CLASS`) to be queried.

I was afraid this was the case.   What is the right course of action?  
- We could do nothing, because when a tray is restarted it will pick back up.  Is killing the tray and leaving it dead really a common use case to support?
- We could store something in extensible state, like a list of window IDs that are being used by trayResizes.  On destroy we could look up the window id, and if its in the list set the width to 0.  I kind of hate to add to state though if its not needed.

Any other ideas?

XMonad/Util/Hacks.hs
TheMC47: You're right, I overlooked it

TheMC47 (approved):

TheMC47:
> * What about: on destroy could we search all windows with the trayerQuery, and if it is not found we could send pixel width action to zero.

I like this. I don't think it's sensible to use more than one systray (is it? if I open a second tray, it doesn't receive anything). So I find this to be safe

slotThe:
On Sun, Nov 07 2021 09:13, tulth wrote:
> Any other ideas?

Perhaps we could watch for `UnmapEvent`s?  Normally this wouldn't work
but I'd imagine a tray or dock only gets unmapped when it's actually
being destroyed

liskin:
> Perhaps we could watch for `UnmapEvent`s? Normally this wouldn't work but I'd imagine a tray or dock only gets unmapped when it's actually being destroyed

Worth a try but I'd be worried that by the time we get to evaluating the Query the window's gone anyway.

> What about: on destroy could we search all windows with the trayerQuery, and if it is not found we could send pixel width action to zero.

This will be quite expensive I'm afraid. We had something like that in ManageDocks a couple years ago and had to get rid of it.

I don't have any other ideas myself unfortunately. I think it's probably okay to not handle it for the time being and just wait if anyone complains (possibly add a note about it in the docs or something).

liskin:
@TheMC47 Now that we have quite a lot of functions related to xmobar/trayer in X.U.Hacks, I'm wondering if it's perhaps time for these to graduate out of Hacks? I'm thinking `XMonad.Hooks.StatusBar.Tray` or something like that perhaps. What do you think?

I'd also consider simplifying the interface a bit. If we wanted to be consistent with the naming of the `trayerAboveXmobarEventHook` and `trayAbovePanelEventHook` functions, we could have e.g. `trayerPaddingXmobarEventHook` and `trayPaddingXmobarEventHook`and not export the other ones. It makes the docs quite complicated, for (I think) little benefit: it looks like premature abstraction—trying to make it general to cover usecases we don't have and might never have. Is there a status bar app other than xmobar that uses X props but a different markup? Probably not, so no need to have an API for that.

But yeah, this is just some final bikeshedding. Overall I love it, and I'm super glad that @tulth came up with this idea and listened to all our suggestions over the past few weeks. Good job indeed!

TheMC47:
> @TheMC47 Now that we have quite a lot of functions related to xmobar/trayer in X.U.Hacks, I'm wondering if it's perhaps time for these to graduate out of Hacks? I'm thinking `XMonad.Hooks.StatusBar.Tray` or something like that perhaps. What do you think?

Yes I was also thinking about it


> I'd also consider simplifying the interface a bit. If we wanted to be consistent with the naming of the `trayerAboveXmobarEventHook` and `trayAbovePanelEventHook` functions, we could have e.g. `trayerPaddingXmobarEventHook` and `trayPaddingXmobarEventHook`and not export the other ones. It makes the docs quite complicated, for (I think) little benefit: it looks like premature abstraction—trying to make it general to cover usecases we don't have and might never have. Is there a status bar app other than xmobar that uses X props but a different markup? Probably not, so no need to have an API for that.

Yeah it seems that I overdid it with the generalization (probably because I was thinking of the STM thing). Sure thing, let's clean it up. (we could merge it as it is and do the cleaning later, as you want @tulth)

tulth:
Ok, so what is the way forward?
I think i should:
- [ ] rename as described, ie ```trayerPaddingXmobarEventHook```
- [ ] ? not export the ```trayResizeHook```
- [ ] CHANGES.md file

liskin:
@tulth Yes, if you agree with my points, then please do the three things. You don't need to take the names I suggested, but I think mentioning xmobar and pad in the function name makes it easier to see what it's all about.  You can also simplify the implementation if you're only going to export two hooks (one specific to trayer and another one with arguments for the Query and possibly for the prop name). If at any point you feel like we're bikeshedding too much, we can merge it as it is and I (or @TheMC47 or someone) will clean it up when we get to it.

tulth:
I think I have checked the boxes.  Hoping its acceptable!

slotThe:
@liskin @TheMC47 this looks fine to me at this point; what do you say to merging after a rebase?

liskin:
@slotThe I wanted to clean it up a little more but then got caught up in life struggles so never got back to it. I managed to catch up a bit with xmonad stuff yesterday so I can try to take care of this tomorrow. Feel free to merge tomorrow evening otherwise. :-)

XMonad/Util/Hacks.hs
```diff
@@ -145,3 +157,74 @@ trayAbovePanelEventHook trayQ panelQ ConfigureEvent{ev_window = w, ev_above = a}
...
+-- | 'trayPaddingXmobarDefProp' is default property name,
+-- @"_XMONAD_TRAYPAD"@, to use with 'xmonadPropLog''
+trayPaddingXmobarDefProp :: String
+trayPaddingXmobarDefProp = "_XMONAD_TRAYPAD"
+
+-- | 'trayerPaddingXmobarEventHook' is a simple trayer-specific event
+-- hook that watches for trayer window resize changes and update the
+-- value in the property specified by 'trayPaddingXmobarDefProp'
+trayerPaddingXmobarEventHook
+  :: (Event -> X All) -- ^ event hook
```
TheMC47, line 195:
```suggestion
  :: Event -> X All -- ^ event hook
```

XMonad/Util/Hacks.hs
```diff
@@ -145,3 +157,74 @@ trayAbovePanelEventHook trayQ panelQ ConfigureEvent{ev_window = w, ev_above = a}
...
+trayerPaddingXmobarEventHook = trayerPaddingXmobarEventHook' trayPaddingXmobarDefProp
+
+-- | 'trayerPaddingXmobarEventHook'' is a more generic version of
+-- 'trayerPaddingXmobarEventHook' that allows the user to specify the
+-- property to use with 'xmonadPropLog'' when
+-- 'trayPaddingXmobarDefProp' is not desired.  This is still a
+-- trayer-specific hook.
+trayerPaddingXmobarEventHook'
+  :: String           -- ^ 'xmonadPropLog'' string to use
+  -> (Event -> X All) -- ^ event hook result
```
TheMC47, line 205:
```suggestion
  -> Event -> X All -- ^ event hook result
```

XMonad/Util/Hacks.hs
```diff
@@ -145,3 +157,74 @@ trayAbovePanelEventHook trayQ panelQ ConfigureEvent{ev_window = w, ev_above = a}
...
+
+-- | 'trayPaddingXmobarEventHook' is the fully generic tray resize hook.
+-- 'trayPaddingXmobarEventHook' is not trayer-specific; note the prefix is
+-- @tray@, not @trayer@.
+-- Both the action to take and the tray identification query are given
+-- as arguments.
+trayPaddingXmobarEventHook
+  :: (Int -> X())     -- ^ action to take when query succeeds, pixels to action
+  -> Query Bool       -- ^ query to identify the tray window
+  -> (Event -> X All) -- ^ event hook result
```
TheMC47, line 216:
```suggestion
  -> Event -> X All -- ^ event hook result
```

CHANGES.md
```diff
@@ -32,6 +32,13 @@
     - Flipped how `(^?)`, `(~?)`, and `($?)` work to more accurately
       reflect how one uses these operators.
 
+  * `XMonad.Util.Hacks`
+
+    - Added `trayerPaddingXmobarEventHook` to communicate tray (e.g.,
+      trayer) resize events to XMobar so that padding space may be
+      reserved on xmobar for the tray.  Requires XMobar version that
+      supports the `<hspace=X/>` feature.
```
TheMC47, line 40:
```suggestion
      reserved on xmobar for the tray.  Requires XMobar >= 0.40.
```

TheMC47 (approved):
Last remarks, no need to delay this further (`X.H.StatusBar.Tray` can definitely come later).

slotThe:
> @slotThe I wanted to clean it up a little more but then got caught up in life struggles so never got back to it. I managed to catch up a bit with xmonad stuff yesterday so I can try to take care of this tomorrow. Feel free to merge tomorrow evening otherwise. :-)

Taking you by your word here :)  Though as already discussed on IRC, from my point of view an API change while this is still not released is totally possible

liskin:
Took me a while, sorry, but here are the cleanups I propose: https://github.com/xmonad/xmonad-contrib/pull/683

Issue #638, opened 2021-10-31
Terminal with noborder doesn't come back after closing a swallowed app


syunsuke:
Hi everyone.

I thought I found a problem.
So I report the problem.

I've read the CONTRIBUTING.md
But I've never issued. so The way to issue may be wrong.
If so, would you tell me?

### Problem Description

When I use smartBorders of XMonad.Layout.NoBorders module 
and I set XMonad.Hooks.WindowSwallowing hook in my xmonad.hs,
I open "a" terminal and then call "chromium" on the terminal, 
so chromium opens well.

But I close the chromium, workspace is empty!! 

I think the terminal which called the chromium should come back,
but it goes away.

And acutually if I don't use smartBorders nor noborder,
the terminal comes back.

### Steps to Reproduce

I set smartborder and windowswallowing in my xmonad.hs

I opened one kitty terminal

I issued "chromium" on the terminal.
(or "nigrogen" my favorit wallpaper seting app)

The gui app was opened.

I closed the gui app.

That workspace was empty.

### Configuration File

Please include the smallest _full_ configuration file that reproduces
the problem you are experiencing:

```haskell
import XMonad
import XMonad.Layout.NoBorders
import XMonad.Hooks.WindowSwallowing

main = do
  xmonad 
    $ def { terminal          = "kitty"

          , handleEventHook   = myHandleEventHook
                                <+> handleEventHook def
          
          , layoutHook        = smartBorders $ Tall 1 0.03 0.5
          -- , layoutHook        = Tall 1 0.03 0.5
          }

---------------------------------------------------------------
-- WindowSwallowing
---------------------------------------------------------------
myHandleEventHook 
  = swallowEventHook (className =? "kitty" 
                 <||> className =? "Alacritty"
                 <||> className =? "XTerm") (return True)

```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17.0

slotThe:
[...]

PR #644, opened 2021-11-07
X.H.WindowSwallowing: Fix single window getting lost

slotThe:
### Description

ConfigureEvents may occur after a window has been deleted, an UnmapEvent
has already been sent (and thus xmonad already unmanaged the window),
but before a DestroyWindowEvent is caught by the eventHook.  For
example, this is the case when one uses smartBorders with a single
window (such that smartBorders is "active").  The ConfigureEvents
sensibly already have an empty stack (because the UnmapEvent has already
been received), which we then copy to the history.

Whenever a parent window has been found, the sensible thing to do is to
always restore it.  The fact that oldStack is Nothing simply encodes an
empty workspace and is thus something we definitely need to handle as
well.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/638

### Checklist

        manually, ...) and concluded: I've tested them manually and @syunsuke (who was the original reporter) also confirmed that theses changes work for them.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,11 @@
     - Added `transposeChars` to interchange the characters around the
       point and bound it to `C-t` in the Emacs XPKeymaps.
 
+  * `XMonad.Hooks.WindowSwallowing`
+
+    - Fixed windows getting lost when used in conjunction with
+      `smartBorders` and a single window.
+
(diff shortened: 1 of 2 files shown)


liskin (approved):
Looks good.

Issue #208, opened 2017-08-01
XMonad.Prompt does not support font fallback


NickHu:
### Problem Description

This is most noticable with the `XMonad.Prompt.Unicode` module. Many of the Unicode glyphs are not visible in any single font, and the way we would like to achieve coverage is by matching in sequence through a font preference list.
I would have expected something like

```haskell
myXPConfig = def {
  font = "xft:Terminus, xft:Noto Emoji"
}
```

to work, but this appears to instead crash XMonad when the prompt is raised.

I suspect font fallback is just unimplemented; `XMonad.Util.Font` does mention `fallBack`, but I don't think it has the semantics above.

Suggestions for fonts to use for `XMonad.Prompt.Unicode` would also be good - Symbola seems okay aside from its regular text being very ugly.

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [x] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

laMudri:
Should this be an issue on the XMonad core? In my experience, no part of XMonad supports fallback fonts, so I'm guessing the root of the problem is in XMonad itself.

geekosaur:
The core doesn't use fonts and therefore doesn't support them in any sense. xmonad-contrib can be built with support for X11-xft, but the contrib module would still need to be coded for fallback support; and if contrib wasn't built with xft support, fallback fonts aren't a thing. (This is all X11, not xmonad. It's somewhat ancient.)

People whose only exposure to X11 is indirect via Gtk or Qt mostly don't realize that those toolkits *exist* because X11 doesn't support all the fancy stuff they consider "fundamental", like fallback fonts.

Vonfry:
Any process on this issue? I have the same issue. The fallback fonts cannot work such as tab and prompt.

iliayar:
I've succeeded with moving [xmobar fallback fonts logic](https://github.com/jaor/xmobar/blob/master/src/Xmobar/X11/MinXft.hsc). Seems like it may add performance overhead. It also requires manual bindings from the Xft library, because https://hackage.haskell.org/package/X11-xft doesn't have binding for vital `XftCharExists` function. I still have some strange behavior with Prompt sometimes.

muellert:
[...]

PR #646, opened 2021-11-13
Fallback fonts

iliayar:
### Description

Related to #208. Moving [MinXft](https://github.com/jaor/xmobar/blob/master/src/Xmobar/X11/MinXft.hsc) library and [xmobar's fallback fonts logic](https://github.com/jaor/xmobar/blob/master/src/Xmobar/X11/Text.hs) to XMonad prompts, tabs labels, etc. So now, for example, prompts are able to display text and emoji. You can specify fonts to use through comma-separated list.  It also requires manual bindings from the Xft library, because https://hackage.haskell.org/package/X11-xft doesn't have binding for vital `XftCharExists` function.

### Checklist

### DIFF
--- a/XMonad/Util/Font.hs
+++ b/XMonad/Util/Font.hs
@@ -41,15 +41,16 @@ import Control.Exception as E
 import Text.Printf (printf)
 
 #ifdef XFT
-import Graphics.X11.Xft
+import qualified Data.List.NonEmpty as NE
 import Graphics.X11.Xrender
+import Graphics.X11.Xft
 #endif
 
 -- Hide the Core Font/Xft switching here
 data XMonadFont = Core FontStruct
                 | Utf8 FontSet
 #ifdef XFT
-                | Xft  XftFont
+                | Xft  (NE.NonEmpty XftFont)
 #endif
 
 -- $usage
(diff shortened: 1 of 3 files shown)


slotThe:
@iliayar we now have the CI set up for the X11-xft repo, so could you PR that portion of this against https://github.com/xmonad/X11-xft?  Thanks!

XMonad/Util/Font.hs
```diff
@@ -41,15 +41,15 @@ import Control.Exception as E
...
 import Graphics.X11.Xrender
+import Graphics.X11.Xft
 #endif
 
 -- Hide the Core Font/Xft switching here
 data XMonadFont = Core FontStruct
                 | Utf8 FontSet
 #ifdef XFT
-                | Xft  XftFont
+                | Xft  (NonEmpty XftFont)
```
slotThe, line 52: This is debatable, one might also argue that `[XftFont]` fits better with the rest of the module and functions; use your own judgement here

slotThe (commented):
@iliayar As a test, I've updated this PR to use the new X11-xft API (build failures are to be expected)

To use, add something like

```
  - location:
    git: https://github.com/xmonad/X11-xft.git
    commit: 6cc5fd928cc932728f3395242b6c8115a481f675
```

to your `stack.yaml` and set `allow-newer: true`.

One small thing I noticed is that the ascent calculation is a little too simplistic.  With only one font specified for the prompt (`font = "xft:iosevka-11"`) I get an expected

1font

However, if I specify an additional font (`font = "xft:iosevka ff-11,FontAwesome-9"`) then the extents are a bit screwed up:

2fonts

XMonad/Util/Font.hs
```diff
@@ -150,9 +159,9 @@ textExtentsXMF (Core fs) s = do
   let (_,a,d,_) = textExtents fs s
   return (a,d)
 #ifdef XFT
-textExtentsXMF (Xft xftfont) _ = io $ do
-  ascent  <- fi <$> xftfont_ascent  xftfont
-  descent <- fi <$> xftfont_descent xftfont
+textExtentsXMF (Xft xftfonts) _ = io $ do
+  ascent  <- fi <$> xftfont_max_ascent xftfonts
+  descent <- fi <$> xftfont_min_descent xftfonts
```
slotThe, line 164: Actually, isn't the descent also always positive?  Wouldn't we want the maximal descent here so that we can nicely center text in e.g. the prompt (indeed, I just tried that and it works)?

slotThe: 140a6d27a07a8a5bcb3eec24359ffb5b9ce37665

xmonad-contrib.cabal
```diff
@@ -80,6 +80,7 @@ library
        ghc-options: -Wwarn=unused-imports
 
     if flag(use_xft)
+        extra-libraries: Xrandr Xrender
```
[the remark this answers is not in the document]
slotThe: You're totally right, that was still from MinXft

slotThe:
Thanks!

Issue #146, opened 2017-02-12
Unsafe `getWindowAttributes` calls in ~20 modules


LadyBoonami:
I've been running into some xmonad crashes lately, but they've been pretty hard to isolate and reproduce. All of them quit with the same `user error in getWindowAttributes` message, so I've decided to grep for `getWindowAttributes` through both `xmonad` and `xmonad-contrib`. Here's what I've come up with:

* Multiple comments hint towards `getWindowAttributes` being an unsafe function. It is implemented as a wrapper around an `unsafe` `foreign` C function. However, its [documentation](http://hackage.haskell.org/package/X11-1.8/docs/Graphics-X11-Xlib-Extras.html#v:getWindowAttributes) doesn't mention anything. (Maybe I should open an issue there as well?)
* The `xmonad` code was mostly fine, only one unsafe call I fixed in xmonad/xmonad#76.
* Out of 29 `getWindowAttributes` calls spread over 21 modules in `xmonad-contrib`, only two catch the exception (details further down below).

Considering the amount of changes necessary to safely wrap all those, I wanted to ask for advice on how to proceed before doing anything on my own.

P.S.: Sorry for not using your issue template, but it didn't seem appropriate for the kind of problem.

---

Sorted + filtered output of `grep -Hn getWindowAttributes **/*.hs`

GOOD:

    XMonad/Actions/UpdatePointer.hs:71:        Just w  -> do tryAttributes <- io $ try $ getWindowAttributes dpy w
    XMonad/Util/DebugWindow.hs:158:-- Graphics.X11.Extras.getWindowAttributes is bugggggggy

BAD:

    XMonad/Actions/ConstrainedResize.hs:48:    wa <- io $ getWindowAttributes d w
    XMonad/Actions/FlexibleManipulate.hs:83:    [wpos, wsize] <- io $ getWindowAttributes d w >>= return . winAttrs
    XMonad/Actions/FlexibleResize.hs:54:    wa <- io $ getWindowAttributes d w
    XMonad/Actions/FloatKeys.hs:47:    wa <- io $ getWindowAttributes d w
    XMonad/Actions/FloatKeys.hs:65:    wa <- io $ getWindowAttributes d w
    XMonad/Actions/FloatKeys.hs:115:    wa <- io $ getWindowAttributes d w
    XMonad/Actions/FloatSnap.hs:97:    wa <- io $ getWindowAttributes d w
[...]

PR #647, opened 2021-11-13
Prefer safe alternatives to getWindowAttributes

slotThe:
### Description

Since X11 is unlikely to get safe bindings anytime soon, I figured this was better than nothing.

##### X.Prelude: Add safeGetWindowAttributes

Move the function from X.U.DebugWindow, where it was defined already.
This is a safe version of getWindowAttributes, returning a Maybe instead
of throwing an exception, in case the window attributes could not be
retrieved.

##### Prefer safe alternatives to getWindowAttributes

Whenever possible, prefer the safe wrappers withWindowAttributes or
safeGetWindowAttributes to getWindowAttributes.

Places where these are not applicable are limited to layouts, where
there is not good "default value" to give back in case these calls fail.
In these cases, we let the exception handling of the layout mechanism
handle it and fall back to the Full layout.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/146

### Checklist

        manually, ...) and concluded: it compiles...!

  I'm not sure this warrants an entry since it's not a user-facing change.

### DIFF
--- a/XMonad/Actions/ConstrainedResize.hs
+++ b/XMonad/Actions/ConstrainedResize.hs
@@ -44,8 +44,8 @@ import XMonad
 
 -- | Resize (floating) window with optional aspect ratio constraints.
 mouseResizeWindow :: Window -> Bool -> X ()
-mouseResizeWindow w c = whenX (isClient w) $ withDisplay $ \d -> do
-    wa <- io $ getWindowAttributes d w
+mouseResizeWindow w c = whenX (isClient w) $ withDisplay $ \d ->
+  withWindowAttributes d w $ \wa -> do
     sh <- io $ getWMNormalHints d w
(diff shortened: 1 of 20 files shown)


slotThe:
Will merge in a few days if no one objects

liskin:
I haven't had time to go through the code in detail but the general idea looks all right, so don't expect me to object. There's also nothing stopping you from adding the safe alternatives to X11 if you feel that's right, but it'd mean some painful dependency bumps so it'd have to be worth it long-term.

slotThe:
> There's also nothing stopping you from adding the safe alternatives to X11 if you feel that's right, but it'd mean some painful dependency bumps so it'd have to be worth it long-term.

Exactly; I think I'd only do this if I was actually planning on a safe API for X11 but I'm not really going to do anything close to that anytime soon

PR #651, opened 2021-11-21
add .desktop prompt launcher

tulth:
### Description

Add a prompt bar that can launch `.desktop` files.

Example use
```haskell
import XMonad.Prompt.DotDesktop ( appLaunchPrompt )

myDotDeskTopLaunch = appLaunchPrompt (
  myXPConfig { complCaseSensitivity = CaseInSensitive
             , searchPredicate = fuzzyMatch
             , sorter = fuzzySort
             , maxComplRows = Just 2
             })

myKeymap :: [(String, X ())]
myKeymap =
  [ ("M-M1-d", myDotDeskTopLaunch),
...
```

### Checklist

### DIFF
--- a/XMonad/Prompt/DotDesktop.hs
+++ b/XMonad/Prompt/DotDesktop.hs
@@ -1,0 +1,146 @@
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE LambdaCase #-}
+module XMonad.Prompt.DotDesktop
+    ( appLaunchPrompt
+    ) where
+
+import XMonad ( spawn, io, X )
+import XMonad.Prompt ( mkXPrompt, XPConfig(searchPredicate) )
+import XMonad.Prompt.Shell ( Shell(Shell), split )
+import XMonad.Prompt.DotDesktopParser ( runDotDesktopParser )
+
+import qualified Data.Map as M
+import Control.Applicative ( Alternative((<|>)) )
+import Control.Monad (filterM)
+import Control.Monad.Except
+    ( runExceptT, ExceptT (ExceptT), liftEither )
+import Control.Exception ( try, Exception )
+import Data.Functor ( (<&>) )
+import Data.List ( isSuffixOf, dropWhileEnd )
+import Data.Maybe ( fromMaybe, maybeToList, listToMaybe )
+import System.Directory (listDirectory, doesDirectoryExist)
+import System.Environment ( lookupEnv )
+import System.FilePath ((</>))
+
+import Data.Char (isSpace)
+import Data.Either (rights, lefts)
+import XMonad.Prelude (join)
+
+isDotDesktop :: FilePath -> Bool
+isDotDesktop = isSuffixOf ".desktop"
+
+trimWhitespace :: String -> String
+trimWhitespace = dropWhileEnd isSpace . dropWhile isSpace
+
+cmdFilter :: String -> String  -- fixme future do something other than dropping these
+cmdFilter ('%':'f':xs) = cmdFilter xs
+cmdFilter ('%':'F':xs) = cmdFilter xs
+cmdFilter ('%':'u':xs) = cmdFilter xs
+cmdFilter ('%':'U':xs) = cmdFilter xs
+cmdFilter ('%':'c':xs) = cmdFilter xs
+cmdFilter ('%':'k':xs) = cmdFilter xs
+cmdFilter ('%':'i':xs) = cmdFilter xs
+cmdFilter ('%':'%':xs) = '%' : cmdFilter xs
+cmdFilter (x:xs) = x : cmdFilter xs
+cmdFilter "" = ""
+
+convertExceptionToString :: Exception e => IO (Either e a) -> IO (Either String a)
+convertExceptionToString = fmap convertExceptionToStringHelper
+
+convertExceptionToStringHelper :: Exception e => Either e a -> Either String a
+convertExceptionToStringHelper = either (Left . convertExceptionToStringHelperHelper) Right
+
+convertExceptionToStringHelperHelper :: Exception e => e -> String
+convertExceptionToStringHelperHelper = show :: Exception e => e -> String
+
+doReadFileLBS :: String -> ExceptT String IO String
+doReadFileLBS = ExceptT . convertExceptionToString . try @IOError . readFile
+
+getVal :: String -> String -> M.Map String String -> Either String String
+getVal msg k kvmap = maybeToEither msg $ M.lookup k kvmap
+
+maybeToEither :: b -> Maybe a -> Either b a
+maybeToEither _ (Just a) = Right a
+maybeToEither b Nothing = Left b
+
+doParseFile :: String -> ExceptT String IO DotDesktopApp
+doParseFile filePath = do
+  content <- doReadFileLBS filePath
+  parsed <- liftEither $ runDotDesktopParser content
+  let kvMaybe = snd <$> listToMaybe (rights parsed)
+  keyVals <- liftEither $
+    maybe
+      (Left $ "Parse Resulted in no KeyVals in file " ++ filePath)
+      Right
+      kvMaybe
+  let errMsg = "Unable to find Name in file " ++ filePath
+  nom <- liftEither $ getVal errMsg "Name" keyVals
+  exc <- liftEither $ getVal errMsg "Exec" keyVals
+  typ <- liftEither $ getVal errMsg "Type" keyVals
+  return DotDesktopApp { fileName = filePath
+                       , name = nom
+                       , type_ = typ
+                       , exec = exc
+                       , cmd = (trimWhitespace . cmdFilter) exc
+                       }
+
+data DotDesktopApp = DotDesktopApp { fileName :: String
+                             , name :: String
+                             , type_ :: String
+                             , exec :: String
+                             , cmd :: String
+                             } deriving Show
+
+getXdgDataHome :: IO (Maybe FilePath)
+getXdgDataHome = do
+  envXdgDataHome <- envXdgDataHomeIO
+  defaultXdgDataHome <- defaultXdgDataHomeIO
+  return $ envXdgDataHome <|> defaultXdgDataHome
+  where
+    defaultXdgDataHomeIO = lookupEnv "HOME" <&> fmap (</> ".local" </> "share")
+    envXdgDataHomeIO = lookupEnv "XDG_DATA_HOME"
+
+getXdgDataDirs :: IO [FilePath]
+getXdgDataDirs =
+  fromMaybe defaultXdgDataDirs <$> envXdgDataDirsIO
+  where
+    defaultXdgDataDirs = split ':' "/usr/local/share:/usr/share"
+    envXdgDataDirsIO = lookupEnv "XDG_DATA_DIRS" <&> (<&> split ':')
+
+getAppFolders :: IO [FilePath]
+getAppFolders = do
+  xdgDataHome <- maybeToList <$> getXdgDataHome
+  xdgDataDirs <- getXdgDataDirs
+  let possibleAppDirs = xdgDataHome ++ xdgDataDirs <&> (</> "applications")
+  filterM doesDirectoryExist possibleAppDirs
+
+getDirContents :: FilePath -> ExceptT String IO [FilePath]
+getDirContents dir = do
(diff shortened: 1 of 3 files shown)


tulth:
I am interested in reactions to this idea and if there is a better way to go about it.

I am very happy with this in my setup.  It is particularly useful for launching steam games.

XMonad/Prompt/DotDesktopParser.hs
```diff
@@ -0,0 +1,171 @@
...
+module XMonad.Prompt.DotDesktopParser
+    ( runDotDesktopParser
+    ) where
+
+import Data.Maybe ( catMaybes )
+import Control.Monad ( MonadPlus(..) )
+import Control.Applicative ( Alternative(..) )
+import qualified Data.Map as MAP
+
+newtype Parser a = Parser { parse :: String -> [(a,String)] }
```
slotThe, line 12: There already is quite a capable parser in `base` called [ReadP](https://hackage.haskell.org/package/base-4.16.0.0/docs/Text-ParserCombinators-ReadP.html).  Some modules, like [X.P.OrgMode](https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/src/XMonad.Prompt.OrgMode.html#line-355), already use that and so it would be nice if this could as well.  

I wouldn't be opposed to a general X.U.Parser though, since the commutative nature of `ReadP` ironically makes some things harder (see e.g. [here](https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/src/XMonad.Prompt.OrgMode.html#endBy1))

XMonad/Prompt/DotDesktop.hs
```diff
@@ -0,0 +1,146 @@
...
+  where
+    defaultXdgDataHomeIO = lookupEnv "HOME" <&> fmap (</> ".local" </> "share")
+    envXdgDataHomeIO = lookupEnv "XDG_DATA_HOME"
+
+getXdgDataDirs :: IO [FilePath]
+getXdgDataDirs =
+  fromMaybe defaultXdgDataDirs <$> envXdgDataDirsIO
+  where
+    defaultXdgDataDirs = split ':' "/usr/local/share:/usr/share"
+    envXdgDataDirsIO = lookupEnv "XDG_DATA_DIRS" <&> (<&> split ':')
```
slotThe, line 108: We already depend on `directory` and that module has quite capable handling of home and [xdg dirs](https://hackage.haskell.org/package/directory-1.3.7.0/docs/System-Directory.html#g:3)

XMonad/Prompt/DotDesktop.hs
```diff
@@ -0,0 +1,146 @@
...
+
+getAppFolders :: IO [FilePath]
+getAppFolders = do
+  xdgDataHome <- maybeToList <$> getXdgDataHome
+  xdgDataDirs <- getXdgDataDirs
+  let possibleAppDirs = xdgDataHome ++ xdgDataDirs <&> (</> "applications")
+  filterM doesDirectoryExist possibleAppDirs
+
+getDirContents :: FilePath -> ExceptT String IO [FilePath]
```
slotThe, line 117: I don't think ExceptT is really necessary here, since all you do with is is run it immediately after calling the function.  Why not just return an `IO (Either String [FilePath])` instead?

XMonad/Prompt/DotDesktop.hs
```diff
@@ -0,0 +1,146 @@
...
+cmdFilter ('%':'f':xs) = cmdFilter xs
+cmdFilter ('%':'F':xs) = cmdFilter xs
+cmdFilter ('%':'u':xs) = cmdFilter xs
+cmdFilter ('%':'U':xs) = cmdFilter xs
+cmdFilter ('%':'c':xs) = cmdFilter xs
+cmdFilter ('%':'k':xs) = cmdFilter xs
+cmdFilter ('%':'i':xs) = cmdFilter xs
+cmdFilter ('%':'%':xs) = '%' : cmdFilter xs
+cmdFilter (x:xs) = x : cmdFilter xs
+cmdFilter "" = ""
```
slotThe, line 45: Well this doesn't look fun :)  I'm not familiar with `.desktop` files at all; what exactly are you doing here?

XMonad/Prompt/DotDesktop.hs
```diff
@@ -0,0 +1,146 @@
...
+
+convertExceptionToString :: Exception e => IO (Either e a) -> IO (Either String a)
+convertExceptionToString = fmap convertExceptionToStringHelper
+
+convertExceptionToStringHelper :: Exception e => Either e a -> Either String a
+convertExceptionToStringHelper = either (Left . convertExceptionToStringHelperHelper) Right
+
+convertExceptionToStringHelperHelper :: Exception e => e -> String
+convertExceptionToStringHelperHelper = show :: Exception e => e -> String
```
slotThe, line 54:
```suggestion
exceptionToString :: Exception e => Either e a -> Either String a
exceptionToString = either (Left . show) Right
```

optionally, keep the `IO` in there if you link that better

XMonad/Prompt/DotDesktop.hs
tulth: I want to just execute the .desktop command specified, without arguments.
Per the spec here:
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables
there are many field codes for arguments which I cannot pass directly to the shell to execute, so I just remove them.

geekosaur: You might want to see if using `xdg-open` to run it gives you more options.

XMonad/Prompt/DotDesktop.hs
tulth: Agreed.  I pulled on this and ended up removing all ExceptT

XMonad/Prompt/DotDesktop.hs
tulth: I used this suggestion, thanks!  much cleaner!

XMonad/Prompt/DotDesktopParser.hs
tulth: I like this idea!  I think will take me a while to re-implement as I am unfamiliar with ReadR but I will start on it.

XMonad/Prompt/DotDesktop.hs
tulth: I do not understand how I would use xdg-open?  Could you provide an example?

Just for context, here is an example from `gimp.desktop`:
```
[Desktop Entry]
Version=1.0
Type=Application
Name=GNU Image Manipulation Program
...
Exec=gimp-2.10 %U
...
```
Basically the user could type `image`, fuzzyMatch finds and displays `GNU Image Manipulation Program`, and if it is selected, the Exec is `gimp-2.10 %U`.  This code filters out the %U and passes `gimp-2.10` to the shell to launch.

liskin: `gtk-launch` or `kioclient5 exec` can be used to run a .desktop file: https://askubuntu.com/a/1114798/950919

tulth: ah, I get it.  I did not realize you could just xdg-open the .desktop file.

What do all think?  Just use the exec with the cmd filter or use xdg-open or gtk-launch?

liskin: Well it's a bit silly though. `xdg-open` doesn't run the desktop file, it opens it in some app. `gtk-launch` works, but it expects just `gimp.desktop`, not `/usr/share/applications/gimp.desktop`. `kioclient3 exec`, on the other hand, only works with `/usr/share/applications/gimp.desktop` but not with `gimp.desktop`. It's just chaos. :-(

XMonad/Prompt/DotDesktopParser.hs
tulth: re-implemented this using ReadP.  While it is working for me, I probably need to test this more.

XMonad/Prompt/DotDesktop.hs
slotThe: `gtk-launch` is (well, at least on my distro) installed with the `gtk` package itself, so chances are users will definitely have this installed.

Will the fact that it goes into `/usr/share/applications` by itself be a problem?  Out of the current paths that the code searches (`~/.local/share/applications`, `/usr/share/applications`, and `/usr/local/share/applications`) on the first one seems not to get searched by `gtk-launch`.

liskin: `gtk-launch` does look into `~/.local/share/applications` here.

slotThe: Mh, doesn't work on my end; a `.desktop` file that's only in `~/.local/share/applications` but not in any of the other directories will fail to be found

tulth:
> Mh, doesn't work on my end; a .desktop file that's only in ~/.local/share/applications but not in any of the other directories will fail to be found

Arg!  How strange.  I introduced some broken desktop files and logging locally to double check the behavior on my end and I cannot seem to reproduce the issue yet.

logging diffs:
```diff
modified   XMonad/Prompt/DotDesktop.hs
@@ -54,6 +54,8 @@ getDirContents dir = do
 getDotDesktopApps :: IO [DotDesktopApp]
 getDotDesktopApps = do
   appFolders <- getAppFolders
+  appendFile "/home/tulth/.xmonad/debug.log" $
+    unlines $ ("appfolder: " ++ ) <$> appFolders
   contentsPerFolder <- mapM getDirContents appFolders
   let folderFiles = join $ rights contentsPerFolder
       dotDesktopFiles = filter isDotDesktop folderFiles
@@ -63,6 +65,7 @@ getDotDesktopApps = do
   let parseErrs = lefts parseResults
       dotDesktopApps = rights parseResults
   mapM_ print parseErrs
+  appendFile "/home/tulth/.xmonad/debug.log" $ unlines parseErrs
   return dotDesktopApps
```

I thought perhaps a parse failure in a folder would cause a bigger problem, but that was no issue on my system.
It give the expected result on my system and user and system desktops are listed, and the broken files i created show up.
debug.log:
```
appfolder: /home/tulth/.local/share/applications
appfolder: /usr/share/applications
Parse Resulted in no KeyVals in file /home/tulth/.local/share/applications/broken_user.desktop
Parse Resulted in no KeyVals in file /usr/share/applications/broken_system.desktop
```

Any ideas?  Is `XDG_DATA_DIRS` is set or no?  On my system `XDG_DATA_DIRS` is not set and it picks up defaults.

slotThe:
> Any ideas? 

Sorry, I was referring to `gtk-launch` in that comment, not your implementation

XMonad/Prompt/DotDesktop.hs
liskin: That's really weird, I looked into the source of gtk/glib and I don't see why it wouldn't look there. Maybe try stracing it?

slotThe: Oh indeed, `strace` tells me that the directory *is* searched.  Upon closer inspection, the `.desktop` file contains a non-existant executable; instead of telling me that, `gtk-launch` told me that the `.desktop` file couldn't be found, however ._.

slotThe:
@tulth what do you think about the idea of using `gtk-launch` instead?  Semms like a standard tool that people will definitely have installed and would simplify this quite a bit

tulth:
> @tulth what do you think about the idea of using `gtk-launch` instead? Semms like a standard tool that people will definitely have installed and would simplify this quite a bit

I was thinking about this as well, but I do not think it makes much difference.
The .desktop files still have to be parsed to find the names and the filter out only applications.  The only thing it saves is filtering out the %X patterns.

slotThe:
> I was thinking about this as well, but I do not think it makes much difference. The .desktop files still have to be parsed to find the names and the filter out only applications. The only thing it saves is filtering out the %X patterns.

I don't use desktop files, so forgive my ignorance here, but the man page for `gtk-launch` says that it "launches an application using the given name".  Does that mean there are `.desktop` files for things that aren't executable programs?  What would those be?  Wouldn't `gtk-launch` also ignore them or does it do other things as well?

geekosaur:
I note that `gtk-launch` also knows how to activate an already running application instead of just starting a new one. (Of course, starting a new one should also do that.) This does, however, imply that some `.desktop` files may contain DBus activations instead of executables.

slotThe:
@tulth any updates?

slotThe:
@tulth friendly ping

PR #653, opened 2021-11-22
Fix memory leaks on `historyHook` and `workspaceHistoryHook`

RubenAstudillo:
### Description

Both of these hooks leaked memory on long running sessions.
per-cost-center

To fix it I had to force the thunks stored on the hooks as the compiler wouldn't be able to see where they would be demanded as they depended on the user interaction. Now the graph looks like this.
fixed-leak

The most polemic change on this PR is the inclusion of the `parallel` library as a dependency. I needed for the `Control.Seq` module which gave me combinators to force just enough of the `Data.Seq` and `[]` data types.

### Checklist

  - [🗸] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [🗸] I've considered how to best test these changes (property, unit,
        manually, ...) and concluded: It shows the memory improvements!

  - [🗸] I updated the `CHANGES.md` file (None)

### DIFF
--- a/XMonad/Util/ExtensibleState.hs
+++ b/XMonad/Util/ExtensibleState.hs
@@ -94,7 +94,7 @@ modify f = put . f =<< get
 -- | Like @modify@ but the result value is applied strictly in respect to
 -- the monadic environment.
 modify' :: (ExtensionClass a, XLike m) => (a -> a) -> m ()
-modify' f = (\a -> let res = f a in res `seq` put res) =<< get
+modify' f = (put $!) . f =<< get
 
 -- | Add a value to the extensible state field. A previously stored value with the same
 -- type will be overwritten. (More precisely: A value whose string representation of its type
--- a/XMonad/Util/ExtensibleState.hs
+++ b/XMonad/Util/ExtensibleState.hs
@@ -20,6 +20,7 @@ module XMonad.Util.ExtensibleState (
                               -- $usage
                               put
                               , modify
+                              , modify'
                               , remove
                               , get
                               , gets
@@ -89,6 +90,11 @@ modifyStateExts f = State.modify $ \st -> st { extensibleState = f (extensibleSt
 -- is none.
 modify :: (ExtensionClass a, XLike m) => (a -> a) -> m ()
 modify f = put . f =<< get
+
+-- | Like @modify@ but the result value is applied strictly in respect to
+-- the monadic environment.
+modify' :: (ExtensionClass a, XLike m) => (a -> a) -> m ()
+modify' f = (\a -> let res = f a in res `seq` put res) =<< get
 
 -- | Add a value to the extensible state field. A previously stored value with the same
 -- type will be overwritten. (More precisely: A value whose string representation of its type
--- a/XMonad/Actions/GroupNavigation.hs
+++ b/XMonad/Actions/GroupNavigation.hs
@@ -1,3 +1,4 @@
+{-# language DeriveGeneric, DeriveAnyClass #-}
 ----------------------------------------------------------------------
 -- |
 -- Module      : XMonad.Actions.GroupNavigation
@@ -34,13 +35,14 @@ module XMonad.Actions.GroupNavigation ( -- * Usage
 
 import Control.Monad.Reader
 import Control.Monad.State
-import Control.Seq (using, seqFoldable, rseq)
+import Control.DeepSeq
 import Data.Map ((!))
 import qualified Data.Map as Map
 import Data.Sequence (Seq, ViewL (EmptyL, (:<)), viewl, (<|), (><), (|>))
 import qualified Data.Sequence as Seq
 import qualified Data.Set as Set
 import Graphics.X11.Types
+import GHC.Generics
 import Prelude hiding (concatMap, drop, elem, filter, null, reverse)
 import XMonad.Core
 import XMonad.ManageHook
@@ -157,7 +159,7 @@ orderedWorkspaceList ss wsids = rotateTo isCurWS wspcs'
 -- The state extension that holds the history information
 data HistoryDB = HistoryDB (Maybe Window) -- currently focused window
                            (Seq Window)   -- previously focused windows
-               deriving (Read, Show)
+               deriving (Read, Show, Generic, NFData)
 
 instance ExtensionClass HistoryDB where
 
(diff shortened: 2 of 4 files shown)


slotThe (commented):
This is neat, thanks!

I wonder if we could get away with just stictifying the respective data structures (since neither is exported); i.e.,

```haskell
data SP a b = SP !a !b
  deriving (Read, Show)

newtype WorkspaceHistory = WorkspaceHistory
  { history :: [StrictPair ScreenId WorkspaceId]
    -- ^ Workspace Screens in reverse-chronological order.
  } deriving (Read, Show)

---

data HistoryDB = HistoryDB !(Maybe Window) -- currently focused window
                           !(Seq Window)   -- previously focused windows
               deriving (Read, Show)
```

This is still not ultra-strict but since we're not really interesting in forcing things more than to WHNF it should be fine I think.  We may or may not still need the odd call to `seq` to force the list, not entirely sure about that right now

EDIT: Oh, and 

>     * [🗸] I've considered how to best test these changes (property, unit,
>       manually, ...) and concluded: XXX

:)

RubenAstudillo:
> I wonder if we could get away with just stictifying the respective data structures (since neither is exported); i.e.,
> 
> ```haskell
> data SP a b = SP !a !b
>   deriving (Read, Show)
> 
> newtype WorkspaceHistory = WorkspaceHistory
>   { history :: [StrictPair ScreenId WorkspaceId]
>     -- ^ Workspace Screens in reverse-chronological order.
>   } deriving (Read, Show)
> 
> ---
> 
> data HistoryDB = HistoryDB !(Maybe Window) -- currently focused window
>                            !(Seq Window)   -- previously focused windows
>                deriving (Read, Show)
> ```
> 
> This is still not ultra-strict but since we're not really interesting in forcing things more than to WHNF it should be fine I think. We may or may not still need the odd call to `seq` to force the list, not entirely sure about that right now
> 

We are interested on forcing things to more than WHNF. On the `WorkspaceHistory` patch I needed to force at least until each element of each pair on the `history` list. WHNF on that data structure would be the first `cons` of that list, not enough.

On a related note, given that we are concerned with forcing evaluation until certain "depth", using strict data types as `StrictPair` as above is not enough. We would also need:
- A strict list container so we match to WHNF any `StrictPair`.
- Match on the `WorkspaceHistory` to WHNF before the `put` calls. As it is a newtype, it doesn't have extra bottoms. But we still need its head matched, so it forces the strict list and then the `StrictPair`.

So, instead of doing those two things, we can use the combinators on `Control.Seq` and match before the `put` call to get the correct behavior.

The same is true in respect to `HistoryDB`. We would need a strict `Seq` data structure. A bang pattern as `!(Seq Window)` is not enough to deal with this leak as it forces only the head. Every other element remains as a thunk until a undefined point in the future. 

> EDIT: Oh, and
> > ```
> > * [🗸] I've considered how to best test these changes (property, unit,
> >   manually, ...) and concluded: XXX
> > ```

Thanks. I should have put something on those XXX. I edited that part on the original post.

RubenAstudillo:
I was missing the `parallel` dependency for the test build. Allow the workflows to be re run.

liskin:
> The most polemic change on this PR is the inclusion of the `parallel` library as a dependency. I needed for the `Control.Seq` module which gave me combinators to force just enough of the `Data.Seq` and `[]` data types.

Would https://hackage.haskell.org/package/deepseq work here as well? Both seem to be maintained by the Core Libraries Committee but deepseq is distributed with GHC so adding that dependency is easier.

RubenAstudillo:
Yeah, we are basically doing the same as in `deepseq` now. It is a better dependency. I will use that, re-run my tests and publish an update.

RubenAstudillo:
@liskin using `deepseq` exposes other tradeoffs. The `NFData` typeclass instance has to be derived for `WorkspaceHistory`. This can be done on two ways per the [docs](https://hackage.haskell.org/package/deepseq-1.4.6.1/docs/Control-DeepSeq.html)

1.  Derive `Generic, Generic1` and then `NFData`. The `Generic`s instances are just to for using the derive anyclass instance for `NFData`. But these are heavyweigh for just forcing.
2. Use `GeneralizedNewtypeDeriving` for deriving `NFData`. This requires a `NFData` instance on `ScreenId` on the main `xmonad` project.

I consider both of these alternatives worse than the current status. Specially considering `ScreenId` is a **newtype**, so it doesn't introduce an extra bottom between it and the base `Int` it contains. So the strictness of `ScreenId` is equivalent to a shallow `seq`. I use this on the current code but it is not apparent for the `NFData` deriving mechanism.

I vote for keep using `parallel` instead of `deepseq`. Thought?

slotThe:
> I vote for keep using `parallel` instead of `deepseq`. Thoughts?

While `GeneralizedNewtypeDeriving` is probably a no-go as we would like to keep compatibility with xmonad 0.17.0 a bit longer than a few weeks (well, I would, anyways), I don't really see a problem with deriving `Generic`.  It's not like this will affect a lot of data structures such that actual compile times are affected.

> I consider both of these alternatives worse than the current status. Specially considering `ScreenId` is a **newtype**, so it doesn't introduce an extra bottom between it and the base `Int` it contains. So the strictness of `ScreenId` is equivalent to a shallow `seq`. I use this on the current code but it is not apparent for the `NFData` deriving mechanism.

While this is true, I think the _idea_ of what we want to achieve is expressed much more clearly by "this is an instance of NFData".

RubenAstudillo:
@slotThe You are right in that the intent is more clear with an `NFData` instance!
fixed-leak-deepseq
The memory figures are basically the same. I still had to do an instance by hand but the `liftRnf/liftRnf2` and `rwhnf` from `deepseq` where useful. Can you give permission to re-run the GH workflows?

RubenAstudillo:
I think is merge ready.

XMonad/Util/ExtensibleState.hs
```diff
@@ -90,6 +91,11 @@ modifyStateExts f = State.modify $ \st -> st { extensibleState = f (extensibleSt
 modify :: (ExtensionClass a, XLike m) => (a -> a) -> m ()
 modify f = put . f =<< get
 
+-- | Like @modify@ but the result value is applied strictly in respect to
+-- the monadic environment.
+modify' :: (ExtensionClass a, XLike m) => (a -> a) -> m ()
+modify' f = (\a -> let res = f a in res `seq` put res) =<< get
```
slotThe, line 97: This is pretty much the same as `(put $!) . f =<< get` right?

slotThe (approved):
LGTM now!

XMonad/Util/ExtensibleState.hs
RubenAstudillo: Only if `a` is an instance of `NFData`, `$!` uses `deepseq` instead of `seq`.

RubenAstudillo: It is! I am changing it :+1:

RubenAstudillo:
Using the lastest version with `$!` the memory profile is the same
fixed-leak-deepseq
Again, let the workflows re-run and I think is ready to merge. Thanks for all you patience and time @slotThe  and @liskin :muscle:  .

slotThe:
Thanks!

PR #659, opened 2021-11-30
New Module: XMonad.Util.Parser

slotThe:
### Description

Inspired in part by the first draft of https://github.com/xmonad/xmonad-contrib/pull/651 (where the need for it arose again, which was not helped by `ReadP`s lack of discoverability), this is a parser combinator module for xmonad; based on `ReadP`, but hopefully a bit more intuitive to use.  This is especially with regards to how other combinator libraries work, as well as the interplay with normal `Control.Applicative` functions.  Plus, `Semigroup` and `Monoid` instances :)

##### New module: XMonad.Util.Parser

This module provides a parser combinator library based on base's ReadP,
which aims to function more like other popular combinator libraries like
attoparsec and megaparsec.

In particular, the Alternative and Monoid instances are left-biased now,
so combinators like `many` and `optional` from Control.Applicative work
in a more intuitive manner.  Further, some functions (like `endBy1`)
only return the "most successful" parse, instead of returning all of
them.  We can now get away with providing a single parsing result
instead of ReadP's list of results (as such, parsers need to be
disambiguated earlier instead of trimming the list down after parsing).

##### X.P.OrgMode: Use X.U.Parser

Since we now have an "internal" parser library in xmonad, use it.  This
allows us to get rid of some hacks in this module that were needed
because of ReadP's parsing behaviour.

##### X.U.EZConfig: Use X.U.Parser

Using X.U.Parser works almost as a drop-in replacement for ReadP here.
In some places (like `parseSpecial`) we need to be a little bit more
careful when constructing the parser, but this is offset a much simpler
`readKeySequence`.

##### X.U.EZConfig: Add simple unit tests

Add very basic unit tests for EZConfig to see if it can parse all of the
keys (and key combinations) that it promises to parse.

The long-term goal here should be to write a pretty-printer for EZConfig
and to check whether that's a proper inverse (either in the normal sense
or in the inverse semigroup sense), as the tests for X.P.OrgMode do.

### Checklist

        manually, ...) and concluded:

          - Some tests for EZConfig are included (hopefully to be expanded later on),
          - X.P.OrgMode passes its existing tests.

### DIFF
--- a/XMonad/Util/Parser.hs
+++ b/XMonad/Util/Parser.hs
@@ -1,0 +1,215 @@
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE InstanceSigs               #-}
+{-# LANGUAGE TypeApplications           #-}
+--------------------------------------------------------------------
+-- |
+-- Module      : XMonad.Util.Parser
+-- Description : A parser combinator library for xmonad
+-- Copyright   : (c) 2021  slotThe <soliditsallgood@mailbox.org>
+-- License     : BSD3
+-- Maintainer  : slotThe <soliditsallgood@mailbox.org>
+-- Stability   : experimental
+-- Portability : non-portable
+--
+-- A small wrapper around the 'ReadP' parser combinator in @base@,
+-- providing a more intuitive behaviour.  While it's theoretically nice
+-- that 'ReadP' is actually commutative, this makes a lot of parsing
+-- operations rather awkward—more often than not, one only wants the
+-- argument that's parsed "first".
+--
+-- Due to the left-biased nature of the chosen semigroup implementation,
+-- using functions like 'many' or 'optional' from "Control.Applicative"
+-- now yields more consistent behaviour with other parser combinator
+-- libraries.
+--
+--------------------------------------------------------------------
+module XMonad.Util.Parser (
+  -- * Usage
+  -- $usage
+
+  -- * Running
+  Parser,
+  runParser,
+
+  -- * Primitive Parsers
+  eof,
+  num,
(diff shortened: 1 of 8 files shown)


tests/EZConfig.hs
```diff
@@ -0,0 +1,39 @@
...
+
+regularKeys :: ([String], [KeySym])
+regularKeys = unzip . map (first (: ""))
+            $ zip ['!'    .. '~'   ] [xK_exclam       .. xK_asciitilde]
+           ++ zip ['\xa0' .. '\xff'] [xK_nobreakspace .. xK_ydiaeresis]
+
+-- | QuickCheck can handle the 8! combinations just fine.
+modifiers :: [String]
+modifiers = map concat $
+  permutations ["M-", "C-", "S-", "M1-", "M2-", "M3-", "M4-", "M5-"]
```
slotThe, line 36: This may or may not be a bit overkill, but even on my stone-age laptop `stack test` doesn't take more than ~1.1 seconds or so for the actual tests

XMonad/Util/Parser.hs
```diff
@@ -0,0 +1,215 @@
...
+
+-- | Parse the first one or more characters satisfying the predicate.
+-- Fails if none, else succeeds exactly once having consumed all the
+-- characters.  Hence, this is __not__ the same as @many1 (satisfy p)@.
+munch1 :: (Char -> Bool) -> Parser String
+munch1 = coerce ReadP.munch1
+
+-- | @endBy p sep@ parses zero or more occurrences of @p@, separated and
+-- ended by @sep@.
+endBy :: ReadP a -> ReadP sep -> ReadP [a]
```
TheMC47, line 195: Any reason this isn't `Parser a -> Parser sep -> Parser [a]`?

slotThe: There is a perfectly valid reason—I'm a dunce :)

PR #660, opened 2021-12-06
X.H.BorderPerWindow: Remove unused import

exorcist365:
### Description

Unused imports bad

### Checklist

### DIFF
--- a/XMonad/Hooks/BorderPerWindow.hs
+++ b/XMonad/Hooks/BorderPerWindow.hs
@@ -23,7 +23,6 @@ module XMonad.Hooks.BorderPerWindow ( -- * Usage
                                     ) where
 
 
-import Graphics.X11.Xlib (Dimension)
 import XMonad
 import XMonad.Util.ActionQueue (enqueue, actionQueue)
 


slotThe:
Mh, I wonder why the CI didn't catch this?

liskin:
> Mh, I wonder why the CI didn't catch this?

Same reason `import Data.List (head)` doesn't generate a warning despite `head` already being imported from Prelude, I think. (Not that I know what that reason is precisely.)

slotThe:
> Same reason `import Data.List (head)` doesn't generate a warning despite `head` already being imported from Prelude, I think. (Not that I know what that reason is precisely.)

But it does, right?

``` console
$ cat test.hs
module Main where

import Data.List (head)

main :: IO ()
main = print $ head ['a'..]

$ ghc -Wall test.hs
Loaded package environment from /home/slot/.ghc/x86_64-linux-8.10.7/environments/default
[1 of 1] Compiling Main             ( test.hs, test.o )

test.hs:3:1: warning: [-Wunused-imports]
    The import of ‘Data.List’ is redundant
      except perhaps to import instances from ‘Data.List’
    To import instances alone, use: import Data.List()
  |
3 | import Data.List (head)
  | ^^^^^^^^^^^^^^^^^^^^^^^
Linking test ...
```

slotThe:
> GHC 8.4 does not, and it's the only one for which we enable `-Wunused-imports`.

Oh; interesting that something (seemingly) so fundamental was reworked that recently

PR #664, opened 2021-12-21
XMonad.Hooks.Statusbar: Add parentheses to example

chezbgone:
### Description

Add (required) missing parentheses to example xmobar config code in the multiple status bars section of XMonad.Hooks.Statusbar.

### Checklist

        manually, ...) and concluded: no change needed(?)

### DIFF
--- a/XMonad/Hooks/StatusBar.hs
+++ b/XMonad/Hooks/StatusBar.hs
@@ -377,7 +377,7 @@ statusBarPipe cmd xpp = do
 --
 -- > xmobarrc_top
 -- > Config { ...
--- >        , commands = [ Run XPropertyLog "_XMONAD_LOG_1", ... ]
+-- >        , commands = [ Run (XPropertyLog "_XMONAD_LOG_1"), ... ]
 -- >        , template = "%_XMONAD_LOG_1% }{ ..."
 -- >        }
 --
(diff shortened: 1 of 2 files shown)


XMonad/Hooks/StatusBar.hs
```diff
@@ -377,7 +377,7 @@ statusBarPipe cmd xpp = do
 --
 -- > xmobarrc_top
 -- > Config { ...
--- >        , commands = [ Run XPropertyLog "_XMONAD_LOG_1", ... ]
+-- >        , commands = [ Run (XPropertyLog "_XMONAD_LOG_1"), ... ]
```
slotThe, line 380: I don't think that this is needed when one just uses a config file for xmobar (e.g., see [here](https://github.com/liskin/dotfiles/blob/home/.xmobarrc#L17)), which is what the documentation is concerned with

TheMC47: I think so too, thanks for the contribution @chezbgone but I'd keep it as is

chezbgone: I see, I actually use a "pure Haskell config" (described in 9.3 of https://xmobar.org/#configurations-written-in-pure-haskell), which gave me a type error for the example given, so I was a bit confused.

I agree that we should keep the original as slotThe noted, since the documentation should be for standard configurations, and also the type error was quite easy to resolve.

Issue #669, opened 2021-12-27
GitHub Actions: Add separate hlint action


slotThe:
### Problem Description

Due to hlint's [inconsistent packaging of releases](https://github.com/ndmitchell/hlint/issues/1326) it may be better to introduce a separate action for it.  This would enable us to

  + have more fine-grained control over the version that we use (e.g. in a situation like this we could just roll back to the last release),
  + decouple the "hlint" step from the "does it work with some recent versions of GHC" step, which seems conceptually nicer.

For example, the `haskell-language-server` people seem to have already done this [here](https://github.com/haskell/haskell-language-server/pull/2537)

ndmitchell:
The latest release of hlint now has a release binary.

slotThe:
> The latest release of hlint now has a release binary.

That's great news; thanks for the heads up!

liskin:
I'll just note that the actions used in HLS CI are now part of the "official" Haskell GH Actions: https://github.com/haskell/actions/tree/main/hlint-setup, https://github.com/haskell/actions/tree/main/hlint-run so if this ever breaks again, we should try those.

liskin:
Note that even with haskell-ci, we do have fine-grained control over the version we use, see https://github.com/xmonad/xmonad-contrib/commit/154388aa208fdb2e45da8afa6f72c72deb0251b9#diff-eacabc2a664e928ef8b473245e842e4b29b9904fb1e4f64d565abbab93323e5cR12
So this really is only about the decoupling. Not sure if worth it, as we've been back to "just works" for a while.

slotThe:
I didn't know about the `hlint-version` options, you're right that this is probably enough.  Feel free to close if you don't think the decoupling is worth it

liskin:
I just realized it's not just decoupling, but also that the https://github.com/haskell/actions/tree/main/hlint-run action produces annotations that GitHub understands and shows them directly in the PR diff. So this will likely be a quality-of-life improvement after all, and I'll leave this open then.

(I'll deal with https://github.com/xmonad/xmonad/pull/401 and https://github.com/xmonad/xmonad/pull/402 today but won't be spending any more time on hlint this weekend…)

PR #671, opened 2021-12-28
X.L.BinarySpacePartition: Ignore minimized windows

SimSaladin:
### Description

Hidden windows are now ignored by the layout so that hidden windows in the stack don't offset position calculations in the layout.

Also in X.H.ManageHelpers: added `isMinimized`

### Checklist

        manually, ...) and concluded: I use the BSP layout with hidden windows present a lot and this patch is essential to me. At least one other person on the mailing list is also happy with it: https://mail.haskell.org/pipermail/xmonad/2021-May/015432.html

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,10 +59,17 @@
     - Flipped how `(^?)`, `(~?)`, and `($?)` work to more accurately
       reflect how one uses these operators.
 
+    - Added `isMinimized`
+
   * `XMonad.Actions.WindowNavigation`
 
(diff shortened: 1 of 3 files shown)


slotThe (approved):
I don't personally use the BSP layout but this definitely seems like a desirable change; thanks!

EDIT: Don't mind the hlint CI failure, this is related to https://github.com/xmonad/xmonad-contrib/issues/669, which we haven't addressed yet.

slotThe:
Thanks!

PR #675, opened 2022-01-02
keysMoveWindow: change signature type D to ChangeDim in XMonad.Actions.FloatKeys

afreakk:
### Description

keysMoveWindow takes a (Dimension, Dimension) which is a word32, which doesnt take negative numbers.
At the moment it works to just put a negative number in there, but the compiler shows warnings.
So this change should get rid of compiler warnings when doing `keysMoveWindow (0,-20)` for example.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -90,6 +90,11 @@
       `xmonad-contrib` to be compiled with `X11-xft` version 0.3.4 or
       higher.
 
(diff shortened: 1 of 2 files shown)


TheMC47 (approved):
Thank you 🎉 can you please squash the commits?

PR #676, opened 2022-01-03
X.A.FloatKeys: Export Type ChangeDim

exorcist365:
### Description

behind them.

### Checklist

        manually, ...) and concluded: that the type is not currently exported

### DIFF
--- a/XMonad/Actions/FloatKeys.hs
+++ b/XMonad/Actions/FloatKeys.hs
@@ -19,7 +19,7 @@ module XMonad.Actions.FloatKeys (
                 keysMoveWindowTo,
                 keysResizeWindow,
                 keysAbsResizeWindow,
-                P, G,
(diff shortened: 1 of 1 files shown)


slotThe (commented):
Since it's just a type alias and people aren't really going to write their own function with it (at least AFAICT the functions in X.A.FloatKeys do not lend themselves to a lot of custom modifications) is there any reason we should export this?

exorcist365:
Well, so are P and G and they are exported, if it's exported it's going to appear on the hackage page so people know what it represents without looking through the source code.

slotThe:
> Well, so are P and G and they are exported, if it's exported it's going to appear on the hackage page so people know what it represents without looking through the source code.

Oh the Haddocks do indeed look quite horrible with this not exported—thanks!

Issue #677, opened 2022-01-06
Deprecate X.C.Prime


geekosaur:
### Problem Description

`XMonad.Config.Prime` doesn't work with most extensions and has confused at least one user who stumbled across it. Let's deprecate it now and make it go away in a few versions, since it's unlikely to ever be updated.

### Steps to Reproduce

Try to build a config around `XMonad.Config.Prime`, using extensions beyond anything directly supported by it. `X.U.NamedScratchpad` and `X.U.EZConfig` specifically came up.

### Configuration File

Please include the smallest _full_ configuration file that reproduces
the problem you are experiencing:

```haskell
module Main (main) where

import XMonad
import XMonad.Config.Prime

main :: IO ()
main = xmonad $ do
  nothing
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17.0
    - [x] With `xmonad-contrib` version 0.17.0

TheMC47:
:+1:  Seems like a cool proof of concept, but it also requires a nontrivial effort to keep up-to-date.

geekosaur:
And `X.C.Monad` is even worse.

slotThe:
:+1: from me as well.  We could start the deprecation process (along with starting to take care of #595) once https://github.com/xmonad/xmonad-web/pull/49 is merged

liskin:
No strong opinion from me. Last time I looked, X.C.Prime seemed somewhat nice, but I've never heard about anyone using it, and if it's actually impossible or impractical to use it, we might as well deprecate it. After all, we've been moving towards extensions having an `XConfig a → XConfig a` API anyway, which probably makes X.C.Prime unnecessary, right?

geekosaur:
Yes. But my main worries are (a) do we even know if this and `X.C.Monad` have been updated with new fields or defaults added to `XConfig`? (b) you *can* use this with most extensions but it's noticeably more painful.

geekosaur:
Also `X.C.Monad` says in its own description "probably broken".

Issue #595, opened 2021-08-15
Consider removing user specific modules from contrib


IvanMalison:
It seems really strange to me that user specific configuration is included in xmonad-contrib. My suspicion is that very few people use these prebuilt configs, and it seems kind of ridiculous that everyone is forced to compile/download modules associated with this code.

It seems like a good solution might be to take these user specific configurations and put them in a separate repository that people could use if they want to try these configuration out.

slotThe:
I would imagine these are from the time when "real world" xmonad configs
were not all that available online and it was nice to have a reference.

On Sat, Aug 14 2021 23:26, Ivan Malison wrote:
> It seems like a good solution might be to take these user specific
> configurations and put them in a separate repository that people could
> use if they want to try these configuration out.

We could do this, but I wonder if the trouble of deprecating them,
waiting 2--3 releases (however long that's going to take) and finally
moving them to a new repo is really worth it in this case.

TheMC47:
> It seems like a good solution might be to take these user specific configurations and put them in a separate repository that people could use if they want to try these configuration out.

The website seems like a good candidate for that, we could also maybe add some screenshots

ssbothwell:
I'm pretty new around here but I agree that it doesn't make much sense to have these configs in the repo. I agree with @TheMC47  that the website is a great place to put this type of content.

IvanMalison:
> We could do this, but I wonder if the trouble of deprecating them, waiting 2--3 releases (however long that's going to take) and finally moving them to a new repo is really worth it in this case.

Not clear that this sort of thing ACTUALLY needs to be done.

The new release that is coming is ALL KINDS of backwards incompatible with the last one. I don't see any reason why would need to ease in the removal of these modules. I HIGHLY doubt that many people use them anyway.

slotThe:
> The new release that is coming is ALL KINDS of backwards incompatible with the last one. I don't see any reason why would need to ease in the removal of these modules.

It is true that there are many breaking changes, but most are more of the form "X has been deprecated, use Y instead" or other small changes to the API.  I would classify removing an entire module as a bit of a different thing (though maybe I'm being too careful here).  

> I HIGHLY doubt that many people use them anyway.

I have actually seen the occasional user module importing one of these to use a utility function or two.  It's true that these are probably the among "safest" modules we could remove in terms of users noticing, but it _is_ removing entire modules.  

[...]

PR #679, opened 2022-01-12
Deprecate user and abandoned modules

slotThe:
### Description

NOTE: I'm bad at writing things like this, so feel free to suggest improvements.

Now that the user configs are on the website, it is time to deprecate them.  At the same time deprecated X.C.Monad, which hasn't been updated since 2008 and X.C.Prime, which has confused users quite recently, thinking it to be a better starting place.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/677
Fixes: https://github.com/xmonad/xmonad-contrib/issues/595
Related: https://github.com/xmonad/xmonad-web/pull/49

### Checklist

        manually, ...) and concluded: Deprecation warnings work as expected.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,13 @@
       now return a `Parser` (from `XMonad.Util.Parser`) instead of a
       `ReadP`.
 
+  * `XMonad.Config.{Arossato,Bluetile,Dmwit,Droundy,Monad,Prime,Saegesser,Sjanssen}`
+
+    - Deprecated all of these modules.  The user-specific configuration
+      modules may still be found [on the website].
+
+[on the website]: https://xmonad.org/configurations.html
+
 ### New Modules
 
   * `XMonad.Layout.CenteredIfSingle`
--- a/XMonad/Config/Arossato.hs
+++ b/XMonad/Config/Arossato.hs
@@ -16,6 +16,7 @@
 ------------------------------------------------------------------------
 
 module XMonad.Config.Arossato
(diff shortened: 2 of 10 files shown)


psibi (approved):

geekosaur (approved):
LGTM!

CHANGES.md
```diff
@@ -10,6 +10,13 @@
       now return a `Parser` (from `XMonad.Util.Parser`) instead of a
       `ReadP`.
 
+  * `XMonad.Config.{Arossato,Bluetile,Dmwit,Droundy,Monad,Prime,Saegesser,Sjanssen}`
+
+    - Deprecated all of these modules.  The user-specific configuration
+      modules will continue to live on [on the website].
```
liskin, line 16: double "on"

liskin (approved):
Minor nit, as usual. Rest LGTM.

CHANGES.md
geekosaur: I'd say that is actually correct: "live on" and "on the website" are separate clauses. Might find a way to rephrase though.

liskin: Oh indeed, nevermind then. Sorry :-)

slotThe: Yeah this is perhaps a bit awkwardly phrased...  0e28874eb60715903d27a794ad36dbab00547253 rephrases it a tad

slotThe:
> Oh, one more thing: how do we know when to remove it? Will we just git blame it in a year or two? (Fine with that probably. :-))

That was the original plan yes (I'm sure we will remember!)

Issue #472, opened 2021-03-04
X.A.Search promptSearchBrowser not showing


davama:
### Problem Description

Dont see prompt when type keybinding `mod-s` to do `prompSearchBrowser`
Keybinding `mod+shift-s` which uses clipboard `selectSearchBrowser`, works as intended.

Used to work before, but about a 1 month ago I updated to the latest master branch and came accross some issues, which were kindly address [here](https://www.reddit.com/r/xmonad/comments/la5n75/couldnt_match_expected_type_io_t0_with_actual/). Honestly never tested this `mod-s` keybinding until today.....

Hopefully provided all relavent info from my `xmonad.hs` file

`xmonad --recompile` produces no error. But is `XMonad recompilation process exited with success!`

### Configuration File

```haskell
import qualified XMonad.Actions.Search as S


myInternet      = "google-chrome-stable"


myXPConfig :: P.XPConfig
myXPConfig = P.def
 { P.font        = myFont 8
 , P.bgColor     = (colorLook Hue 1)
 , P.fgColor     = (colorLook White 1)
 , P.borderColor = (colorLook Red 0)
 , P.promptBorderWidth = 1
 , P.position    = P.Bottom
 , P.height      = 15
 --, P.autoComplete = Just 500000
 , P.historySize = 10 }


searchEngineMap method = M.fromList $
 [ 
 ((0, xK_a), method S.amazon)
 , ((0, xK_d), method S.dictionary)
 , ((0, xK_e), method S.ebay)
 , ((0, xK_g), method S.google)
 , ((0, xK_i), method S.images)
 , ((0, xK_j), method $ S.searchEngine "Jira" "https://blabla/browse/")
 , ((0, xK_m), method S.maps)
 , ((0, xK_s), method $ S.searchEngine "Salesforce" "https://blabla/_ui/search/ui/UnifiedSearchResults?str=")
 , ((0, xK_w), method S.wikipedia)
 , ((0, xK_y), method S.youtube)
 ]

...
...
 , ((mod4Mask .|. shiftMask, xK_s), SM.submap $ searchEngineMap $ S.selectSearchBrowser myInternet) -- clipboard web browser search
 , ((mod4Mask, xK_s), SM.submap $ searchEngineMap $ S.promptSearchBrowser myXPConfig myInternet ) -- prompt search

```

Please let me know if you need more info from me.

Thank you,
Dave

slotThe:
[...]

PR #680, opened 2022-01-15
X.A.Submap: Add `visualSubmap`

slotThe:
### Description

tl;dr: [screenshot] of the new behaviour.

I'm not quite happy with this, but the reason I'm posting it anyways is that I'm interested in feedback.  We could make this use X.U.NamedActions, which may be conceptually nicer and could make the specification of descriptions a bit cleaner.  This would, however, introduce a regression with regards to `submapNamed`—it could not be in X.U.NamedActions anymore, as we would need to import that module from X.A.Submap.  It may also not be clear which submap implementation is chosen when one specifies named actions via EZConfig (which is something that I haven't touched on at all here).

##### X.Prelude: Add keymaskToString, keyToString

This technically introduces a regression with regards to the way that modifier masks are printed in X.U.NamedActions and X.H.DebugEvents. However, since this way of printing masks is move in line with X.U.EZConfig, I personally don't think that this is noteworthy.

##### X.U.XUtils: Add framework for manipulating simple windows

Adds several new functions and type for manipulating "simple windows".

There are several ways to draw windows in X.U.XUtils and other places already, but they are all quite manual.  Most of the time one does not want to think about dimensions much and assumes that the code is smart enough to "figure it out"; this is an attempt to do exactly that.

There is a less-managed version `showSimpleWindow`, which just creates and shows the windows, as well as a wrapper-like `withSimpleWindow` that also destroys the window once some action finishes executing.

With these functions it should be possible to refactor some contrib modules that currently draw windows manually, like X.U.EasyMotion.

##### X.A.Submap: Add visualSubmap

Add a `visualSubmap` function, which works much like the regular `submap` one, except that it visualises the available choices in a pop-up window.

Related: https://github.com/xmonad/xmonad-contrib/issues/472

[Screenshot]: https://i.imgur.com/mqRGWVw.png

### Checklist

        manually, ...) and concluded: works manually...  Not sure there is much chance to test the window stuff, though with `keyToString` we may now have a chance to property test EZConfig!  I've not done this here, though.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -116,6 +116,11 @@
       resetting borders back in layouts where you want borders after calling
       `voidBorders`.
 
+  * `XMonad.Prelude`
+
+    - Added `keymaskToString` and `keyToString` to show a key mask and a
+      key in the style of `XMonad.Util.EZConfig`.
+
 ## 0.17.0 (October 27, 2021)
 
 ### Breaking Changes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -121,6 +121,12 @@
     - Added `keymaskToString` and `keyToString` to show a key mask and a
       key in the style of `XMonad.Util.EZConfig`.
 
+  * `XMonad.Util.XUtils`
+
+    - Added `withSimpleWindow`, `showSimpleWindow`, `WindowConfig`, and
+      `WindowRect` in order to simplify the handling of simply popup
+      windows.
+
 ## 0.17.0 (October 27, 2021)
 
 ### Breaking Changes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -124,8 +124,13 @@
   * `XMonad.Util.XUtils`
 
     - Added `withSimpleWindow`, `showSimpleWindow`, `WindowConfig`, and
-      `WindowRect` in order to simplify the handling of simply popup
+      `WindowRect` in order to simplify the handling of simple popup
       windows.
+
+  * `XMonad.Actions.Submap`
+
+    - Added `visualSubmap` to visualise the available keys and their
+      actions when inside a submap.
 
 ## 0.17.0 (October 27, 2021)
 
--- a/XMonad/Actions/Submap.hs
+++ b/XMonad/Actions/Submap.hs
@@ -17,13 +17,18 @@ module XMonad.Actions.Submap (
                              -- * Usage
                              -- $usage
                              submap,
+                             visualSubmap,
                              submapDefault,
-                             submapDefaultWithKey
+                             submapDefaultWithKey,
+
+                             -- * Utilities
+                             subName,
                             ) where
 import Data.Bits
-import XMonad.Prelude (fix, fromMaybe)
-import XMonad hiding (keys)
 import qualified Data.Map as M
+import XMonad hiding (keys)
+import XMonad.Prelude (fix, fromMaybe, keyToString)
+import XMonad.Util.XUtils
 
 {- $usage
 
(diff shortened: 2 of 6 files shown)


slotThe:
After some more experiments with NamedActions, I've concluded that it's not worth it.  This is pretty much as good as I'm able to make it; I will merge in a bit if no one has any other objections

liskin (approved):
Well… https://twitter.com/iamdevloper/status/397664295875805184 :-)

(I have limited capacity these days and this doesn't seem to have potential to break important stuff, so I haven't read it line by line, sorry.)

XMonad/Util/XUtils.hs
```diff
@@ -157,6 +166,86 @@ paintTextAndIcons w fs wh ht bw bc borc ffc fbc als strs i_als icons = do
...
+ where
+  getRectangle :: WindowRect -> X Rectangle
+  getRectangle = \case
+    CenterWindow -> gets $ screenRect . W.screenDetail . W.current . windowset
+    CustomRect r -> pure r
+
+-- | Like 'showSimpleWindow', but fully manage the window; i.e., destroy
+-- it after the given function finishes its execution.
+withSimpleWindow :: WindowConfig -> [String] -> X a -> X a
+withSimpleWindow wc strs doStuff = do
```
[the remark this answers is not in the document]
geekosaur: Don't most `with…` functions look like that, when it comes down to it?

liskin:
> After some more experiments with NamedActions, I've concluded that it's not worth it. This is pretty much as good as I'm able to make it; I will merge in a bit if no one has any other objections

Have you considered simply putting `visualSubmap` into `NamedActions` or perhaps into a separate module? (Probably have, just checking. I don't really know if NamedActions are good/popular enough to try very hard to use it.)

XMonad/Actions/Submap.hs
```diff
@@ -71,28 +117,38 @@ submapDefault = submapDefaultWithKey . const
...
+                    else return (m, keysym)
+              _ -> return (0, 0)
+        -- Remove num lock mask and Xkb group state bits
+        m' <- cleanMask $ m .&. ((1 `shiftL` 12) - 1)
+        io $ do ungrabPointer dpy currentTime
+                ungrabKeyboard dpy currentTime
+                sync dpy False
+        pure (m', s)
+
+    lookupAction (m', s)
```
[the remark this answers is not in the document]
slotThe: We could totally do that (I didn't put too much thought into how ghastly the type signature of `submapWorker` looks because it's not exported anyways).  The new ergonomics also don't seem too bad: 

``` haskell
    withSimpleWindow wc descriptions submapWorker >>= \(m', s) ->
        maybe (pure ()) snd (M.lookup (m', s) keys)
```

XMonad/Util/XUtils.hs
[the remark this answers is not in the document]
slotThe: Is there a `bracket` for `X`?  I couldn't find it[^1]; though I guess we could wrap this in a `userCode` or something in either case.

[^1]: Too bad we're a state on top of a reader instead of just a reader with some IORefs or TVars for the mutable bits; we could then just depend on `unliftio` and wouldn't have to think about things like this ever again :/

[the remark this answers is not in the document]
liskin: Oh crap, indeed! Nevermind then.

slotThe:
> Have you considered simply putting `visualSubmap` into `NamedActions` or perhaps into a separate module? (Probably have, just checking. I don't really know if NamedActions are good/popular enough to try very hard to use it.)

I have, yeah, but both options feel very wrong to me as this is decidedly a submap thing (NamedActions already weirdly has `submapName` and I really don't want to further muddle the waters)

It's not a huge boost in this case, I doubt we'd gain very much from using NamedActions (just that it would be conceptually nicer)

XMonad/Util/XUtils.hs
```diff
@@ -241,8 +241,8 @@ showSimpleWindow WindowConfig{..} strs = do
 
 -- | Like 'showSimpleWindow', but fully manage the window; i.e., destroy
 -- it after the given function finishes its execution.
-withSimpleWindow :: WindowConfig -> [String] -> X a -> X a
-withSimpleWindow wc strs doStuff = do
+withSimpleWindow :: WindowConfig -> [String] -> X a -> X (Maybe a)
+withSimpleWindow wc strs doStuff = userCode $ do
```
liskin, line 245: Not sure if this is useful at all. A bracket would ensure the window is destroyed when `doStuff` throws, but that won't happen here, so it probably does nothing useful? I'd either move the `userCode` to just wrap `doStuff` (and then be a bit uncomfortable about the fact that there's still other stuff between create/destroy like showWindow/paintWindow which isn't surrounded by catch, but code like this worked before so maybe not a problem really) or just drop it. :-)

XMonad/Actions/Submap.hs
```diff
@@ -117,38 +118,33 @@ submapDefault = submapDefaultWithKey . const
...
 
 -----------------------------------------------------------------------
 -- Internal stuff
 
-submapWorker :: (X (KeyMask, KeySym) -> X (KeyMask, KeySym)) -- ^ Get keys
-             -> ((KeyMask, KeySym) -> X ())                  -- ^ Lookup and execute action, if any
-             -> X ()
-submapWorker getKeys lookupAction = do
+submapWorker :: X (KeyMask, KeySym)
+submapWorker = do
```
liskin, line 128: Now it's really just a `getKey` (or `waitForKeyPress` or whatever).

XMonad/Util/XUtils.hs
geekosaur: I think you can only reach code like this via a handler or hook that already goes through `userCode`.

liskin: > I think you can only reach code like this via a handler or hook that already goes through `userCode`.

Indeed, but you may still want to minimize the damage if you expect something to actually fail in a real-world usage. (Which I'm not really sure can happen here, hence my suggestion to just drop it and forget it, until perhaps we get a report of it happening.)

slotThe: Mh I guess you're right; I'll remove it and if this blows up we'll have an excuse to release 0.17.2!

CHANGES.md
```diff
@@ -103,6 +103,22 @@
...
 
+  * `XMonad.Prelude`
+
+    - Added `keymaskToString` and `keyToString` to show a key mask and a
+      key in the style of `XMonad.Util.EZConfig`.
+
+  * `XMonad.Util.XUtils`
+
+    - Added `withSimpleWindow`, `showSimpleWindow`, `WindowConfig`, and
+      `WindowRect` in order to simplify the handling of simply popup
```
liskin, line 114:
```suggestion
      `WindowRect` in order to simplify the handling of simple popup
```

liskin (approved):

PR #685, opened 2022-02-05
X.U.EZConfig: Add property tests

slotThe:
### Description

Property tests for the EZConfig parser.  Plus, some improvements for the key pretty-printer as a result of it.

##### X.Prelude: Concat modifier keys instead of unwording

Instead of printing "M- S- C-", print "M-S-C".

##### X.Prelude: Improve keyToString output

The keysymToString function prints the key names in pure ASCII; e.g., `keysymToString 250` returns "uacute" instead of "ú".  This is undesirable when printing these keysyms in places like visualSubmap.

Thus, move all of the key infrastructure (heh) from X.U.EZConfig to X.Prelude and look up the name of the key if possible.  For better composability, slightly change the signature for `regularKeys` and the associated parser.

##### tests: Add property tests for X.U.EZConfig

Adds property tests in both directions for the parser in X.U.EZConfig. As with the tests for X.P.OrgMode, these (modulo `Maybe` noise) take the shape of the operation of an inverse semigroup:

    pp ∘ p ∘ pp ≡ p        and        p ∘ pp ∘ p = p,

where pp is pretty-printing and p is parsing.

### Checklist

        manually, ...) and concluded: that I'm adding tests!


geekosaur:
Do we know that nobody's going to want the proper X11 name of a key (see for example Compose definition files, or `xdotool`)? Or do we provide and document an alternative that allows access to the actual name?

slotThe:
> Do we know that nobody's going to want the proper X11 name of a key (see for example Compose definition files, or `xdotool`)? Or do we provide and document an alternative that allows access to the actual name?

Well, `keysymToString` is still exported by [Graphics.X11.Xlib.Misc](https://hackage.haskell.org/package/X11/docs/Graphics-X11-Xlib-Misc.html#v:keysymToString) so that should be fine, I suppose?

slotThe:
I think the existence of `keysymToString` is good enough (and we don't export our keysym->string function anyways but only the whole `keyToString`), so I will merge this in a few if there are no other objections

tests/EZConfig.hs
```diff
@@ -30,15 +44,59 @@ spec = do
...
+-- Newtypes and Arbitrary instances
+
+newtype AKeyMask = AKeyMask KeyMask
+  deriving newtype (Show)
+
+instance Arbitrary AKeyMask where
+  arbitrary :: Gen AKeyMask
+  arbitrary = fmap (coerce @Word32 . sum . nub) . listOf . elements $
+    -- 2^1 = 2 is lock; EZConfig doesn't parse this.
+    0 : [2^i | i <- [0, 2, 3, 4, 5, 6, 7] :: [Int]]
```
liskin, line 80: Might as well just use `[shiftMask, controlMask, mod1Mask, mod2Mask, mod3Mask, mod4Mask, mod5Mask]` … :-)

tests/EZConfig.hs
```diff
@@ -30,15 +44,59 @@ spec = do
...
+-- | Parsing preserves all info that printing does.
+prop_encodePreservation :: KeyString -> Property
+prop_encodePreservation (coerce -> s) = parse s === (parse . pp =<< parse s)
+ where parse = runParser (parseKeySequence def)
+       pp    = unwords . map keyToString
+
+-- | Printing preserves all info that parsing does.
+prop_decodePreservation :: NonEmptyList (AKeyMask, AKeySym) -> Property
+prop_decodePreservation (getNonEmpty >>> coerce -> xs) =
+  Just (pp xs) === (fmap pp . parse $ pp xs)
```
liskin, line 56: The commit message says `pp ∘ p ∘ pp ≡ p`, should be `… ≡ pp`, right?

tests/EZConfig.hs
```diff
@@ -30,15 +44,59 @@ spec = do
...
+       pp    = unwords . map keyToString
 
 -- | QuickCheck can handle the 8! combinations just fine.
 modifiers :: [String]
-modifiers = map concat $
-  permutations ["M-", "C-", "S-", "M1-", "M2-", "M3-", "M4-", "M5-"]
+modifiers = map concat $ permutations mods
+
+mods :: [String]
+mods = ["M-", "C-", "S-", "M1-", "M2-", "M3-", "M4-", "M5-"]
```
liskin, line 65: Neither of the tested props detects that `M-` isn't preserved in `pp ∘ p`. Is there a reason why you don't test `pp ∘ p ≡ id` and `p ∘ pp ≡ id` instead?

slotThe: For one, the pretty-printer, being a pure function, does not know about the users mod key and so prints `M-` as whatever mod key it actually is (`M1-`, `M2-`, ...).

The more "fundamental" reason is that we allow permutations of the modifiers, but the pretty-printer arranges things deterministically.  So for example `p ∘ pp ∘ p "S-M-a" ≡ p "S-M-a"` but `pp ∘ p "S-M-a" = "M-S-a" ≠ "S-M-a"`.

tests/EZConfig.hs
slotThe: Oh, right, that's much clearer :)

tests/EZConfig.hs
liskin: Oh, I missed that the list of mods was already used by existing tests, so leaving out `M-` wasn't an option. Now it makes sense that `M-` is here even though for the new tests it's not useful.

liskin (approved):

Issue #290, opened 2019-01-23
XMonad.Prompt.cleanMask doesn't filter out enough modifiers


mgsloan:
For keypresses other than typing, XMonad.Prompt looks up actions in a `M.Map (KeyMask,KeySym) (XP ())`.  Some aspects of `KeyMask` shouldn't be relevant to keyboard entry, and so it is preprocessed with `cleanMask` , which has the following definition:

```haskell
cleanMask :: KeyMask -> XP KeyMask
cleanMask msk = do	
  numlock <- gets numlockMask	
  let highMasks = 1 `shiftL` 12 - 1	
  return (complement (numlock .|. lockMask) .&. msk .&. highMasks)
```

In essence, what this does is allows all modifiers except for numlock and capslock.  The problem is there are a lot of other modifiers that may not be relevant to the prompt.  For example, if I hold down the left mouse button and press enter, instead of the prompt confirming, a box like □ gets inserted as if I typed it.  Holding mod4 and pressing enter does the same.

I have seen a case in the past where typing and pressing special keys types absolute gibberish, I don't recall whether it was unicode or extended ascii, but I suspect it was related to a stuck modifier, and prevented exiting the prompt.  Unfortunately, I have no idea what triggered this behavior.

I have [a solution](https://github.com/mgsloan/xmonad-contrib/commit/de46092b2ef206aa34fd279cab21ef2be293a6d8) on my patched version of xmonad-contrib, but I'm not sure if it'd be good to merge this.  Instead I do the following:

```
cleanMask msk = return (msk .&. (controlMask .|. mod1Mask))
```

I happen to know that my config only cares about `controlMask` and `mod1Mask`, and so this ignores all the other masks and so doesn't succumb to the sticky mask problem.

I propose the following resolution to this issue:

1. Have `cleanMask` use a whitelist, probably something like `[controlMask, shiftMask, mod1Mask, mod2Mask, mod3Mask, mod4Mask]`.  So, not quite as restrictive as my list.

2. Make the function used for cleaning the keymask be part of the XPrompt configuration

What do y'all think?

geekosaur:
One problem is those high masks represent keyboard layouts, and this will matter to some people.

mgsloan:
[...]

PR #686, opened 2022-02-10
X.{A.{{Grid,Tree}Select,Submap},Prompt}: KeyPress handling fixes

liskin:
### Description

This is a cleaned up version of https://github.com/xmonad/xmonad-contrib/pull/590 which additionally addresses https://github.com/xmonad/xmonad-contrib/issues/290 (I believe nobody really depends on the XKB and mouse state not being stripped so it's okay to change behaviour) and https://github.com/xmonad/xmonad/issues/172 (this needs https://github.com/xmonad/xmonad/pull/374).

Original description from https://github.com/xmonad/xmonad-contrib/pull/590 follows:

> This changes `makeXEventhandler` to behave much closer to how xmonad itself handles keypresses. The primary difference lies in that xmonad reads raw keycode and then converts it to unmodified keysym, while `makeXEventhandler` was reading actual keysyms. As a consequence, key definitions like `(shiftMap, xK_Tab)` didn't work with `makeXEventhandler` on many layouts because an actual keysym for 'Shift+Tab' is commonly `ISO_LEFT_TAB`, and not `Tab`.
>
> Additionally, the mask is stripped of its high bits, because apparently X likes to encode current layout group information in bits 13 and up, and we don't really want our control keys to depend on the current layout group. Please let me know if there is a better way to handle this.

One thing to consider: if we are concerned that unconditionally stripping XKB and mouse state might break someone's workflow, we'll need to make this configurable, and then we'll need to move `cleanKeyMask` out of X.Prelude as we can't import X.U.ExtensibleConf.

### Checklist

  - [X] I've considered how to best test these changes (property, unit, manually, ...) and concluded: I manually tested TreeSelect and Prompt and haven't found any breakage.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -132,6 +132,13 @@
     - Added `visualSubmap` to visualise the available keys and their
       actions when inside a submap.
 
+  * `XMonad.Prompt`, `XMonad.Actions.TreeSelect`, `XMonad.Actions.GridSelect`
+
+    - Key bindings now behave similarly to xmonad core:
+      State of mouse buttons and XKB layout groups is ignored.
+      Translation of key codes to symbols ignores modifiers, so `Shift-Tab` is
+      now just `(shiftMap, xK_Tab)` instead of `(shiftMap, xK_ISO_Left_Tab)`.
+
 ## 0.17.0 (October 27, 2021)
 
 ### Breaking Changes


slotThe (commented):
Compiles and doesn't seem to break the prompt for me and the code also looks reasonable, so LGTM!

> One thing to consider: if we are concerned that unconditionally stripping XKB and mouse state might break someone's workflow, we'll need to make this configurable, and then we'll need to move `cleanKeyMask` out of X.Prelude as we can't import X.U.ExtensibleConf.

Reminds me a bit of [this](https://xkcd.com/1172/)—probably not going to be a problem :)

PR #688, opened 2022-02-18
X.H.ManageDocks: Avoid unnecessary refresh (loop) for decorations

liskin:
### Description

This started as an investigation of the root cause of https://github.com/xmonad/xmonad-contrib/issues/565, which was addressed by https://github.com/xmonad/xmonad-contrib/pull/621 without having a full understanding of why xmonad ends up in a busy loop.

What was going on:

* X.L.SubLayouts group update algorithm assumes no duplicities in the Stack, and results in multiplying those duplicities otherwise. Note that this is a reasonable assumption—duplicities in a Stack have no defined behaviour in xmonad (X11 can't place a window on a screen twice), so "fixing" this algorithm is a waste of time.

* X.L.Decoration creates windows which X.H.ManageDocks treats as possible docks, resulting in extra refresh whenever they appear.

* The extra refresh causes X.L.SubLayouts to multiply the duplicities, X.L.Decoration to create new decoration windows, and these invalidate the X.H.ManageDocks strut cache, leading to yet another refresh, and an endless loop of these.

Having concluded that the no-duplicities assumption is a reasonable one, there's nothing really to fix here after https://github.com/xmonad/xmonad-contrib/pull/621. Still, the extra refresh might be causing extra flicker in decorated layouts, and we can easily avoid it, so I do that here. Plus a bit of documentation for X.L.SubLayouts.

### Checklist

  - [X] I've considered how to best test these changes (property, unit, manually, ...) and concluded: I manually checked that the X.H.ManageDocks tweak breaks the endless loop. I don't think it should break anything else: xmonad shouldn't be creating any docks/windows that need X.H.ManageDocks to be avoided.

### DIFF
--- a/XMonad/Hooks/ManageDocks.hs
+++ b/XMonad/Hooks/ManageDocks.hs
@@ -148,15 +148,20 @@ requestDockEvents w = whenX (not <$> isClient w) $ withDisplay $ \dpy ->
     withWindowAttributes dpy w $ \attrs -> io $ selectInput dpy w $
         wa_your_event_mask attrs .|. propertyChangeMask .|. structureNotifyMask
 
--- | Checks if a window is a DOCK or DESKTOP window
+-- | Checks if a window is a DOCK or DESKTOP window.
+-- Ignores xmonad's own windows (usually _NET_WM_WINDOW_TYPE_DESKTOP) to avoid
+-- unnecessary refreshes.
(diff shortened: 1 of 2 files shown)


slotThe (approved):
This is great, thank you for the thorough investigation!

liskin:
Thx for catching the typos :-)

PR #662, opened 2021-12-13
Add SPStorage to exports of XMonad.Util.DynamicScratchpads

kentookura:
### Description

Allow the user of DynamicScratchpads to access the current DynamicScratchpads state.

### Checklist

  - [ x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ x] I've considered how to best test these changes (property, unit,
        manually, ...) and concluded: XXX

  - [ ] I updated the `CHANGES.md` file

Issue #591, opened 2021-08-13
3 Different Scratchpad Implementations; 2 Scratchpad PRs


IvanMalison:
IMO, Scratchpads are in a sorry state in xmonad-contrib. For some reason, there are 3 different scratchpad implementations in the library that all use completely different code. There are also 2 outstanding PRs that attempt to extend the functionality of scratchpads that seem to have stagnated and probably will not get merged:

https://github.com/xmonad/xmonad-contrib/pull/169
https://github.com/xmonad/xmonad-contrib/pull/454

All of this indicates a few things to me:

- Scratchpads are a pretty widely used feature (There wouldn't be so much activity around them if that were not the case)
- People have pretty widely varying preferences about how scratchpads should behave (This explains the 3 different implementations as well as the PRs)
- None of the existing scratchpad implementations are very extensible or well thought out (no offense to the authors intended)

I think that what we should do is take a look at all of the different scratchpad implementations/prs and put some careful thought into how we might design a single scratchpad implementation that could support all of the behaviors that people seem to want.

The fragmentation that currently exists just makes things confusing for new users, and it makes it difficult for more sophisticated users to make contributions.

slotThe:
It is true that we have different scratchpad implementation, but I don't think I have ever seen someone use anything but X.U.NamedScratchpads (plus the loggers).  I'm sure these people exist, but one implementation clearly "won", more or less.  Indeed, so many people seem to be using this that I think our efforts should rather be directed in the direction of improving that module, rather than trying to write yet another implementation.  Even if something better comes along, there is a lot of inertia in old modules (not least because there are so many old configurations lying about)

IvanMalison:
> Indeed, so many people seem to be using this that I think our efforts should rather be directed in the direction of improving that module, rather than trying to write yet another implementation.

Well I agree in the sense that I think it makes sense not to deprecate the module, but I do think an aggresive refactor is order. I believe that it should (mostly) be possible to preserve the existing interface, but implement it in terms of a new interface that is much more flexible (but perhaps exposed in the same module).

There are some issues with the NamedScratchpads implementation. Some examples are:
- It should be possible to control how the dimissal of scratchpads worked. ExclusiveScratchpads uses minimization rather than banishing to NSP. There is clearly a desire in the userbase to have support for this behavior (as indicated by #169). I don't think it would be hard to
[...]

PR #690, opened 2022-03-02
X.U.NamedScratchpads: Add dynamic scratchpad capabilities

slotThe:
### Description

Supersedes/Closes: https://github.com/xmonad/xmonad-contrib/pull/662

This is a first stab at making X.U.NamedScratchpads a little bit more general (while trying as hard as possible to retain backwards compatibility :/).

##### X.U.NamedScratchpad: Use ExtensibleState to store scratchpads

Instead of using the scratchpads that the user specifies (as this will realistically just be "all of them"), create some extensible state for the scratchpads to reside in.

To ensure that this is done in a backwards compatible way, function signatures did not change and they instead just ignore the list of scratchpads given to them and the initialisation is done in the manageHook, which users already have to use anyways.

##### X.U.NamedScratchpad: Add support for dynamic scratchpads

With the extensible state machinery in place[1], we are now able to seamlessly integrate dynamic scratchpad functionality into X.U.NamedScratchpad.  This is one step into the direction of having a single scratchpad implementation that can "do anything".

Related: https://github.com/xmonad/xmonad-contrib/issues/591
Related: https://github.com/xmonad/xmonad-contrib/pull/662

[1]: 70d08b82f43057447e8e2479aa515b6db0d199e4
     (X.U.NamedScratchpad: Use ExtensibleState to store scratchpads)

### Checklist

        manually, ...) and concluded: I've tested this manually and (to my surprise!) is actually works

### DIFF
--- a/XMonad/Util/NamedScratchpad.hs
+++ b/XMonad/Util/NamedScratchpad.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE InstanceSigs     #-}
-{-# LANGUAGE LambdaCase       #-}
 {-# LANGUAGE TypeApplications #-}
 -----------------------------------------------------------------------------
 -- |
@@ -44,14 +43,16 @@ module XMonad.Util.NamedScratchpad (
   ) where
 
 import Data.Coerce (coerce)
+import Data.Map.Strict (Map, (!?))
 import XMonad
 import XMonad.Actions.DynamicWorkspaces (addHiddenWorkspace)
 import XMonad.Actions.SpawnOn (spawnHere)
 import XMonad.Hooks.ManageHelpers (doRectFloat)
 import XMonad.Hooks.RefocusLast (withRecentsIn)
 import XMonad.Hooks.StatusBar.PP (PP, ppSort)
-import XMonad.Prelude (filterM, find, unless, when)
+import XMonad.Prelude (filterM, unless, when)
 
+import qualified Data.Map.Strict    as Map
 import qualified Data.List.NonEmpty as NE
 
 import qualified XMonad.StackSet             as W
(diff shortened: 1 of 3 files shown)


XMonad/Util/NamedScratchpad.hs
```diff
@@ -111,6 +120,12 @@ data NamedScratchpad = NS { name   :: String      -- ^ Scratchpad name
                           , hook   :: ManageHook  -- ^ Manage hook called for application window, use it to define the placement. See @nonFloating@, @defaultFloating@ and @customFloating@
                           }
 
+newtype NSPState = NSPState [NamedScratchpad]
```
slotThe, line 123: This should be a set, morally, and it could be if we went all-in on the idea of "names determine the scratchpad uniquely" (which, tbh, the current code assumes anyways).  Opinions?

XMonad/Util/NamedScratchpad.hs
```diff
@@ -218,7 +234,11 @@ scratchpadWorkspaceTag = "NSP"
 -- | Manage hook to use with named scratchpads
 namedScratchpadManageHook :: NamedScratchpads -- ^ Named scratchpads configuration
                           -> ManageHook
-namedScratchpadManageHook = composeAll . fmap (\c -> query c --> hook c)
+namedScratchpadManageHook nsps = do
+    ns <- liftX $ XS.get >>= \case
+        NSPState [] -> nsps <$ XS.put (NSPState nsps) -- initialise
+        NSPState xs -> pure xs
```
slotThe, line 240: This is quite a hack, but I couldn't see another way of making this a backwards compatible change :/

XMonad/Util/NamedScratchpad.hs
slotThe: For reference, I've now implemented this as a `Map String NamedScratchpad`.  It's conceptually cleaner, although not really any less code,

slotThe:
I'm semi-happy with this (certainly happier than with a standalone DynamicScratchpads module), so I will merge in a bit if there are no further objections

PR #694, opened 2022-03-19
ci: Update supported GHC versions

slotThe:
NOTE: I will prepare the same thing for the `xmonad` repo, but don't want to create too much noise with PRs.  After this is rubber-stamped I will just push it directly.

### Description

+ Prefer GHC 8.10.7 to 8.10.4, as versions seem to have stabilised now.
+ Add support for Stackage LTS 19, which ships with GHC 9.0.2.
+ Since a new version of 9.2 has been released, prefer 9.2.2 over 9.2.1.

Also, explicitly pass -XCPP to hlint; see [1].

[1]: https://github.com/ndmitchell/hlint/issues/1360

### Checklist

        manually, ...) and concluded: CI passes

### DIFF
--- a/.github/workflows/stack.yml
+++ b/.github/workflows/stack.yml
@@ -24,14 +24,14 @@ jobs:
           - resolver: lts-16
             ghc: 8.8.4
             yaml: stack-master.yaml
-          - resolver: lts-17
-            ghc: 8.10.4
+          - resolver: lts-18
+            ghc: 8.10.7
             yaml: stack-master.yaml
-          - resolver: lts-18
-            ghc: 8.10.7
+          - resolver: lts-19
+            ghc: 9.0.2
             yaml: stack.yaml
-          - resolver: lts-18
-            ghc: 8.10.7
+          - resolver: lts-19
+            ghc: 9.0.2
             yaml: stack-master.yaml
 
     steps:
--- a/.github/workflows/haskell-ci-hackage.patch
+++ b/.github/workflows/haskell-ci-hackage.patch
@@ -38,13 +38,13 @@ set in GitHub repository secrets.
    linux:
      name: Haskell-CI - Linux - ${{ matrix.compiler }}
 @@ -31,6 +38,7 @@
-             compilerVersion: 9.0.1
-             setup-method: hvr-ppa
+             compilerVersion: 9.0.2
+             setup-method: ghcup
              allow-failure: false
 +            upload: true
-           - compiler: ghc-8.10.4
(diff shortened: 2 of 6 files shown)


.github/workflows/stack.yml
```diff
@@ -24,15 +24,18 @@ jobs:
...
             yaml: stack.yaml
           - resolver: lts-18
             ghc: 8.10.7
             yaml: stack-master.yaml
+          - resolver: lts-19
+            ghc: 9.0.2
+            yaml: stack-master.yaml
+          - resolver: lts-19
+            ghc: 9.0.2
+            yaml: stack-master.yaml
```
liskin, line 38: Is this intentionally duplicated?

---

I should've added a comment back then about the stack/stack-master stuff: we primarily test using stack-master (uses X11, xmonad master) but we also want to make sure that it works with the latest X11, xmonad releases so we do that with the most and least recent LTS, just to make the test matrix a bit smaller (and thus less likely to hit github limits).

liskin:
I'm a bit confused about why the hlint issue was not a problem before. Can anyone help me understand that? Otherwise this looks all right apart from the one issue in the stack workflow.

.github/workflows/stack.yml
slotThe: Whoops, not at all!  Forgot to `stack-master.yaml -> stack.yaml` there

slotThe:
> I'm a bit confused about why the hlint issue was not a problem before. Can anyone help me understand that?

It seems like the bug only fires when `-XHaskell2010` is passed to hlint (see https://github.com/ndmitchell/hlint/pull/1361).  This was not done before, but the `haskell-ci` version that I built apparently adds it.  I suppose another solution would be to remove `XHaskell2010` altogether and let hlint figure out what the default extensions are.

liskin:
> It seems like the bug only fires when `-XHaskell2010` is passed to hlint (see [ndmitchell/hlint#1361](https://github.com/ndmitchell/hlint/pull/1361)). This was not done before, but the `haskell-ci` version that I built apparently adds it. I suppose another solution would be to remove `XHaskell2010` altogether and let hlint figure out what the default extensions are.

Oh, I see. I was looking at the wrong diff. Will give it some more thought once I get back from pub :-)

liskin:
> @liskin got any insights from the beer? :)

No real insights. I set a reminder for myself to check/drop the hlint patch in 2 months time (kinda hoped the fix would have been released by now) and forcepushed the stack.yaml fixes. Feel free to proceed with other repos, and sorry for the delay.

PR #700, opened 2022-04-06
new module XMonad.Actions.PerLayoutKeys

geekosaur:
### Description

This adds a new module `PerLayoutKeys` based on `XMonad.Actions.PerWorkspaceKeys`, which allows conditionalizing keybindings based on the current layout. This was requested by someone in IRC.

### Checklist

        manually, ...) and concluded: has to be manual

### DIFF
--- a/XMonad/Actions/PerLayoutKeys.hs
+++ b/XMonad/Actions/PerLayoutKeys.hs
@@ -1,0 +1,49 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Actions.PerLayouteKeys
+-- Description :  Define key-bindings on per-layout basis.
+-- Copyright   :  (c) brandon s allbery kf8nh 2022, Roman Cheplyaka, 2008
+-- License     :  BSD3-style (see LICENSE)
+--
+-- Maintainer  :  brandon s allbery kf8ng <allbery.b@gmail.com>
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- Define key-bindings on per-layout basis.
+--
+-----------------------------------------------------------------------------
+
+module XMonad.Actions.PerLayoutKeys (
+                                 -- * Usage
+                                 -- $usage
+                                 chooseActionByLayout,
+                                 bindByLayout
+                                ) where
+
+import XMonad
+import XMonad.StackSet as S
+
+-- $usage
+--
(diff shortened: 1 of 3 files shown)


XMonad/Actions/PerLayoutKeys.hs
```diff
@@ -0,0 +1,49 @@
...
+import XMonad
+import XMonad.StackSet as S
+
+-- $usage
+--
+-- You can use this module with the following in your @~\/.xmonad\/xmonad.hs@:
+--
+-- >  import XMonad.Actions.PerLayoutKeys
+--
+-- >   ,((0, xK_F2), bindOn [("Tall", spawn "rxvt"), ("Mirror Tall", spawn "xeyes"), ("", spawn "xmessage hello")])
```
liskin, line 32:
```suggestion
-- >   ,((0, xK_F2), bindByLayout [("Tall", spawn "rxvt"), ("Mirror Tall", spawn "xeyes"), ("", spawn "xmessage hello")])
```

geekosaur:
I always feel funny about merging my own changes….

slotThe:
Shouldn't we abstract this and have like a `ConditionalKeys` module where `perWorkspace` and `perLayout` are just functions?  Then we could also deprecate PerWorkspace and not have two modules with almost identical implementations

geekosaur:
Aside from people increasingly complaining — not just about xmonad, but pretty much all over the Haskell community — about that kind of breaking change. I'm thinking it might be better to got for simplicity over Haskelly "correctness", especially since the IRC user I mentioned already seems confused about the whole thing.

slotThe:
I mean, I don't think a deprecation would cause a lot of users to complain; especially when the upgrade path is "verbatim replace that function with that other function"

LSLeary:
I have my own spin on "conditional keybindings" which imo is quite powerful and user friendly, but it depends on the `grab` and `customRegrabEvHook` functions I originally exposed for modal keybindings. I was thinking of making it a PR after that one merges. [Preview in gist](https://gist.github.com/LSLeary/a9ec2edf43cdcc9ccf185e2d79d06e90).

Rather than layouts or workspaces, I actually use it to provide a distinct set of bindings on floats.

slotThe:
@LSLeary this look great to me!  Especially the use-case you described convinces me (and hopefully also @geekosaur :)

> but it depends on the grab and customRegrabEvHook functions I originally exposed for modal keybindings.

This should land in #703, so that seems fine to me.

Issue #489, opened 2021-03-21
Add an action that repeats the last action


TheMC47:
### Feature Description

In vim/emacs there's a keybinding that repeats the last command. We could add a module that wraps every bound action in a function that saves it in the extensible state. It should be straightforward to implement, I think it's suitable for new contributors. Feel free to ask for help :tada: 

### Example
Here's a rough sketch of how this should work. (`rememberActions` should work with both `EZConfig` and normal keybindings)
```haskell

myKeys = rememberActions [("M-q", kill), ....]
    ++ [("M-.", repeatLast)]

```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  ~- [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)~

yusi1:
This would be a really cool feature!

PR #705, opened 2022-04-14
added Actions.RepeatAction 

kozlov721:
repeat last action

### Description

Implementation of a feature described in [issue 489](https://github.com/xmonad/xmonad-contrib/issues/489).
Adds an ability to remember and repeat the last performed action.

### Checklist

### DIFF
--- a/XMonad/Actions/RepeatAction.hs
+++ b/XMonad/Actions/RepeatAction.hs
@@ -1,0 +1,84 @@
+{-# LANGUAGE LambdaCase #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Actions.RepeatAction
+-- Description :  Repeat the last performed action.
+-- Copyright   :  (c) 2022 Martin Kozlovsky
+-- License     :  BSD3-style (see LICENSE)
+--
+-- Maintainer  :  <kozlovsky.m7@gmail.com>
+-- Stability   :  unstable
+-- Portability :  not portable
+--
+-- Ability to repeat the last action.
+--
+-----------------------------------------------------------------------------
+
+module XMonad.Actions.RepeatAction (
+  -- * Usage
+  -- $usage
+  rememberAction,
+  rememberActions,
+  repeatLast,
+) where
+
+import XMonad
+
+import qualified XMonad.Util.ExtensibleState as XS
+
+-- $usage
+--
+-- You can use this module with the following in your @xmonad.hs@:
+--
+-- > import XMonad.Actions.RepeatAction
+--
+-- Then join a dedicated key to run the last action with the rest of your
+-- key bindings using the 'rememberActions':
+--
+-- > rememberActions (modm, xK_period) [((modm, xK_c), kill), …]
+--
+-- It can be also used in the same way for "XMonad.Util.EZConfig":
+--
+-- > rememberActions "M-." [("M-c", kill), …]
+--
+-- For example, if you use 'XMonad.Util.EZConfig.additionalKeysP',
+--
+-- > main = xmonad $ … $ def
+-- >   {
+-- >     …
+-- >   }
+-- >  `additionalKeysP` myKeys
+--
+-- you would adjust the call to 'XMonad.Util.EZConfig.additionalKeysP'
+-- like so:
+--
+-- > `additionalKeysP` (rememberActions "M-." myKeys)
+--
+-- For more detailed instructions on editing your key bindings, see
+-- <https://xmonad.org/TUTORIAL.html the tutorial>.
+
+newtype LastAction = LastAction { runLastAction :: X () }
+
+instance ExtensionClass LastAction where
+  initialValue = LastAction $ pure ()
+
+-- | Transforms an action into an action that can be remembered and repeated.
+rememberAction :: X () -> X ()
+rememberAction x = userCode x >>= \case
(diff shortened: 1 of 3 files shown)


XMonad/Actions/RepeatAction.hs
```diff
@@ -0,0 +1,60 @@
...
+-- Ability to repeat the last action.
+--
+-----------------------------------------------------------------------------
+
+module XMonad.Actions.RepeatAction (
+        -- * Usage
+        -- $usage
+          XMonad.Actions.RepeatAction.rememberAction
+        , XMonad.Actions.RepeatAction.rememberActions
+        , XMonad.Actions.RepeatAction.repeatLast
```
slotThe, line 21: You don't need to fully qualify the name here, a simple

```suggestion
         , rememberAction
         , rememberActions
         , repeatLast
```

suffices

XMonad/Actions/RepeatAction.hs
```diff
@@ -0,0 +1,60 @@
...
+-- add a dedicated key to run the last performed action:
+--
+-- > myKeys = ((modm, xK_period), repeatLast) : rememberActions [((modm, xK_c), kill), ...]
+--
+-- / Be careful not to include the key binding for the repetition in the remembered actions! /
+--
+-- For detailed instructions on editing your key bindings, see
+-- "XMonad.Doc.Extending#Editing_key_bindings".
+
+newtype LastAction = LastAction { runLastAction :: X () }
```
slotThe, line 44: Might be useful to also include an example using EZConfig, as a lot of people are using that nowadays

XMonad/Actions/RepeatAction.hs
```diff
@@ -0,0 +1,60 @@
...
+
+instance ExtensionClass LastAction where
```
slotThe, line 46: Indeed, that would be quite bad.  Perhaps then a better interface would be to not expose `repeatLast` individually, but only in combination with `rememberActions`?  Something like `rememberActions :: (KeyMask, KeySym) -> [((KeyMask, KeySym), X ())] -> XConfig a -> XConfig a`?  This could then be used as a combinator.  Might also be worth having a `rememberActionsP :: String -> [(String, X ())] -> XConfig l -> XConfig l`

XMonad/Actions/RepeatAction.hs
```diff
@@ -0,0 +1,60 @@
...
+-- You can use this module with the following in your @xmonad.hs@:
+--
+-- > import XMonad.Actions.RepeatAction
+--
+-- Then use the 'rememberActions' on the list of your keybindings and
+-- add a dedicated key to run the last performed action:
+--
+-- > myKeys = ((modm, xK_period), repeatLast) : rememberActions [((modm, xK_c), kill), ...]
+--
+-- / Be careful not to include the key binding for the repetition in the remembered actions! /
```
slotThe, line 39: Indeed, that would be quite bad.  Perhaps then a better interface would be to not expose `repeatLast` individually, but only in combination with `rememberActions`?  Something like `rememberActions :: (KeyMask, KeySym) -> [((KeyMask, KeySym), X ())] -> XConfig a -> XConfig a`?  This could then be used as a combinator.  Might also be worth having a `rememberActionsP :: String -> [(String, X ())] -> XConfig l -> XConfig l`

kozlov721: I'm a little confused about the type signature and how the function could be easily used, especially if one's key bindigs are a function of `XConfig`. So far I at least changed the interface to safer `rememberActions :: a -> [(a, X ())] -> [(a, X ())]` (which can be used nicely in infix form). I stil expose `repeatLast` as one might want to use it outside of theirs key bindings for some reason.

slotThe: So, for example, if one has a config like

``` haskell
main :: IO ()
main = xmonad
     . …
     $ def
       { …
       }
      `additionalKeysP` myKeys
```

Then with what I was thinking of (`rememberActionsP :: String -> [(String, X ())] -> XConfig l -> XConfig l`) this would just be transformed to

``` haskell
main :: IO ()
main = xmonad
     . …
     . rememberActionsP "M-." myKeys
     $ def
       { …
       }
```

This looks quite ergonomic to me.  Even if one doesn't want all keys to be remembered, this would then just incur adding `additionalKeysP` back to the bottom of main with some `myKeys'` instead.

The current implementation would give

``` haskell
main :: IO ()
main = xmonad
     . …
     $ def
       { …
       }
      `additionalKeysP` (rememberActions "M-." myKeys)
```

which isn't bad either, especially since one can do

``` haskell
      `additionalKeysP` myKeys'
      `additionalKeysP` (rememberActions "M-." myKeys'')
```


If one's keybindings are a function `XConfig -> Keybindings` then even today one has to directly override the `keys` field; in this case you could just write something like `keys = rememberActions "M-." . myKeys` (in your version; for the other `additionalKeys`-like `rememberActions` this would be a little bit more complicated, depending on the exact implementation) and it should be fine

kozlov721: I see. So can I mark this conversation as resolved since the current implementetion seems sufficient enough?

slotThe: I don't have any strong preferences here; if you like the current implementation I'd say let's go with that!

slotThe:
Thank you!

PR #707, opened 2022-04-14
buggy boolean blindness

### LINKED ISSUE #706: `debugManageHookOn` doesn't turn off
(this is basically a reminder for myself)

### Problem Description

I used `X.H.ManageDebug` to debug an issue with a `VirtualBox` window going into a tight loop on startup (which I still haven't solved, but it appears to have fixed itself in the meantime, possibly because it's some first-run greeting window). Afterward I noticed that the `logHook` portion continues to report on every invocation of the `logHook` instead of shutting itself off after the first run.

### Steps to Reproduce

1. Configure your `xmonad.hs` with `debugManageHookOn`
2. Trigger it with the defined key sequence
3. Note that it continues to report after the window appears, every time the `logHook` runs

### Configuration File

Please include the smallest _full_ configuration file that reproduces
the problem you are experiencing:

```haskell
module Main (main) where

import XMonad
import XMonad.Hooks.ManageDebug

main :: IO ()
main = xmonad $ debugManageHookOn "M-S-d" def
```

### Checklist

  - I tested my configuration
    - [x] With `xmonad` version 0.17.0.9 (commit 831ca49 if using git)
    - [x] With `xmonad-contrib` version 0.17.0.9 (commit 923c5a2 if using git)

geekosaur:
`ManageDebug` was continuing to report in the `logHook` even after
the `manageHook` was done. In diagnosing this, I discovered that
the original code was using a tuple of `Bool`s and not even a
comment about which meant what.

The code now uses a proper pair type, and dedicated `data`s for
the two flags that make it clear what each means. This also fixed
the bug, so apparently I had the `Bool`s confused somewhere.

I also took the chance to clarify the documentation a little (a
misleading "persistent", since it doesn't use persistent XS) and
a few more cleanups. Also, it now logs all `manageHook` runs
before the `logHook` in case multiple windows are opened.

Fixes #706.

### Checklist

### DIFF
--- a/XMonad/Hooks/ManageDebug.hs
+++ b/XMonad/Hooks/ManageDebug.hs
@@ -28,16 +28,17 @@ module XMonad.Hooks.ManageDebug (debugManageHook
                                 ) where
 
 import           XMonad
-import           XMonad.Prelude (when)
 import           XMonad.Hooks.DebugStack
 import           XMonad.Util.DebugWindow
 import           XMonad.Util.EZConfig
 import qualified XMonad.Util.ExtensibleState                                                 as XS
 
--- persistent state for manageHook debugging to trigger logHook debugging
-newtype ManageStackDebug = MSD (Bool,Bool)
+-- state for manageHook debugging to trigger logHook debugging
+data MSDFinal = DoLogHook | SkipLogHook deriving Show
+data MSDTrigger = MSDActivated | MSDInactive deriving Show
+data ManageStackDebug = MSD MSDFinal MSDTrigger deriving Show
 instance ExtensionClass ManageStackDebug where
-  initialValue = MSD (False,False)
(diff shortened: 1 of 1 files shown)


geekosaur:
Well, that's a "whoops". Let's try that again.

geekosaur:
Should I be worried about the lts-19 failures? I got those from GHA when I branched my fork, too (and am still somewhat croggled that we have things set up so forks get GHA, although I guess there may be no way to control that?)

slotThe:
> Should I be worried about the lts-19 failures?

FWIW, it works locally for me and the tests on my work [all succeed](https://github.com/slotThe/xmonad-contrib/actions/runs/2171290508).  I just reran the tests here but no luck; perhaps it's some cache issue with this repo?

slotThe (approved):
LGTM! :thumbsup:

PR #710, opened 2022-04-18
migrate build scripts from xmonad-testing

geekosaur:
The testing repo was not a good location for them, and is now
deprecated. They are now in `scripts/build`, and will be documented
in the main repo's `INSTALL.md`.


geekosaur:
The fails are all in ghcup; I would not be surprised if this is due to haskell.org/ghcup being down.

TheMC47 (approved):
Thank you!

PR #714, opened 2022-04-30
Clean up the code.

Ahanaf-Ether:
### Description

behind them.

### Checklist


Ahanaf-Ether:
[Taken from](https://github.com/altercation/dotfiles-tilingwm/blob/master/.xmonad/lib/XMonad/Actions/ConditionalKeys.hs)

Ahanaf-Ether:
Code is practically the same

slotThe (commented):
I now remember bringing this up in https://github.com/xmonad/xmonad-contrib/pull/700#issuecomment-1090572232 when PerLayoutKeys was added.  The consensus seemed to be that the additional complexity is not worth having users search around for this.  Plus, the use-case seems to be quite narrow beyond the two modules that we have now

Ahanaf-Ether:
i agree with you on this one.

Ahanaf-Ether:
@geekosaur might as well merge it?

slotThe:
> @geekosaur might as well merge it?

As you will see in the linked discussion, it was geekosaur who suggested that the additional of such a general module is probably unecessary.

Ahanaf-Ether:
i will consider it as a no. But i will however make a new PR anyways. Making it shorter

Ahanaf-Ether:
> i will consider it as a no. But i will however make a new PR anyways. Making it shorter

geekosaur:
The big problem with generalizing the code isn't just that it's got few uses besides the two already there; it's that the user has to mess with ugly internals to do either of those things with the generalized one. It just doesn't seem to pull its weight while being user friendly (well, to the extent that Haskell ever is — but that also does weigh into it, it's complicated enough as it is).

PR #717, opened 2022-05-03
X.H.EwmhDesktops: Add `_NET_DESKTOP_VIEWPORT` support

Pachin0:
### Description

My implementation of _NET_DESKTOP_VIEWPORT. I don't know how spaghetti it may be since I am not a seasoned haskeller. Part of what makes this so much bigger than the other functions is the sorting part since I don't know how to use the sort' that is already on the log hook of EwmhDesktops since it works with workspace sorting and I needed to sort tuples with workspaces. 

Also the sorting that I put in does not follow the X WorkspaceSort that is provided by the user since you'd have to now ask for instead of a workspace sort, a more general X WorkspaceCompare, which then gets ran though the mkWsSort and mkViewPortSort to both use the same comparison. But that is a question I'd rather ask first. Generally this is more code than what the other ewmh stuff takes to add.

Test this out with polybar, specifically with this commit. 

https://github.com/polybar/polybar/commit/bc9dda266f57f7ae3a2300a9418d94c3fe62c466

### Checklist

  - [] I've considered how to best test these changes (property, unit,

  - [] I updated the `CHANGES.md` file (Ill do this if this even goes through)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -181,6 +181,11 @@
     - Added support for dynamic scratchpads in the form of
       `dynamicNSPAction` and `toggleDynamicNSP`.
 
+  * `XMonad.Hooks.EwmhDesktops`
+
+    - Added support for `_NET_DESKTOP_VIEWPORT`, which is required by
+      some status bars.
+
 ### Other changes
 
   * Migrated the sample build scripts from the deprecated `xmonad-testing` repo to
--- a/XMonad/Hooks/EwmhDesktops.hs
+++ b/XMonad/Hooks/EwmhDesktops.hs
@@ -293,6 +293,10 @@ instance ExtensionClass WindowDesktops where initialValue = WindowDesktops (M.si
 newtype ActiveWindow = ActiveWindow Window deriving Eq
 instance ExtensionClass ActiveWindow where initialValue = ActiveWindow (complement none)
 
+-- | Cached @_NET_DESKTOP_VIEWPORT@
+newtype MonitorTags = MonitorTags [WorkspaceId] deriving (Show,Eq)
+instance ExtensionClass MonitorTags where initialValue = MonitorTags []
+
 -- | Compare the given value against the value in the extensible state. Run the
 -- action if it has changed.
 whenChanged :: (Eq a, ExtensionClass a) => a -> X () -> X ()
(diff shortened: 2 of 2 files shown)


Pachin0:
There was a spot where I forgot to change <+> to <>. Ill fix that with next commit if this is desired.

slotThe:
[ Haven't looked too deeply yet ]

> Part of what makes this so much bigger than the other functions is the sorting part since I don't know how to use the sort' that is already on the log hook of EwmhDesktops since it works with workspace sorting and I needed to sort tuples with workspaces.
>
> Also the sorting that I put in does not follow the X WorkspaceSort that is provided by the user since you'd have to now ask for instead of a workspace sort, a more general X WorkspaceCompare, which then gets ran though the mkWsSort and mkViewPortSort to both use the same comparison. But that is a question I'd rather ask first. Generally this is more code than what the other ewmh stuff takes to add.

Seems to me, since you're ignoring the `Position`s anyways and only sort on the `WorkspaceId`, that you can make it work with the normal `workspaceSort`.  In `mkViewPorts` you have the `WindowSet` so you can certainly extract a list of `WindowSpace`s (your `vWs` and `hWs` are already screens, so you'd just have to call [workspace](https://hackage.haskell.org/package/xmonad/docs/XMonad-StackSet.html#t:Screen) on them) and sort that with the users preferred sorting function before adding positions and the like.

Pachin0:
The problem with sorting the workspaces first and then adding the position is that to add the positions I have to find the tag of the visible workspaces and then add the position. It is much more convenient to create the list of tuples with the respective positions, sort by the tags and then just discard the workspace id's. The solution to doing it the way you suggested would be by doing a Map and then just updating the keys, although I feel it is much more resource intensive than just the simple list and sort.

slotThe:
Well, you could also, instead of having a list of `WorkspaceId`s, just have a list of `WindowSpace`s and keep the current order of adding positions and sorting

Pachin0:
Small detail. The reason why I didn't go all the way and just used the [WindowSpace] instead of the [WorkspaceId] is because WindowSpace is not a member of Ord and M.fromList wants it to have Ord. Probably the same after compilation. Don't know how Ord would be implemented for WindowSpace

slotThe (commented):
Thanks!  This looks fine to me.  I pushed a commit with mostly stylistic changes; if you agree then I think we can add an entry to CHANGES, squash, and merge

Pachin0:
This looks great. If you want go ahead.

slotThe:
Thanks!

PR #718, opened 2022-05-08
Extend X.U.Run with an EDSL for spawning processes

slotThe:
### Description

This extends X.U.Run with an EDSL for spawning (external) processes.  I will let an example from the documentation speak for itself:

``` haskell
  do url <- getSelection  -- from XMonad.Util.XSelection
     proc $ inEmacs
        >-> withEmacsLibs [ElpaLib "dash", ElpaLib "s", OwnFile "arXiv-citation"]
        >-> asBatch
        >-> eval (elispFun $ "arXiv-citation" <> asString url)
```

is essentially equivalent to (line breaks mine)

``` shell
  /usr/bin/sh -c "emacs -L /home/slot/.config/emacs/elpa/dash-20220417.2250
                        -L /home/slot/.config/emacs/elpa/s-20210616.619
                        -l /home/slot/.config/emacs/lisp/arXiv-citation.el
                        --batch
                        --eval '(arXiv-citation \"<url-in-the-primary-selection>\")'"
```

##### X.U.Run: Add an EDSL to spawn external programs

See above.

##### X.U.Run: Improve and add documentation

This adds documentation for the new EDSL, as well as small fixes to existing docs.

Importantly, I've added myself as a maintainer of the file (even though we don't really care about this at this point) and updated the copyright; the changes seem large enough to warrant this.

### Checklist

        manually, ...) and concluded: Tested manually; confirmed to work.  Most of the logic is trivial, and in X, so not sure how hard I should think about further tests.

### DIFF
--- a/XMonad/Util/Run.hs
+++ b/XMonad/Util/Run.hs
@@ -86,7 +86,7 @@ import qualified XMonad.Util.ExtensibleConf as XC
 
 import Codec.Binary.UTF8.String (encodeString)
 import Control.Concurrent (threadDelay)
-import System.Directory (getDirectoryContents, getHomeDirectory)
+import System.Directory (getDirectoryContents)
 import System.IO
 import System.Posix.IO
 import System.Posix.Process (createSession, executeFile, forkProcess)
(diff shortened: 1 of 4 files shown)


XMonad/Util/Run.hs
```diff
@@ -180,3 +237,222 @@ spawnPipe' encoding x = io $ do
...
+-- | Convenient type alias.
+type Input = ShowS
+
+-----------------------------------------------------------------------
+-- Combinators
+
+-- | Combine inputs together.
+(>->) :: X Input -> X Input -> X Input
+(>->) = liftA2 (.)
+infixr 3 >->
```
slotThe, line 346: I know that this is the `Semigroup` instance for `X`, but I thought having a pipe symbol would be cute.  Not entirely sure about this though, since everywhere else we're composing things with `<>` (or `>>`).

slotThe:
I'm pretty happy with the ergonomics of this now.  Not like the code is very complicated, so I'll merge this towards the weekend if no one objects

PR #722, opened 2022-05-29
Complete #708 by updating _NET_SUPPORTED.

### LINKED ISSUE #708: _NET_DESKTOP_VIEWPORT Support in ewmh
Recently I was looking at other panels other than xmobar and saw that panels, such as polybar grab the window name, workspace list, active workspace from the ewmh compliant window managers such as I3, bspwm, etc. But I found out that polybar for instance, does not work smoothly with xmonad's worspaces because xmonad's ewmh module for some reason does not set _NET_DESKTOP_VIEWPORT
Which is vital for polybar and other panels to know how to know, which monitor is displaying what workspace when there is a multi-monitor setup. In the current case, if polybar is run on a multi-monitor setup with xworspaces setup polybar gets confused between which workspace to mark as focused with the workspace on the second or third monitor. For polybar to work in multimonitor setup, there is an option for the xworspaces module that allows polybar to only display the workspace that is on the current screen, that way you run 2 panels and each tells you which workspace is displaying. XMonad I believe could just set _NET_DESKTOP_VIEWPORT. That way we can leverage the more elegant ewmh solution for workspaces in panels. 

Forgive if I don't know how to implement this myself, I read some discussions about polybar's pin-workspace functionality and they mentioned that that works with _NET_DESKTOP_VIEWPORT, then I searched the xmonad ewmh source code and confirmed it wasn't implemented in anyway. 

Here is the polybar discussion about viewport 
https://github.com/polybar/polybar/issues/840

geekosaur:
### Description

#708 is incomplete because it doesn't add `_NET_DESKTOP_VIEWPORT` to `_NET_SUPPORTED`. This update makes that addition, which should make polybar happy. Closes #708.

### Checklist

        manually, ...) and concluded: needs to be manual

### DIFF
--- a/XMonad/Hooks/EwmhDesktops.hs
+++ b/XMonad/Hooks/EwmhDesktops.hs
@@ -511,6 +511,7 @@ setSupported = withDisplay $ \dpy -> do
                          ,"_NET_ACTIVE_WINDOW"
                          ,"_NET_WM_DESKTOP"
                          ,"_NET_WM_STRUT"
+                         ,"_NET_DESKTOP_VIEWPORT"
                          ]
     io $ changeProperty32 dpy r a aTOM propModeReplace (fmap fromIntegral supp)
 


slotThe (approved):
LGTM!  I don't have the time to test this right now, but feel free to merge once we hear back from @compromyse (or someone else tests it)

geekosaur:
@compromyse, can you test this? Either copy the files directly over your tree, or install [the GitHub command line tool](https://cli.github.com) and do `gh pr checkout 722`.

compromyse:
I have tested this and it does seem to work

PR #725, opened 2022-06-17
CenteredIfSingle allow resizing in both dimensions

emptyflask:
### Description

Modified CenteredIfSingle to accept a second ratio, so that centered windows can be scaled in both directions.

### Checklist

  - [x] I've considered how to best test these changes (manually) and concluded: it works, and is an improvement IMHO

  - [ ] I updated the `CHANGES.md` file (not sure if this is necessary, since this modifier isn't in a released version yet)

### DIFF
--- a/XMonad/Layout/CenteredIfSingle.hs
+++ b/XMonad/Layout/CenteredIfSingle.hs
@@ -1,5 +1,5 @@
-{-# LANGUAGE TypeSynonymInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeSynonymInstances  #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -26,8 +26,8 @@ module XMonad.Layout.CenteredIfSingle
   ) where
 
 import XMonad
-import XMonad.Prelude (fi)
 import XMonad.Layout.LayoutModifier
+import XMonad.Prelude (fi)
 
 -- $usage
 -- You can use this module by including  the following in your @~\/.xmonad/xmonad.hs@:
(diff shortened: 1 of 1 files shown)


slotThe (commented):
Thanks for the contribution!

We can't really just introduce a breaking change like that; some users would be very annoyed!  What we could do is change `CenteredIfSingle` into a pattern for `CenteredIfSingleXY` (definitely not a name anyone should use) and then whatever second argument would mimic the current functionality

emptyflask:
This isn't in an official release yet, so I don't know if it counts as a breaking change, but I completely understand not wanting to alter the existing function anyway.

As for the alternate method name,  I don't really mind `CenteredIfSingleXY`. Maybe `ScaledAndCenteredIfSingle`? I'm not sure. Naming things is hard...

@elkowar since this is your module, any thoughts? And thanks for making it in the first place, I've wanted something like this ever since I first started using xmonad.

slotThe:
> This isn't in an official release yet, so I don't know if it counts as a breaking change, but I completely understand not wanting to alter the existing function anyway.

Oh, indeed, I thought this made it in just before 0.17.0 hit.  In that case I'm more open to just changing this; though input from other people would be appreciated!

XMonad/Layout/CenteredIfSingle.hs
```diff
@@ -36,30 +36,36 @@ import XMonad.Layout.LayoutModifier
...
-    start = fi rx + sides
-    width = fi $ fi rw - sides * 2
+    left = floor $ fi rw * (1.0 - ratioX) / 2
+    top = floor $ fi rh * (1.0 - ratioY) / 2
+
+    startX = fi rx + left
+    width = fi $ fi rw - left * 2
+
+    startY = fi ry + top
+    height = fi $ fi rh - top * 2
```
TheMC47, line 71: There's a lot of repetition here, maybe we should take care of that

TheMC47 (changes requested):
Thanks for the review! I have just a small comment.

TheMC47:
> > This isn't in an official release yet, so I don't know if it counts as a breaking change, but I completely understand not wanting to alter the existing function anyway.
> 
> Oh, indeed, I thought this made it in just before 0.17.0 hit. In that case I'm more open to just changing this; though input from other people would be appreciated!

I think it's worth it to break the API!

XMonad/Layout/CenteredIfSingle.hs
emptyflask: @TheMC47 I made some changes here: [e576f85](https://github.com/xmonad/xmonad-contrib/pull/725/commits/e576f85bf32820b328c768a0228ac3c3167bf5ec)

XMonad/Layout/CenteredIfSingle.hs
```diff
@@ -36,30 +36,42 @@ import XMonad.Layout.LayoutModifier
...
 instance LayoutModifier CenteredIfSingle Window where
-  pureModifier (CenteredIfSingle ratio) r _ [(onlyWindow, _)] = ([(onlyWindow, rectangleCenterPiece ratio r)], Nothing)
+  pureModifier (CenteredIfSingle ratioX ratioY) r _ [(onlyWindow, _)] = ([(onlyWindow, rectangleCenterPiece ratioX ratioY r)], Nothing)
   pureModifier _ _ _ winRects = (winRects, Nothing)
 
--- | Layout Modifier that places a window in the center of the screen, 
--- leaving room on the left and right if there is only a single window
+-- | Layout Modifier that places a window in the center of the screen,
+-- leaving room on all sides if there is only a single window
 centeredIfSingle :: Double -- ^ Ratio of the screen the centered window should take up. Should be a value between 0.0 and 1.0
```
slotThe, line 54: Might be good to write "horizontal ratio" here now

slotThe (commented):
> I think it's worth it to break the API!

Let's do it then, I suppose :)

XMonad/Layout/CenteredIfSingle.hs
TheMC47: Awesome, thanks

TheMC47 (approved):

PR #727, opened 2022-06-22
X.H.UrgencyHook: Default for DzenUrgencyHook and UrgencyConfig

ilya-bobyr:
As the XMonad config is commonly customized by saying

    def { startupHook = ...
        , manageHook = ...
        , ...
        }

It seems consistent to allow the same for an individual hook config:

    let urgencyHook = def { suppressWhen = ...
                          , remindWhen = ...
                          }

### Description

behind them.

### Checklist

        manually, ...) and concluded: this is configuration only change, if it compiles, it is good.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,11 @@
     - Deprecated the module; use the new dynamic scratchpad
       functionality of `XMonad.Util.NamedScratchpad` instead.
 
+  * `XMonad.Hooks.UrgencyHook`
+
+    - Deprecated `urgencyConfig`; use `def` from the new `Default`
+      instance of `UrgencyConfig` instead.
+
 [on the website]: https://xmonad.org/configurations.html
 
 ### New Modules
@@ -198,6 +203,10 @@
 
     - Added an EDSL—particularly geared towards programs like terminals
       or Emacs—to spawn processes from XMonad in a compositional way.
+
+  * `XMonad.Hooks.UrgencyHook`
+
+    - Added a `Default` instance for `UrgencyConfig` and `DzenUrgencyHook`.
 
 ### Other changes
 
--- a/XMonad/Hooks/UrgencyHook.hs
+++ b/XMonad/Hooks/UrgencyHook.hs
@@ -197,11 +197,11 @@ import Foreign.C.Types (CLong)
 -- instead.
 withUrgencyHook :: (LayoutClass l Window, UrgencyHook h) =>
                    h -> XConfig l -> XConfig l
-withUrgencyHook hook = withUrgencyHookC hook urgencyConfig
+withUrgencyHook hook = withUrgencyHookC hook def
 
 -- | This lets you modify the defaults set in 'urgencyConfig'. An example:
 --
--- > withUrgencyHookC dzenUrgencyHook { ... } urgencyConfig { suppressWhen = Focused }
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -191,6 +191,11 @@
     - Added an EDSL—particularly geared towards programs like terminals
       or Emacs—to spawn processes from XMonad in a compositional way.
 
+  * `XMonad.Hooks.UrgencyHook`
+
+    - Added an `Default` instance for `UrgencyConfig` and
```
slotThe, line 196:
```suggestion
     - Added a `Default` instance for `UrgencyConfig` and
```

XMonad/Hooks/UrgencyHook.hs
```diff
@@ -254,7 +254,13 @@ minutes secs = secs * 60
 -- Use a variation of this in your config just as you use a variation of
 -- 'def' for your xmonad definition.
 urgencyConfig :: UrgencyConfig
-urgencyConfig = UrgencyConfig { suppressWhen = Visible, remindWhen = Dont }
+urgencyConfig = def
+
+-- | The default 'UrgencyConfig'. suppressWhen = Visible, remindWhen = Dont.
+-- Use a variation of this in your config just as you use a variation of
```
slotThe, line 260: We probably want to at least put these in a monospace font (between `@`s)

slotThe:
Additionally, we probably now want to deprecate `urgencyConfig` and `dzenUrgencyHook` (or would that be too annoying for users?  I don't know)

CHANGES.md
ilya-bobyr: Done

XMonad/Hooks/UrgencyHook.hs
ilya-bobyr: Wrapped `suppressWhen = Visible, remindWhen = Dont` in `@`s.
Is this what you had in mind?

ilya-bobyr:
> Additionally, we probably now want to deprecate `urgencyConfig` and `dzenUrgencyHook` (or would that be too annoying for users? I don't know)

I do not have a strong opinion, though I also do not have a lot of Haskell experience %)
It does seem reasonable to deprecate `urgencyConfig` in favor of the `Default` instance.
Not sure why was `Default` was not used in the first place.  I assumed there might be a reason I was not aware of.

As for `dzenUrgencyHook` - I do not think it should be deprecated.
There could be other urgency hook handlers.  And `borderUrgencyHook` and `spawnUrgencyHook` are already defined.
Most code will probably be more readable if it is using `dzenUrgencyHook` directly, instead of the `def` instance.

In short, I could add deprecated to `urgencyConfig`.
Having one way to do simple things is good :)

XMonad/Hooks/UrgencyHook.hs
slotThe: Yes exactly; Haddock will print this in a monospaced font, which I think looks nicer for code

slotThe:
> In short, I could add deprecated to `urgencyConfig`. Having one way to do simple things is good :)

This seems fine to me; let's do it!

ilya-bobyr:
> > In short, I could add deprecated to `urgencyConfig`. Having one way to do simple things is good :)
> 
> This seems fine to me; let's do it!

Added :)

slotThe:
Thank you!

PR #729, opened 2022-06-29
X.H.Modal: Add EZConfig support

slotThe:
### Description

I had a little bit of time on my hands today (yay, food poisoning), so I figured I'd try my hands at hacking EZConfig support into X.H.Modal, as discussed in https://github.com/xmonad/xmonad-contrib/pull/703.  Worked out pretty well, I think.

##### X.H.Modal: Cleanup

Mostly small things, like making imports line up with the provided comments.  Also:

+ Rename mode' -> modeWithExit.  This seems like a better name for discoverability reasons.

+ Make the fields of Mode strict, because they have no reason not to, really.

##### X.H.Modal: Add support for EZConfig-style bindings

Add support for EZConfig-style bindings while also maintaining some guarantees as to which type of representation we will store in the extensible state.  This means that parsing of the keys will happen no later than the call to `modal`.

Users can choose to use `mkKeysEz` or `mkKeysFun` to create a new collection of keys to bind for a mode.  This is deemed more ergonomic than exporting the respective constructors directly.

##### X.H.Modal: Update documentation

Now that we use a slightly different setup, as well as EZConfig support, rewrite the introduction to the module a little.

### Checklist

        manually, ...) and concluded: it's more-or-less a straightforward refactor; I've tested manually that it doesn't cause things to explode and it seems fine.

    The changelog entry for `X.H.Modal` probably does not need to be modified.

### DIFF
--- a/XMonad/Hooks/Modal.hs
+++ b/XMonad/Hooks/Modal.hs
@@ -84,6 +84,8 @@ import           XMonad.Util.Parser            ( runParser )
 -- >
 -- > import XMonad
 -- > import XMonad.Hooks.Modal
+-- > import XMonad.Util.EZConfig
+-- > import qualified Data.Map as M
 -- >
 -- > main :: IO ()
 -- > main =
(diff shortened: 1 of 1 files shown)


XMonad/Hooks/Modal.hs
```diff
@@ -89,35 +100,107 @@ import           XMonad.Util.Loggers
...
+
+-- | The mode type. Use 'mode' or 'modeWithExit' to create modes.
+data Mode k = Mode
+  { label     :: !String
+  , boundKeys :: !k
   }
 
+-- | A convenient alias for specifying new modes.
+type NewMode = Mode BoundKeys
+
```
TheMC47, line 197: Wouldn't 
```haskell
data Mode = Mode
  { label     :: !String
  , boundKeys :: !BoundKeys
  }
```
do the work? I don't see where we're setting the type variable to anything else

TheMC47 (approved):
This is awesome, thanks! I hope the food poisoning wasn't too bad

Anyways, I'm just a bit confused on one thing (just see the comment). Other than that LGTM

XMonad/Hooks/Modal.hs
slotThe: > I don't see where we're setting the type variable to anything else

For the extensible config/state we're using

``` haskell
newtype ModeConfig = MC [Mode KeysIntern] deriving Semigroup
```

and

``` haskell
newtype CurrentMode = CurrentMode
  { currentMode :: Maybe (Mode KeysIntern)
  }
```

This is the main reason for both the type variable there, as well as the `GetBoundKeys` class.  If we just had, for example, `newtype ModeConfig = MC [Mode]` (with a baked in `BoundKeys`) then we couldn't—at the type level—ensure that the internal representation of `XConfig Layout -> Map (ButtonMask, KeySym) (X ())` is used in the state.  We'd have to do it via some note or contract of "please do this", which seemed a bit dodgy to me; especially since the solution isn't much more complicated.

TheMC47: Right, but since we're not exposing the types anyway, do we really need `BoundKeys`? What I actually meant was

```haskell
data Mode = Mode
  { label     :: !String
  , boundKeys :: !KeysIntern
  }
```
with the following interface:

```haskell
sayHelloMode :: NewMode
sayHelloMode = mode "Hello" $ mkKeysEz
  [ ("h", xmessage "Hello, World!")
  , ("g", xmessage "Goodbye, World!")
  ]


sayHelloMode :: NewMode
sayHelloMode = mode "Hello" $
  const (M.fromList [ ((noModMask, xK_h), xmessage "Hello, World!")
                    , ((noModMask, xK_g), xmessage "Goodbye, World!")
                    ])
```

and the new signatures (eventually renaming `KeysIntern`):
```
mkKeysEz :: [(String, X ())] -> KeysIntern
modeWithExit :: String -> String -> KeysIntern -> NewMode
```

slotThe: Oh!  Why yes, this make everything a lot easier :sweat_smile:

TheMC47 (approved):

Issue #728, opened 2022-06-28
NamedScratchpads somewhat broken.


exorcist365:
### Problem Description

After commit 3fc830aa09368dca04df24bf7ec4ac817f2de479 my scratchpads suddenly stopped working after I restarted the window manager, it does however fix itself after opening a new window but having to do that after every time I restart the wm and want to bring my scratchpad is kinda... annoying, don't you think?

### Steps to Reproduce

Run a version of xmonad-contrib pushed after the 1st of April, restart the window manager and then press your scratchpad keybind.

### Configuration File

Quite difficult to post here since my setup is split into multiple files, so you better take a look at my [dotfiles](https://gitlab.com/exorcist365/dotfulls/-/tree/master/.config/xmonad/src)

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17 (commit master)
    - [x] With `xmonad-contrib` version 0.17 (commit master and the one mentioned above)

geekosaur:
Are you using NamedScratchpad or standard Scratchpad? I suspected the new scratchpad implementation would break the latter, and I'm surprised it's taken this long for a report to come in.

geekosaur:
Sorry, I missed the title. I've been using several NamedScratchpads without issue since that commit, so I'm not sure what could be wrong.

slotThe:
> ### Configuration File
>
> Quite difficult to post here since my setup is split into multiple files, so you better take a look at my [dotfiles](https://gitlab.com/exorcist365/dotfulls/-/tree/master/.config/xmonad/src)

Could you perhaps try to produce a (minimal) example with which you can reproduce this?  Much like @geekosaur, I've been using scratchpads extensively since this commit and I've never noticed this

exorcist365:
ok, so:
```hs
fmSP, ghciSP, ocamlSP, musicSP, terminalSP :: String
fmSP = "FFF"
ghciSP = "GHCI"
ocamlSP = "OCaml"
musicSP = "Music"
terminalSP = "Terminal"

scratchpads :: [NamedScratchpad]
scratchpads = [ NS { name = ghciSP
                   , query = isGHCI
                   , cmd = spawnGHCI
                   , hook = centerFloat
                   }
              , NS { name = ocamlSP
                   , query = isOcaml
                   , cmd = spawnOcaml
[...]

PR #730, opened 2022-06-29
X.U.NamedScratchpad: Initialise if necessary

slotThe:
### Description

Fixes: https://github.com/xmonad/xmonad-contrib/issues/728

Since 3fc830aa09368dca04df24bf7ec4ac817f2de479, scratchpads are now added in namedScratchpadManageHook.  This, however, means that we need some kind of MapRequestEvent to happen before processing scratchpads, otherwise the manageHook didn't run yet and our extensible state is being left empty.  When trying to open a scratchpad right after starting xmonad—i.e., before having opened a window—this may not be the case.

Here is a small config to reproduce the problem.  When first starting up xmonad, pressing `M-m` does not do anything since we haven't opened a window yet.  After, e.g., spawning a terminal everything works.

``` haskell
import XMonad
import XMonad.StackSet
import XMonad.Util.EZConfig (additionalKeysP)
import XMonad.Util.NamedScratchpad

main :: IO ()
main = xmonad $ def
  { manageHook = manageHook def <> namedScratchpadManageHook scratchpads }
 `additionalKeysP` [("M-m", namedScratchpadAction scratchpads "xterm")]

scratchpads :: [NamedScratchpad]
scratchpads =
  [NS "xterm"
      "xterm"
      (className =? "XTerm")
      (customFloating $ RationalRect (1 / 6) (1 / 6) (2 / 3) (2 / 3))]
```

### Checklist

        manually, ...) and concluded: Tested it manually and it at least works for me.  @exorcist365

    Don't think this is necessary.

### DIFF
--- a/XMonad/Util/NamedScratchpad.hs
+++ b/XMonad/Util/NamedScratchpad.hs
@@ -131,9 +131,16 @@ instance ExtensionClass NSPState where
   initialValue :: NSPState
   initialValue = NSPState mempty
 
--- | Construct an 'NSPState' from an ordinary list of scratchpads.
-buildNSPState :: NamedScratchpads -> NSPState
-buildNSPState nsps = NSPState . Map.fromList $ zip (map name nsps) nsps
+-- | Try to fill the 'NSPState' with the given list of scratchpads.  In
+-- case the state is already non-empty, don't do anything and return
+-- that state.  Otherwise, fill the state with the given scratchpads.
+fillNSPState :: NamedScratchpads -> X NSPState
+fillNSPState nsps = do
+    nsp@(NSPState xs) <- XS.get
+    let nspState = NSPState . Map.fromList $ zip (map name nsps) nsps
+    if null xs
+        then nspState <$ XS.put nspState
+        else pure nsp
 
 -- | Manage hook that makes the window non-floating
 nonFloating :: ManageHook
@@ -160,7 +167,11 @@ runApplicationHere = spawnHere . cmd
 
 -- | Action to pop up specified named scratchpad
 --
--- NOTE: Ignores its first argument and uses 'NSPState' instead.
+-- Note [Ignored Arguments]: Most of the time, this function ignores its
+-- first argument and uses 'NSPState' instead.  The only time when it
+-- does not is when no other window has been opened before in the
+-- running xmonad instance.  If this is not your use-case, you can
+-- safely call this function with an empty list.
 namedScratchpadAction :: NamedScratchpads -- ^ Named scratchpads configuration
                       -> String           -- ^ Scratchpad name
                       -> X ()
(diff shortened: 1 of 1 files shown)


exorcist365:
Thank you!

exorcist365:
It does indeed work.

XMonad/Util/NamedScratchpad.hs
```diff
@@ -219,14 +226,16 @@ nsHideOnFocusLoss scratches = withWindowSet $ \winSet -> do
 
 -- | Execute some action on a named scratchpad.
 --
--- NOTE: Ignores its first argument and uses 'NSPState' instead.
+-- NOTE: Most of the time, this function ignores its first argument and
+-- uses 'NSPState' instead.  The only time when it does not is when no
+-- other window has been opened before in the running xmonad instance.
```
liskin, line 231: Those other `NOTE`s for functions like `spawnHereNamedScratchpadAction` and `allNamedScratchpadAction` etc. are now outdated as well. Maybe just refer to the note here instead of copypasting?

liskin (approved):
If only we had a way to ~~type~~factcheck comments in addition to code… :-))

XMonad/Util/NamedScratchpad.hs
geekosaur: I liked the ghc-style Note we had recently. Should we consider that for future notes as well?

liskin: That one wasn't rendered by haddock (iiuic, which I probably don't) while this one needs to be. But if it can be made to work (or perhaps even just works), why not.

geekosaur: That has nothing to do with the note style, it's just using the right markup to start the comment block.

geekosaur: To be more clear, this is rendered by the `-- |` and is where the new-style note would go. The others would then have `See Note …`.

slotThe: I think it's neat having something to refer to by name, so I've given the user-facing note a name for now

XMonad/Util/NamedScratchpad.hs
```diff
@@ -219,14 +233,15 @@ nsHideOnFocusLoss scratches = withWindowSet $ \winSet -> do
 
 -- | Execute some action on a named scratchpad.
 --
--- NOTE: Ignores its first argument and uses 'NSPState' instead.
+-- This function /almost always/ ignores its first argument; see Note
```
liskin, line 236: third, actually

XMonad/Util/NamedScratchpad.hs
```diff
@@ -160,23 +167,29 @@ runApplicationHere = spawnHere . cmd
...
+-- [Ignored Arguments] for 'namedScratchpadAction'.
 spawnHereNamedScratchpadAction :: NamedScratchpads           -- ^ Named scratchpads configuration
                                -> String                     -- ^ Scratchpad name
                                -> X ()
 spawnHereNamedScratchpadAction = customRunNamedScratchpadAction runApplicationHere
 
 -- | Action to pop up specified named scratchpad, given a custom way to initially start the application.
 --
--- NOTE: Ignores its second argument and uses 'NSPState' instead.
+-- This function /almost always/ ignores its first argument; see Note
```
liskin, line 191: second, actually

liskin (approved):
Final round of comment fact-checking. :-)

PR #731, opened 2022-07-03
Fix WindowNavigation borders vs. transparency

geekosaur:
### Description

If pixels have an alpha channel, both `pixelToString` and
`averagePixels` fail and the result is `BadValue` errors and black
borders. Mask out the alpha channels so they work correctly.

Closes https://github.com/xmonad/xmonad/issues/395, https://github.com/xmonad/xmonad/issues/398

### Checklist

### DIFF
--- a/XMonad/Util/Font.hs
+++ b/XMonad/Util/Font.hs
@@ -39,6 +39,7 @@ import XMonad.Prelude
 import Foreign
 import Control.Exception as E
 import Text.Printf (printf)
+import Data.Bits ((.&.))
 
 #ifdef XFT
 import qualified Data.List.NonEmpty as NE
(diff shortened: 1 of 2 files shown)


geekosaur:
Open question: should the masking be pushed down into the functions instead of being on their uses?

slotThe:
> Open question: should the masking be pushed down into the functions instead of being on their uses?

I would say so; seems like the easiest way to prevent these kind of footguns

liskin:
Seconded. Also, some explanatory comments about alpha not being welcome in Pixel for XQueryColor and the like might be good, and WindowNavigation isn't a good place for those.

geekosaur:
Pushed the masking down inside `averagePixels` and `pixelToString`, with accompanying Haddock comments about the masking and why it's necessary.

liskin (approved):

Issue #726, opened 2022-06-19
XMonad.Actions.SpawnOn.getPPIDOf assumes wrong format of proc/*/stat files


johannesgerer:
### Problem Description

manageHook crashes with `Prelude.read: no parse`, when trying to evaluate `XMonad.Actions.SpawnOn.getPPIDOf 1162897`:

```
cat /proc/1162897/stat
1162897 (tmux: server) S 1 1162897 1162897 0 -1 4194368 4578 25034 0 0 649 111 22 14 20 0 1 0 6731470 242098176 2747 18446744073709551615 4214784 4799021 140724568486368 0 0 0 0 3674116 134433283 0 0 0 17 3 0 0 0 0 0 5019376 5098152 12685312 140724568492670 140724568492736 140724568492736 140724568502201 0
```

### Steps to Reproduce

This is difficult to pull of, but the example makes it clear that following [code](https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/src/XMonad.Actions.SpawnOn.html#getPPIDOf):

```
getPPIDOf :: ProcessID -> Maybe ProcessID
getPPIDOf thisPid =
    case unsafePerformIO . tryJust (guard . isDoesNotExistError) . readFile . printf "/proc/%d/stat" $ toInteger (traceShowId thisPid) of
      Left _         -> Nothing
      Right contents -> case lines contents of
                          []        -> Nothing
                          first : _ -> case words first of
     (BUG IN THIS LINE)                  _ : _ : _ : ppid : _ -> Just $ fromIntegral (read ppid :: Int)
                                         _                    -> Nothing

```
assumes that the 4th word is the ppid which is not correct, as seen example file above.

geekosaur:
More precisely, the second field can have spaces, as shown in this case. Since all the other fields are numeric and it's also possible to put parentheses in a process name (and they don't have to match),m the correct way to parse the line is to find the final closing parenthesis and consider that the end of the second field.

johannesgerer:
Yes, what seems to work on my system is:

```
                          first : _ -> case words (dropWhile (/= ')') first) of
                                         _ : _ : ppid : _ -> Just $ fromIntegral (read ppid :: Int)
```

slotThe:
> Yes, what seems to work on my system is:
>
> ```
>                           first : _ -> case words (dropWhile (/= ')') first) of
>                                          _ : _ : ppid : _ -> Just $ fromIntegral (read ppid :: Int)
[...]

PR #732, opened 2022-07-04
X.A.SpawnOn, X.H.WindowSwallowing: Fix parsing of process PPIDs

liskin:
### Description

This fixes several issues related to parsing of parent PIDs:

* A process with lines or spaces or parentheses in its process name
  would confuse the code in X.A.SpawnOn and possibly lead to a
  `Prelude.read: no parse` exception.

* `X.H.WindowSwallowing.isChildOf` looked for the parent PID anywhere in
  the output of pstree, so single-digit parent PIDs would be considered
  as parents of any process with that digit anywhere in its chain of
  parent PIDs. (Note that apps in PID namespaces like in Flatpak often
  have single-digit PIDs.)

* `pstree` is no longer required in `$PATH`.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/726

### Checklist

  - [X] I've considered how to best test these changes (property, unit, manually, ...) and concluded:

    - I only tested `isChildOf`, `getPPIDOf` and `getPPIDChain` in ghci, so this definitely needs more testing. In particular, I'm not a user of SpawnOn, so to test that I'd need to set up a synthetic xmonad config and I don't think I'll be able to do that during the week. 😞


XMonad/Util/ExtensibleState.hs
```diff
@@ -89,12 +91,20 @@ modifyStateExts f = State.modify $ \st -> st { extensibleState = f (extensibleSt
...
--- the monadic environment.
+-- | Apply an action to a stored value of the matching type or the initial value if there
+-- is none.
+modifyM :: (ExtensionClass a, XLike m) => (a -> m a) -> m ()
+modifyM f = put =<< f =<< get
+
+-- | Like @modify@ but the result value is forced to WHNF before being stored.
 modify' :: (ExtensionClass a, XLike m) => (a -> a) -> m ()
-modify' f = (put $!) . f =<< get
+modify' = modifyM' . (pure .)
```
slotThe, line 103: Trying to think really hard whether `(pure .)` introduces too much lazyness here, but probably not (at least the core looks very similar)

slotThe (commented):
Code looks fine to me; will try to test this with `spawnOn` in the coming days

xmonad-contrib.cabal
```diff
@@ -454,6 +456,7 @@ test-suite tests
   build-depends: base
                , QuickCheck >= 2
                , X11 >= 1.10 && < 1.11
+               , bytestring >= 0.10 && < 0.12
```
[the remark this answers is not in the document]
liskin: SpawnOn is imported in some other module which is being tested in the test suite, so this drags in X.U.Process and a bytestring dep.

XMonad/Util/ExtensibleState.hs
slotThe: Ah, right, now I remember.  The bottleneck was `put` and since we didn't even use a strict version of `=<<` before, this also should not matter now.  And yeah, there might be a rewrite rule for `=<< pure` anyways

slotThe (approved):

PR #735, opened 2022-07-21
`X.P.Unicode`: `BS` -> `String` to support Unicode descriptions

### LINKED ISSUE #733: `X.P.Unicode`: support Unicode symbols in description
### Problem Description

`X.P.Unicode` allows you to work with Emoji in Unicode, but does not allow you to use
non-Latin characters in the description of these Emojis. `X.P.Unicode` works fine, but
cannot display such characters correctly.

image

I managed to write a patch to add support, but for this I had to abandon `ByteString`.
I don't know if this is acceptable or necessary achieve operability with `ByteString`...

### Steps to Reproduce

1. Install Noto font
1. Create a file with an Emoji description (`/home/user/.xmonad/unicode-data`):

```
1f47e;alien monster | द एलियन
```

3. Recompile the Xmonad config
4. Sparm Unicode prompt
5. Start typing characters to display completion

### Configuration File

```haskell
import XMonad
import XMonad.Prompt
import XMonad.Prompt.Unicode

-- Prompts
--  default prompt settings
myXPConfig :: XPConfig
myXPConfig = def
  { font = "xft:monospace:size=9,Noto Color Emoji:pixelsize=12:style=Regular,Noto Sans Devanagari UI"
  , height = 17
  , position = Top
  , promptBorderWidth = 0
  , historyFilter = deleteAllDuplicates
  , searchPredicate = fuzzyMatch
  }

myKeys conf@(XConfig {XMonad.modMask = modm}) = M.fromList
  [ ((modm .|. controlMask, xK_u), unicodePrompt "/home/user/.xmonad/unicode-data" myXPConfig)
  ]

main = do
  xmonad = def
    { keys = myKeys <+> keys def
    }
```

### Checklist

  - I tested my configuration
    - [x] With `xmonad` version XXX (commit xmonad/xmonad@ed5c866 if using git)
    - [x] With `xmonad-contrib` version XXX (commit ea97c35 if using git)

PRESFIL:
This PR changes `X.P.Unicode` string type from `ByteString` to `String` to support Unicode
in descriptions. Closes #733.

### Description

`X.P.Unicode` allows you to work with Emoji in Unicode, but does not allow you to use
non-Latin characters in the description of these Emojis. `X.P.Unicode` works fine, but
cannot display such characters correctly.

image

See how to reproduce in #733.

Chose `String` instead of `Text` because I thought it was possible to do without an extra
dependency for `xmonad-contrib`. The branch is editable if you know how to add support
without dropping the `ByteString`.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -112,6 +112,11 @@
     - Added the ability to specify alphabetic (`#A`, `#B`, and `#C`)
       [priorities] at the end of the input note.
 
+  * `XMonad.Prompt.Unicode`
+
+    - Fixed the display of non-ASCII characters in the description of Unicode
+      characters
+
   * `XMonad.Prompt`
 
     - Added `transposeChars` to interchange the characters around the
--- a/XMonad/Prompt/Unicode.hs
+++ b/XMonad/Prompt/Unicode.hs
@@ -23,12 +23,12 @@ module XMonad.Prompt.Unicode (
  mkUnicodePrompt
  ) where
 
+import Codec.Binary.UTF8.String (decodeString)
 import qualified Data.ByteString.Char8 as BS
 import Numeric
 import System.IO
 import System.IO.Error
 import Text.Printf
-import Control.Arrow (second)
 
 import XMonad
 import XMonad.Prelude
@@ -42,7 +42,7 @@ instance XPrompt Unicode where
   commandToComplete Unicode s = s
   nextCompletion Unicode = getNextCompletion
 
-newtype UnicodeData = UnicodeData { getUnicodeData :: [(Char, BS.ByteString)] }
+newtype UnicodeData = UnicodeData { getUnicodeData :: [(Char, String)] }
   deriving (Read, Show)
 
 instance ExtensionClass UnicodeData where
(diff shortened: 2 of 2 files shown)


slotThe (commented):
Looks fine to me, thanks!  I don't think the fact that we used ByteString here was particularly important, so dropping it is probably fine.  Could you add an entry to CHANGES.md?

liskin:
I imagine it might have been done for performance reasons as the list of all unicode codepoints is fairly large and doing it with String can be an order of magnitude slower. Whether that's actually a problem in practice, I have no idea. If it is then depending on text would probably be fine I guess.

PRESFIL:
OK, I updated `CHANGES.md`.

PRESFIL:
On the performance side, I've noticed performance drops in `XMonad.Prompt.Man`
when doing incremental searches. But there are a lot of man pages (5633 lines in prompt). 

Or when using `XMonad.Prompt.Zsh`. But the reason is `XMonad.Prompt`, it's entirely
writen on `String` base, so rewriting it is a separate issue.

PRESFIL:
On the performance side, I've noticed performance drops in `XMonad.Prompt.Man`
when doing incremental searches. But there are a lot of man pages (5633 lines in prompt). 

Or when using `XMonad.Prompt.Zsh`. But the reason is `XMonad.Prompt`, it's entirely
writen on `String` base, so rewriting it is a separate issue.

PRESFIL:
Do I have to update the changes according to master in order for this PR to be merged?

slotThe:
Sorry, seems like everyone was equally busy with other things :)

> On the performance side, I've noticed performance drops in `XMonad.Prompt.Man` when doing incremental searches. But there are a lot of man pages (5633 lines in prompt).
>
> Or when using `XMonad.Prompt.Zsh`. But the reason is `XMonad.Prompt`, it's entirely writen on `String` base, so rewriting it is a separate issue.

You have noticed these performance drops only _with this patch applied_?  Because to me it seems like X.P.Unicode is an extra prompt, not something that interacts with X.P.Man in any way.

PRESFIL:
I only noticed this in `X.P.Man`, it also does use `String` and lots of man pages.
It doesn't related with this patch.

I think this is the topic of a separate issue.

slotThe:
Thanks!

PRESFIL:
:rocket:

PR #736, opened 2022-07-25
print EWMH type and state in debugWindow

geekosaur:
### Description

behind them.

### Checklist


XMonad/Util/DebugWindow.hs
```diff
@@ -13,6 +13,8 @@
 -- "XMonad.Hooks.DebugEvents" and "XMonad.Hooks.DebugStack" for practical uses.
 --
 -----------------------------------------------------------------------------
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+{-# HLINT ignore "Evaluate" #-}
```
slotThe, line 17: We're ignoring this globally [here](https://github.com/xmonad/xmonad-contrib/blob/master/.hlint.yaml), so I guess this isn't necessary

geekosaur: Mm, I think those were added by vs code, which I've been experimenting with. It's not quite smart enough to find the global ones because it thinks the global context is my own config. I'll strip them back out.

PR #742, opened 2022-08-14
New module: XMonad.Layout.SideBorderDecoration

slotThe:
### Description

New module that allows for having a configurable border position around
windows.  Originally found in a comment by L. S. Leary in [1].  Enough
people have requested this on IRC in recent memory to warrant a proper
module.

[1]: https://github.com/xmonad/xmonad/issues/152#issuecomment-362716434

-------

### Checklist

        manually, ...) and concluded: Tested this manually; seems to work on every side.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -87,6 +87,11 @@
 
     This module implements modal keybindings for xmonad.
 
+  * `XMonad.Layout.SideBorderDecoration`
+
+    This module allows for having a configurable border position around
+    windows; i.e., it can move the border to either cardinal direction.
+
 ### Bug Fixes and Minor Changes
 
   * `XMonad.Prompt.OrgMode`
--- a/XMonad/Layout/SideBorderDecoration.hs
+++ b/XMonad/Layout/SideBorderDecoration.hs
@@ -1,0 +1,193 @@
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE InstanceSigs          #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NamedFieldPuns        #-}
+--------------------------------------------------------------------
+-- |
+-- Module      : XMonad.Layout.SideBorderDecoration
+-- Description : Configure the border position around windows.
+-- Copyright   : (c) 2018  L. S. Leary
+--                   2022  Tony Zorman
+-- License     : BSD3
+-- Maintainer  : Tony Zorman <soliditsallgood@mailbox.org>
+--
+-- This module allows for having a configurable border position around
+-- windows; i.e., it can move the border to any cardinal direction.
+--
+--------------------------------------------------------------------
+module XMonad.Layout.SideBorderDecoration (
+  -- * Usage
+  -- $usage
+  sideBorder,
+
+  -- * Border configuration
+  SideBorderConfig (..),
+  def,
+
+  -- * Re-exports
+  Direction2D (..),
+
+  -- * Lower-level hooks
+  sideBorderLayout,
+) where
+
+import qualified XMonad.StackSet as W
+
+import XMonad
+import XMonad.Layout.Decoration
+import XMonad.StackSet (Stack)
+import XMonad.Util.Types
+
+{- $usage
+
+To use this module, first import it into your configuration file:
+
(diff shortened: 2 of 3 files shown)


LSLeary:
Some quick points:
  - Since these are border replacements rather than title bars, we don't really want fonts being rendered. In the original comment, I said you can use the same font colour as the background, but that's a hack—can't we provide a `const ""` shrinker instead?
  - It would be nice to have a more convenient interface that doesn't require as much fiddling from non-Haskellers; a `sideBorder` function that just takes two border colours and a width and does the rest.
  - Those three key config values are duplicated in the standard config, so we could provide a config modifier that actually reads them from there before modifying the `borderWidth` and `layoutHook` accordingly.

slotThe:
Thanks, these are good points.  I pushed some thoughts in a80f7532018ad2fc7d94826544fc0ee3c4f37963 (docs TBD).

>     * Those three key config values are duplicated in the standard config, so we could provide a config modifier that actually reads them from there before modifying the `borderWidth` and `layoutHook` accordingly.

I was thinking about this when copying over the module, but decided against it at first.  The problem is that we can'd use ExtensibleConf here, since we have a type-changing update to `XConfig`, which doesn't play very nice with that interface.  I don't think there is a clean way to ensure that users are only applying this once to their config.  On the other hand, users are quite used to adding layout modifiers.  Maybe it doesn't matter if the documentation explicitly says to add only the combinator though; it's `sideBorder` in a80f7532018ad2fc7d94826544fc0ee3c4f37963 for reference.

slotThe:
Okay, documentation should also be done now (I hope)

LSLeary:
I still need to have a proper look, but I might not be able to do so in the next day or two, so I'll just put in my current thoughts now:
  - I'm not that familiar with XC, but it seems like something of an oversight that `once` can't touch the `layoutHook`. I've mocked up a solution of sorts: https://gist.github.com/LSLeary/4b2a45a519e8ca961fabacd35894b7d4#file-oncel-hs
    Perhaps it merits a PR of its own?
  - While it is good to have the combinator setting `borderWidth = 0`, as is, it doesn't give us much over the layout modifier. Most of my point for suggesting this was that it could become `sideBorder :: Direction2D -> XConfig l -> XConfig (SideBorder l)`, seamlessly reinterpreting the config. It's perhaps a problem that (without `onceL`) it could be applied twice and see `borderWidth = 0`, but I don't think that's such a big issue. Of course, it should admit to supporting only a single application.
  - At this point, you should probably add yourself to the copyright, since you've written most of it.
  - I see you using the word *either* in a couple of places, which suggests only two directions. Perhaps you want *any*.

slotThe:
> I still need to have a proper look, but I might not be able to do so in the next day or two, so I'll just put in my current thoughts now:
>
>   * I'm not that familiar with XC, but it seems like something of an oversight that `once` can't touch the `layoutHook`. I've mocked up a solution of sorts: https://gist.github.com/LSLeary/4b2a45a519e8ca961fabacd35894b7d4#file-oncel-hs
>     Perhaps it merits a PR of its own?

Oh, indeed.  I didn't think this through far enough to realise that all we need is an instance of `LayoutClass` for an either-like layout type!  I definitely think that this is worth a PR.

>   * While it is good to have the combinator setting `borderWidth = 0`, as is, it doesn't give us much over the layout modifier. Most of my point for suggesting this was that it could become `sideBorder :: Direction2D -> XConfig l -> XConfig (SideBorder l)`, seamlessly reinterpreting the config. It's perhaps a problem that (without `onceL`) it could be applied twice and see `borderWidth = 0`, but I don't think that's such a big issue. Of course, it should admit to supporting only a single application.

Probably not, but we've sort of slowly been transferring over to telling users to use combinators for most things that this seems like a natural thing to do.  The layout modifier is exported either way, so it shouldn't matter much.  I'll update the docs to remind users not to apply it multiple times (or in combination with the layout modifier) though; best to be explicit about things like that.

>   * I see you using the word _either_ in a couple of places, which suggests only two directions. Perhaps you want _any_.

:+1:

LSLeary (approved):
Barring the points I've already mentioned, it looks fine.

slotThe:
Thanks for the review!

PR #743, opened 2022-08-16
X.U.Loggers: Add variants of logTitle with urgent windows support

alternateved:
### Description

This is a small addition which enables to add logging for urgent windows. There might be a lot of room for improvement since those added variants looks almost exactly like `logTitles` and `logTitlesOnScreen`.

### Checklist

        manually, ...) and concluded: Tested this manually, urgent windows are logged properly.

### DIFF
--- a/XMonad/Util/Loggers.hs
+++ b/XMonad/Util/Loggers.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE MultiWayIf #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Util.Loggers
@@ -32,12 +33,15 @@ module XMonad.Util.Loggers (
 
     -- * XMonad Loggers
     -- $xmonad
-    , logCurrent, logLayout, logTitle, logTitles
+    , logCurrent, logLayout
+    , logTitle, logTitles, logTitles'
     , logConst, logDefault, (.|)
     -- * XMonad: Screen-specific Loggers
     -- $xmonad-screen
     , logCurrentOnScreen, logLayoutOnScreen
-    , logTitleOnScreen, logWhenActive, logTitlesOnScreen
+    , logTitleOnScreen, logWhenActive
+    , logTitlesOnScreen, logTitlesOnScreen'
+    , TitlesFormat(..)
     -- * Formatting Utilities
     -- $format
     , onLogger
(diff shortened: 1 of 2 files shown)


slotThe (commented):
I knew I should have used a product type instead of giving `logTitles[OnScreen]` two separate arguments :/

If nothing else, we should at least do this now.  Something like `logTitles' :: LoggerFormat -> Logger` (n.b.: a better name would be neat as well)

alternateved:
Hmm, not sure if name-wise this is any better.

XMonad/Util/Loggers.hs
```diff
@@ -203,12 +203,47 @@ logTitlesOnScreen sid formatFoc formatUnfoc = (`withScreen` sid) $ \screen -> do
...
+logTitles' (TitlesFormat' formatFoc formatUnfoc formatUrg) = do
+  sid <- gets $ W.screen . W.current . windowset
+  logTitlesOnScreen' sid (TitlesFormat' formatFoc formatUnfoc formatUrg)
+
+type FocusedFormat = (String -> String)
+type UnfocusedFormat = (String -> String)
+type UrgentFormat = (String -> String)
+
+data TitlesFormat = TitlesFormat FocusedFormat UnfocusedFormat
+data TitlesFormat' = TitlesFormat' FocusedFormat UnfocusedFormat UrgentFormat
```
slotThe, line 246: This is probably easier if we use a record and give it a `Default` instance (users know how to deal with these)

```suggestion
data TitlesFormat' = TitlesFormat'
  { focusedFormat   :: String -> String
  , unfocusedFormat :: String -> String
  , urgentFormat    :: String -> String
  }

instance Default TitlesFormat' where
  def :: TitlesFormat'
  def = TitlesFormat' { … }
```

slotThe (commented):
Sorry, I meant just using a record for the new functions, the old ones should keep the API they have (to maintain backwards compatibility, especially since `logTitles` is used in the tutorial.  Otherwise we could just merge the two together anyways).

XMonad/Util/Loggers.hs
alternateved: Right, that makes much more sense. Thanks!

alternateved:
Not sure what the current strategy is when it comes to default values, especially when it comes to visuals, but I think that this formatting proposition is generic enough.

slotThe:
> Not sure what the current strategy is, when it comes to default values, but I think that this formatting proposition is generic enough.

We could use what [PP](https://hackage.haskell.org/package/xmonad-contrib-0.17.0/docs/src/XMonad.Hooks.StatusBar.PP.html#line-154) uses, but since users will most likely override this anyways I don't think it matters much

slotThe (approved):
LGTM!  I will merge this in a bit if no one else objects

alternateved:
Great! Thanks.

slotThe:
Thank you!

PR #744, opened 2022-08-17
Use `mkAbsolutePath` in `doAppend` (XMonad.Prompt.AppendFile)

geekosaur:
### Description

Enhance `mkAbsolutePath` to support environment variables, then use it in XMonad.Prompt.AppendFile's `doAppend` so filespecs like `~/NOTES` and `$HOME/file.txt` are supported.

### Checklist

        manually, ...) and concluded: manually tested in ghci

### DIFF
--- a/XMonad/Prelude.hs
+++ b/XMonad/Prelude.hs
@@ -1,5 +1,6 @@
-{-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE LambdaCase   #-}
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 --------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Prelude
@@ -60,6 +61,8 @@ import Data.List.NonEmpty (NonEmpty ((:|)))
 import Data.Tuple (swap)
 import GHC.Stack
 import System.Directory (getHomeDirectory)
+import System.Environment (getEnv)
+import Control.Exception (SomeException, handle)
 import qualified XMonad.StackSet as W
 
 -- | Short for 'fromIntegral'.
@@ -80,7 +83,7 @@ chunksOf i xs = chunk : chunksOf i rest
 (!?) xs n | n < 0 = Nothing
           | otherwise = listToMaybe $ drop n xs
 
--- | Multivariant composition.
+-- | Multivariable composition.
 --
 -- > f .: g ≡ (f .) . g ≡ \c d -> f (g c d)
 (.:) :: (a -> b) -> (c -> d -> a) -> c -> d -> b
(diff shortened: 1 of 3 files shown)


geekosaur:
1. Should I add this to `CHANGES`?
2. Should I split this in two by the two commits?

XMonad/Prelude.hs
```diff
@@ -108,14 +111,19 @@ safeGetWindowAttributes w = withDisplay $ \dpy -> io . alloca $ \p ->
...
-  home <- liftIO getHomeDirectory
-  pure $ case ps of
-    '/'       : _ ->                ps
-    '~' : '/' : _ -> home <> drop 1 ps
-    _             -> home <> ('/' : ps)
+  home <- io getHomeDirectory
+  case ps of
+    '/'       : _ -> pure ps
+    '~' : '/' : _ -> pure (home <> drop 1 ps)
+    '$'       : _ -> let (v,ps') = span (`elem` ("_"<>['A'..'Z']<>['a'..'z']<>['0'..'9'])) (drop 1 ps)
```
slotThe, line 124: I wonder whether we should go through the trouble of expanding all env vars (even not at the start of the string), or if that use-case is too niche

slotThe (commented):
>  1. Should I add this to CHANGES?

Yes please (both changes, I would say)

>  2. Should I split this in two by the two commits?

I think it works as two commits, as it's really to different things we're doing here, but if you'd prefer one commit that's fine as well

XMonad/Prelude.hs
slotThe: Something like (using `X.U.Parser`)

``` haskell
pFilePath :: IO (Parser String)
pFilePath = do
  env <- Map.fromList <$> getEnvironment
  let go :: Parser [String]
      go = many1 $ get >>= \case
        '$' -> pEnvVar
        s   -> (s :) <$> munch1 (/= '$')

      pEnvVar :: Parser String
      pEnvVar = do
        s <- choice [varName, string "{" *> varName <* string "}"]
        maybe empty pure (env Map.!? s)
       where
        varName    = munch1 (`elem` validChars)
        validChars = "_" <> ['1' .. '9'] <> ['A' .. 'Z'] <> ['a' .. 'z']
  pure $ fmap concat go
```

might work, but it might also be overkill.

geekosaur:
I actually meant making a separate PR for each commit, as each addresses a different issue.

slotThe:
Oh right.  I don't think that's necessary; I think we tend to think in commits more than PRs anyways

XMonad/Prelude.hs
geekosaur: I've been thinking about this (and expanding `~` support to include `~username`), and currently think we can add that later if someone requests it. Adding a full parser for envars, and in particular (possibly repeatedly) sucking the full list into a Map when probably only one will be used, seems massive overkill.

slotThe: Yeah that's fair enough

geekosaur:
That most recent push is just a rebase.

Issue #645, opened 2021-11-09
Make X.D.Extending refer to X.H.StatusBar


slotThe:
### Description

Currently, the [logHook section of X.D.Extending](https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Doc/Extending.hs#L676) talks about X.H.DynamicLog quite extensively.  Since at this point we heavily encourage users to migrate over to the X.H.StatusBar interface, it would be nice to rewrite this section to use that API instead.

Additionally, in accordance with https://github.com/xmonad/xmonad-contrib/pull/637, it would be neat to have these docs be a little bit more "advanced"—instead of telling users how to _use_ X.H.StatusBar[^1] perhaps discuss a small part of the implementation instead.

[^1]: This should already be taken care of by the documentation itself, as well as the [tutorial](https://xmonad.org/TUTORIAL.html).

PR #751, opened 2022-09-06
X.D.Extending: Round Two

slotThe:
### Description

Closes: https://github.com/xmonad/xmonad-contrib/issues/645

The second round of updates to X.D.Extending, so it's not so old and crufty anymore.  Still not completely done with the module, but getting there!

##### X.D.Extending: Add links to headings

This is more ergonomic to use than the old-style headings; plus, the
generated Haddock is easier to read than just stating
"XMonad.Doc.Extending".

##### X.D.Extending: Update hooks

+ Refer to X.H.StatusBar whenever possible at the start.
+ Fix some small typos throughout the module.

##### X.D.Extending: Rewrite keyDel

Nowadays, removing keys is mostly done via removeKeys and removeKeysP,
so the documentation should reflect that fact.  This part is less
step-by-step instructive than the section about adding keys, but read in
succession it presents a way to deepen the newly acquired knowledge.

##### X.D.Extending: Remove keyAddDel and logHook

+ $keyAddDel can just be a side note in $keyDel, it does not need its
  own section.

+ $logHook is covered in more detail in the tutorial and does not serve
  a real purpose anymore.  One could rewrite it to be more in-depth
  about the inner workings of X.H.StatusBar, but for the time removing
  it seems like the best option.

### Checklist

        manually, ...) and concluded: it's docs

    Don't think this is necessary (at least not until this has seen a _full_ rewrite)

### DIFF
--- a/XMonad/Doc/Extending.hs
+++ b/XMonad/Doc/Extending.hs
@@ -77,13 +77,13 @@ module XMonad.Doc.Extending
     -- ** Editing mouse bindings
     -- $mouse
 
-    -- ** Editing the layout hook
+    -- ** Editing the layout hook #LayoutHook#
     -- $layoutHook
 
-    -- ** Editing the manage hook
+    -- ** Editing the manage hook #ManageHook#
     -- $manageHook
 
-    -- ** The log hook and external status bars
+    -- ** The log hook and external status bars #LogHook#
     -- $logHook
     ) where
 
@@ -187,13 +187,13 @@ interface for writing layout modifiers is implemented in
 "XMonad.Layout.LayoutModifier".
 
 For more information on using those modules for customizing your
-'XMonad.Core.layoutHook' see "XMonad.Doc.Extending#Editing_the_layout_hook".
+'XMonad.Core.layoutHook' see [Editing the layout hook](#g:LayoutHook).
 
 -}
 
 {- $prompts
 
-In the @XMonad.Prompt@ name space you can find modules providing
+In the @XMonad.Prompt@ namespace you can find modules providing
 graphical prompts for getting user input and using it to perform
 various actions.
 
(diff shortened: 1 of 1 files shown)


XMonad/Doc/Extending.hs
```diff
@@ -177,7 +170,7 @@ In the @XMonad.Layout@ namespace you can find modules exporting
 contributed layout algorithms, such as a tabbed layout, a circle, a spiral,
 three columns, and so on.
 
-You will also find modules which provide facilities for combining
```
geekosaur, line 180: Seems to me that "which" and "that" are equally applicable here?

XMonad/Doc/Extending.hs
```diff
@@ -222,7 +215,7 @@ external utilities.
 {- $extending
 #Extending_xmonad#
 
-Since the @xmonad.hs@ file is just another Haskell module, you may
+Since the @xmonad.hs@ file is just another Haskell file, you may
```
geekosaur, line 218: I'd say "is just a Haskell program".

XMonad/Doc/Extending.hs
```diff
@@ -388,83 +381,74 @@ extend xmonad an scratch a particular itch!
...
-'XMonad.Util.EZConfig.removeKeys'.
+> main :: IO ()
+> main = xmonad $ def
+>   { … }
+>  `removeKeysP` ["M-q", "M-S-q"]
 
--}
+Can you guess how 'XMonad.Util.EZConfig.removeKeysP' works?  It's almost
+the same code we wrote above, just accepting a list of keybindings.  Try
+if you can come up with an implementation of
```
geekosaur, line 425: "Try to see if…".

geekosaur (commented):
Most of these are nits, but that last one just seems wrong.

XMonad/Doc/Extending.hs
slotThe: Indeed; I remember my English teacher telling me not to use which in integrated relative clauses but apparently that was just wrong.  I'll probably change this back then to reduce diff noise

geekosaur: English teachers teach a lot of guff derived from trying to turn English into Latin.

PR #754, opened 2022-09-09
made borderresize take a parameter

Quoteme:
Until now it is not possible to change how many pixels near a border is considered "near enough", so that the window can be resized by dragging this border. This is especially frustrating on a touchscreen, where I would wish to have more than just 2 pixels to tap on for resizing windows.
This shoudl allow exactly that. Until now, you were stuck with 2 pixel wide borders for resizing. But now this value is not a global variable but instead a parameter for the borderResize layout modifier

### DIFF
--- a/XMonad/Layout/BorderResize.hs
+++ b/XMonad/Layout/BorderResize.hs
@@ -24,6 +24,7 @@ module XMonad.Layout.BorderResize
     ( -- * Usage
       -- $usage
       borderResize
+    , borderResizeNear
     , BorderResize (..)
     , RectWithBorders, BorderInfo,
     ) where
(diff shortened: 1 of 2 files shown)


XMonad/Config/Bluetile.hs
```diff
@@ -188,7 +188,7 @@ bluetileLayoutHook = avoidStruts $ minimize $ boringWindows $
                         named "Tiled2" tiled2 |||
                         named "Fullscreen" fullscreen
         where
-            floating = floatingDeco $ maximize $ borderResize positionStoreFloat
+            floating = floatingDeco $ maximize $ (borderResize 3) positionStoreFloat
```
geekosaur, line 191: The parentheses are optional here.

XMonad/Layout/BorderResize.hs
```diff
@@ -58,27 +58,24 @@ data BorderInfo = BI { bWin :: Window,
...
 
-newtype BorderResize a = BR (M.Map Window RectWithBorders) deriving (Show, Read)
+data BorderResize a = BR Dimension (M.Map Window RectWithBorders) deriving (Show, Read)
 
-brBorderSize :: Dimension
-brBorderSize = 2
-
-borderResize :: l a -> ModifiedLayout BorderResize l a
-borderResize = ModifiedLayout (BR M.empty)
+borderResize :: Dimension -> l a -> ModifiedLayout BorderResize l a
```
geekosaur, line 63: Either this should be listed as a breaking change on `CHANGES.md`, or the variant that takes a parameter should be a separate function. In the latter case, the Haskell convention is that the variant's name ends with a tick (that is, `borderResize'`). The normal one would then be implemented as the ticked one with `brBorderSize` as the parameter.

XMonad/Layout/BorderResize.hs
```diff
@@ -112,26 +109,26 @@ handleGone wrsGone = mapM_ deleteWindow borderWins
     where
         borderWins = map bWin . concatMap snd . M.elems $ wrsGone
 
-handleAppeared :: M.Map Window Rectangle -> X (M.Map Window RectWithBorders)
-handleAppeared wrsAppeared = do
+handleAppeared :: Dimension -> M.Map Window Rectangle -> X (M.Map Window RectWithBorders)
```
geekosaur, line 112: As with the earlier one, this should probably be provided as a variant with a tick appended, and the same with the following functions.

geekosaur (changes requested):
Otherwise I don't see a problem with this.

XMonad/Layout/BorderResize.hs
slotThe: Maybe `borderResizeNear` or something like that would be a better name as well (not a fan of nondescript primes :)

XMonad/Layout/BorderResize.hs
slotThe: All of these are internal functions, so it's not necessary to have variants here

XMonad/Layout/BorderResize.hs
Quoteme: I pushed a new commit to my fork: [9ff92606d83e9421fd60a36e58ffd3eaeef35338](https://github.com/Quoteme/xmonad-contrib/commit/9ff92606d83e9421fd60a36e58ffd3eaeef35338)
This adds a variant as you suggested.

slotThe:
Thanks!

Issue #756, opened 2022-09-17
Old-style scratchpads not working any more


Maeda1:
### Problem Description

After ugrading to
xmonad 0.17.1-10
xmonad-contrib 0.17.1-9

scratchpad (old-style) is not working anymore (nothing happens). That was working when using xmonad-contrib-0.17.0-20 (and xmonad-0.17.0-20).
Anyway, I am now using named scratchpad, which works well.

### Steps to Reproduce

Use a scratchpad (not named).

### Configuration File

This is not the full file, but some relevant parts about the scratchpad configuration. If this is not enough this is the [link to the full file](https://codeberg.org/Maeda/My-Xmonad-Xmobar-Trayer-configuration-files/src/branch/master/xmonad.hs) .

```haskell
import XMonad
import XMonad.Util.Scratchpad
import qualified XMonad.StackSet as W
import XMonad.Util.EZConfig(additionalKeysP)

myKeys =
        [ ("M-w", scratchPad)
        ]
        where
         scratchPad = scratchpadSpawnActionTerminal "urxvt -pe tabbed"

myManageHook = (composeAll
   [ 
   ])
   <+> manageScratchPad

--Scratchpad config
manageScratchPad :: ManageHook
manageScratchPad = scratchpadManageHook (W.RationalRect l t w h)
  where
    h = 0.25   -- terminal height, 25%
    w = 0.98   -- terminal width, 98%
    t = 0.72   -- distance from top edge, 72%
    l = 0.01   -- distance from left edge, 1%

main = xmonad $ def
        { manageHook = myManageHook
        , modMask = mod4Mask     -- Touche Super
        } `additionalKeysP` myKeys
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17.1-10
    - [x] With `xmonad-contrib` version 0.17.1-9

turboMaCk:
Interestingly I'm using named scratchpads which also stopped working for me after upgrade. I guess there had to be some more general breaking change.

turboMaCk:
It turned out I was using manage hook from now deprecated legacy version which was the reason for breakage. Switching to `namedScratchpadManageHook` fixed the problem.

PR #757, opened 2022-09-18
X.U.Scratchpad: Deprecate

slotThe:
This is i) broken and ii) just the functionality of X.U.NamedScratchpad
rewrapped (and not necessarily improved upon) at this point.

With recent changes to the way named scratchpads work[1], we would have
to export internals of X.U.NamedScratchpad in order to restore
X.U.Scratchpad to its full functionality.  This does not seem worth it,
as the latter does not bring anything substantially new to the table.

Closes: https://github.com/xmonad/xmonad-contrib/issues/756
Related: https://github.com/xmonad/xmonad-contrib/issues/591
[1]: 3fc830aa09368dca04df24bf7ec4ac817f2de479

### Checklist

        manually, ...) and concluded: Haddock and GHC say it's deprecated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,13 @@
 ## _unreleased_
 
 ### Breaking Changes
+
+* `XMonad.Util.NamedScratchpad`:
+
+  - Deprecated the module; use `XMonad.Util.NamedScratchpad` instead.
+
 ### New Modules
+
 ### Bug Fixes and Minor Changes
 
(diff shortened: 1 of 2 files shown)


TheMC47 (approved):
Let's deprecate it :+1: especially that  there's no one maintaining it/contributing to it, and we have something else anyway.

_disclaimer: I don't use it and not that familiar with it_

PR #758, opened 2022-09-29
expose more messages in MouseResizableTile layout

alexandersokolow:
Very small and straightforward change, see the following post for context: https://www.reddit.com/r/xmonad/comments/xpzeg9/normalize_the_fraction_between_master_and_slave/

I've also included the SetSlaveFraction messages, because they could be used in a similar (albeit more complicated, as there are multiple slaves) fashion to create a function that normalizes all of the sizes (instead of just the fraction between master and slave).


geekosaur (approved):
Looks good to me

TheMC47 (approved):
LGTM

slotThe:
Thanks!

PR #759, opened 2022-10-04
X.U.Parser: Achieve feature parity with ReadP

slotThe:
#### X.U.Parser: Feature parity with ReadP

While not many of these more exotic combinators are used right now, it's still nice to have feature parity (i.e., everything that one could want from a basic parser) such that people don't have to add their own combinators, in case they want to use one that's not already implemented.

#### X.U.Parser: Inline definitions

Most of these definitions are probably small enough to be inlined on their own, but tell GHC to try really hard regardless.  This is commonly done by other parser libraries as well; e.g., [1], so it shouldn't cause any issues either way.

[1]: https://hackage.haskell.org/package/parsers-0.12.11

#### X.P.OrgMode: Remove pLast

Use the more aptly named `option` instead.

### Checklist

        manually, ...) and concluded: Pretty much all definitions are small and just what other parsing libraries do.

### DIFF
--- a/XMonad/Util/Parser.hs
+++ b/XMonad/Util/Parser.hs
@@ -135,18 +135,22 @@ instance Semigroup (Parser a) where
   -- produces any result at all, then right parser is not used.
   (<>) :: Parser a -> Parser a -> Parser a
   (<>) = coerce ((<++) @a)
+  {-# INLINE (<>) #-}
 
 instance Monoid (Parser a) where
   -- | A parser that always fails.
   mempty :: Parser a
   mempty = Parser empty
+  {-# INLINE mempty #-}
 
 instance Alternative Parser where
   empty :: Parser a
   empty = mempty
+  {-# INLINE empty #-}
 
   (<|>) :: Parser a -> Parser a -> Parser a
   (<|>) = (<>)
+  {-# INLINE (<|>) #-}
 
 -- | When @-XOverloadedStrings@ is on, treat a string @s@ as the parser
 -- @'string' s@, when appropriate.  This allows one to write things like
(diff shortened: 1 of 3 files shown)


XMonad/Util/Parser.hs
```diff
@@ -123,108 +135,222 @@ instance Semigroup (Parser a) where
...
+gather :: forall a. Parser a -> Parser (String, a)
+gather = coerce (ReadP.gather @a)
+{-# INLINE gather #-}
 
 -- | Succeeds if and only if we are at the end of input.
 eof :: Parser ()
 eof = coerce ReadP.eof
+{-# INLINE eof #-}
 
 -- | Parse an integral number number.
```
geekosaur, line 201: This isn't part of your change, but this comment looks odd. I presume one of the "number"s refers to a parameter; perhaps that needs `@@` or something.

geekosaur (approved):
Otherwise looks reasonable to me. Except for the question of, if we're going for feature parity with other parsers, do we want to instead think about using one of them? Not saying we need to make that decision now, but we should consider whether we're just pointlessly proliferating parsers.

slotThe:
> Except for the question of, if we're going for feature parity with other parsers, do we want to instead think about using one of them? Not saying we need to make that decision now, but we should consider whether we're just pointlessly proliferating parsers.

This is a good point, and I think it's worth thinking about it.  I initially wrote X.U.Parser because ReadP, while being in base and pretty good for most use-cases, has quite confusing default behaviours for a lot of combinators.  This is related to the difference between the truly commutative `(+++)`, which is used basically everywhere, and the more parsec-like, left biased, `(<++)`.  I figured that we didn't want a dependency like `megaparsec` or `attoparsec` and the module was short enough (it's still not above 400 lines) that it didn't needlessly bloat the code base.

PR #760, opened 2022-10-04
X.P.OrgMode: Add orgPromptRefile[To]

slotThe:
### Description

~~This is still a bit WIP in the sense that documentation and other fluff is missing, but it at least seems complete feature-wise.~~ This should now be ready to review.  Since someone asked me whether I could implement this, I figured it was best to open this to "public testing" as soon as possible.

##### X.U.Run: Don't use elispFun in progn

This is already taken care of by execute and eval, so forcing it in progn only hampers composability.

##### X.P.OrgMode: Add mkOrgCfg

This ensures that we always immediately expand the file path upon constructing an `OrgMode` record.  We thus do not have to do this in `mkOrgPrompt` anymore.

##### X.P.OrgMode: Return "exitCode" mkOrgPrompt

More formally, return whether the user cancelled the prompt or not. This is useful in case we want to do things only after a "successful" input.

##### X.P.OrgMode: Add orgPromptRefile[To]

Add orgPromptRefile and orgPromptRefileTo in order to refile entries after insertion.

### Checklist

        manually, ...) and concluded: tested manually; seems to work.  Probably best to write some property tests for the file parsing at some point in the future.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -92,6 +92,12 @@
     in case `lineNavigation` can't find a window.  This benefits
     especially users who use `XMonad.Layout.Spacing`.
 
+* `XMonad.Prompt.OrgMode`
+
+  - Added `orgPromptRefile` and `orgPromptRefileTo` for interactive and
+    targeted refiling of the entered note into some existing tree of
+    headings, respectively.
+
 ### Other changes
 
 ## 0.17.1 (September 3, 2022)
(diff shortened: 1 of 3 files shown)


geekosaur (approved):
I don't see anything wrong here aside from a slight misgiving over `proc` which preexisted your change (see arrow syntax).

PR #761, opened 2022-10-13
Remove keys type

exorcist365:
### Description

behind them.

### Checklist

        manually, ...) and concluded: that the documentation now displays properly

### DIFF
--- a/XMonad/Hooks/Modal.hs
+++ b/XMonad/Hooks/Modal.hs
@@ -117,7 +117,7 @@ import           XMonad.Util.Parser            ( runParser )
 --
 --   - @[(String, X ())]@, or
 --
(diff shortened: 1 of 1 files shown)


TheMC47:
You're right about the haddock docs looking bad with this type hidden... Not the brightest idea I had.

However, I don't think this type should be exposed here. If we want to expose it, either do it in XMonad.Prelude and replace the occurrences, or, better yet, maybe from core.

slotThe:
There might also be an argument for just removing this alias.  We're only using it in two places, really, so it might just unnecessarily confuse people

exorcist365:
I mean... fair.

slotThe (approved):
LGTM.  @TheMC47 what do you think?

TheMC47:
LGTM, just change the commit message to something more descriptive

slotThe:
Thanks

Issue #712, opened 2022-04-26
`XMonad.Actions.FloatKeys` Add simpler functions


TheMC47:
### Problem Description

I wanted to do something simple as "shrink/move/expand my window by x from the right/left/top/bottom". It worked, but with ~some~ a lot of trial, error, and head-scratching.

If you never contributed to xmonad: this issue is for you! It's a straightforward addition that requires minimal Haskell knowledge

### Steps to Reproduce

Read the docs.

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17
    - [x] With `xmonad-contrib` version https://github.com/xmonad/xmonad-contrib/commit/0e106ccfbee21c90eec46537bfb4341de648c9f7

Issue #709, opened 2022-04-18
`insertPosition` misfires if a dock is started


geekosaur:
### Problem Description

In the case where a dock window is started, its window will be `doIgnore`d instead of being placed in the `StackSet`. An `insertPosition` later in the `manageHook` will then use the wrong window, leading to behavior such as `insertPosition Below Newer` switching focus to the previously focused window. This probably also happens with other windows that are `doIgnore`d in the `manageHook`. It may also happen if the new window is `doShift`ed.

I think the only way to guard against this is for `insertPosition` to verify that the focused window on the current workspace is the same as that available via `ask` on the `Query`, and to refuse to operate otherwise.

### Steps to Reproduce

Launch an `xmonad` session with the configuration file below, then restart it such that the status bar also restarts. If using `xmobar`, it must be set to not `override_redirect`; the example here uses `polybar`.

### Configuration File

Please include the smallest _full_ configuration file that reproduces
the problem you are experiencing:

```haskell
module Main (main) where

import XMonad
import XMonad.Hooks.InsertPosition

main :: IO ()
main = xmonad $ docks def {
  startupHook = spawn "polybar",
  manageHook = insertPosition Below Newer <+> manageHook def
}
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17.0
    - [x] With `xmonad-contrib` version 0.17.0

geekosaur:
It looks like `insertPosition` already tries to deal with this via `viewingWs`. I'm not sure why it fails, unless it's because `docks` fires too late and therefore the window is placed only to be removed. This may be a downside of the `docks` combinator.

slotThe:
Some relevant IRC conversation:

> \<geekosaur>	the documentation for it makes it clear what ordering to use, and if you consider what it does you really do need to order it properly relative to things like doShift or doIgnore
> \<geekosaur>	it has code to deal with this, but that code is useless if the shift or ignore, etc. happens *after* the insertPosition
> \<geekosaur>	there is no way for it to retroactively clean up the Stack if you do something after it takes effect
> \<geekosaur>	this may argue for it being a combinator so it can always insert itself leftmost (remember, composition associates to the right), but docks would still have to insert itself rightmost to not comflict
> \<Solid>	mh, I wonder if that would break anything
[...]

Issue #627, opened 2021-10-21
Improve X.A.Navigation2D documentation


geekosaur:
### Problem Description

It is not clear from a cursory reading of the documentation when to use `lineNavigation` vs. `sideNavigation` vs. `centerNavigation`; you have to dig into the documentation for each strategy, and they aren't conveniently linked from the summary.

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration: N/A

PR #763, opened 2022-10-16
A Potpourri of Small Changes

slotThe:
#### 7ec2cf25 Refer to the tutorial instead of X.D.Extending more often

Essentially, whenever the tutorial actually has decent material on the
subject matter.  The replacement is roughly done as follows:

  - logHook → tutorial
  - keybindings → tutorial, as this is thoroughly covered
  - manageHook → tutorial + X.D.Extending, as the manageHook stuff the
    tutorial talks about is a little bit of an afterthought.
  - X.D.Extending (on its own) → tutorial + X.D.Extending
  - layoutHook → tutorial + X.D.Extending, as the tutorial, while
    talking about layouts, doesn't necessarily have a huge focus there.
  - mouse bindings → leave this alone, as the tutorial does not at all
    talk about them.

#### 09c47c2c X.A.FloatKeys: Add direction{Move,Resize}Window

These are simpler, more easily understood, alternatives to the existing
functions.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/712

#### 180a4aa8 X.H.InsertPosition: Add combinator

Users may not see the warning that insertPosition definitely needs to be
inserted at the leftmost position, which can cause undesired behaviour.
Having a combinator that handles this automatically seems like a sane
idea.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/709

       (Note that `docs` wasn't changed since it already inserts itself
       rightmost.)

#### 84c94d83 X.A.Navigation2D: More prominently document strategies

Also, add a note that users who use gaps or spacing may need to look
into these strategies a bit more deeply.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/627

#### f833d949 X.A.Navigation2D: Add sideNavigation as to default tiled navigation

Add sideNavigation as a fallback if needed.  This should not have any
user-facing behaviour change when not using gaps or spacing, as line
navigation is preferred.  However, users who do use spacing or gaps
should now potentially not have to change the default strategy in order
to have a usable module.

### Checklist

        manually, ...) and concluded: most of this is docs, but I've manually tested e.g. the changes to FloatKeys.

    Where appropriate.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -79,6 +79,11 @@
   - Added `directionMoveWindow` and `directionMoveWindow` as more
     alternatives to the existing functions.
 
+* `XMonad.Hooks.InsertPosition`
+
+  - Added `setupInsertPosition` as a combinator alternative to
+    `insertPosition`.
+
 ### Other changes
 
 ## 0.17.1 (September 3, 2022)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -84,6 +84,12 @@
   - Added `setupInsertPosition` as a combinator alternative to
     `insertPosition`.
 
+* `XMonad.Actions.Navigation2D`
+
+  - Added `sideNavigation` as a fallback to the default tiling strategy,
+    in case `lineNavigation` can't find a window.  This benefits
+    especially users who use `XMonad.Layout.Spacing`.
+
 ### Other changes
 
 ## 0.17.1 (September 3, 2022)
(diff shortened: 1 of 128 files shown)


TheMC47 (commented):
LGTM, I'll double-check the changes to `XMonad.Actions.FloatKeys` later/tomorrow. I can't say anything about `XMonad.Actions.Navigation2D` since I'm not familiar with it

TheMC47 (approved):

PR #765, opened 2022-10-19
Create HLint workflow

slotThe:
### Description

Of course I have no idea what I'm doing, but I tried :)

##### 26abde6e CI: Add hlint workflow

Closes: https://github.com/xmonad/xmonad-contrib/issues/669

##### 1a3f9997 CI: Remove hlint from haskell-ci

As we have a proper hlint action now, this is no longer needed.

##### 2c95e278 hlint: Fix new warning

There are some "Reducing duplication" suggestions (which are not warnings) that are still up; I left these in for now.

### Checklist

  - [n/a] I've considered how to best test these changes (property, unit,

### DIFF
--- a/.github/workflows/haskell-ci-hackage.patch
+++ b/.github/workflows/haskell-ci-hackage.patch
@@ -45,12 +45,12 @@ set in GitHub repository secrets.
            - compiler: ghc-8.10.7
              compilerKind: ghc
              compilerVersion: 8.10.7
-@@ -209,8 +217,80 @@
+@@ -240,8 +240,80 @@
            ${CABAL} -vnormal check
        - name: haddock
          run: |
--          $CABAL v2-haddock $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH all
-+          $CABAL v2-haddock $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH --haddock-for-hackage --builddir $GITHUB_WORKSPACE/haddock all
+-          $CABAL v2-haddock --haddock-all $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH all
++          $CABAL v2-haddock --haddock-all $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH --haddock-for-hackage --builddir $GITHUB_WORKSPACE/haddock all
        - name: unconstrained build
          run: |
            rm -f cabal.project.local
--- a/.github/workflows/hlint.yaml
+++ b/.github/workflows/hlint.yaml
@@ -1,0 +1,22 @@
+name: hlint
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  hlint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: 'Set up HLint'
+      uses: haskell/actions/hlint-setup@v2
+      with:
+        version: '3.4.1'
+
+    - name: 'Run HLint'
+      uses: haskell/actions/hlint-run@v2
+      with:
+        path: '["XMonad/", "tests/", "scripts/"]'
+        fail-on: status
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -8,9 +8,9 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.14.3
+# version: 0.15.20221009
 #
-# REGENDATA ("0.14.3",["github","cabal.project"])
+# REGENDATA ("0.15.20221009",["github","cabal.project"])
 #
 name: Haskell-CI
 on:
@@ -26,7 +26,7 @@ on:
 jobs:
   linux:
     name: Haskell-CI - Linux - ${{ matrix.compiler }}
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     timeout-minutes:
       60
     container:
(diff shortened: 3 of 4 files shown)


.github/workflows/hlint.yaml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  hlint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: 'Set up HLint'
+      uses: haskell/actions/hlint-setup@v2
+      with:
+        version: '3.1.6'
```
liskin, line 16: Can we set it to the version we currently use, `3.4.1`? Hopefully there won't be any unexpected warnings then, and if yes, we should investigate why.

.github/workflows/hlint.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  hlint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: 'Set up HLint'
+      uses: haskell/actions/hlint-setup@v2
+      with:
```
liskin, line 15: Previously we failed the check run whenever hlint reported anything, so we should keep that unless that's a deliberate change in behaviour.

```
fail-on: status
```

liskin (changes requested):

.github/workflows/hlint.yaml
liskin: One difference betwen the way it's run by the haskell-ci workflow and this action is that haskell-ci detects `default-language:` and passes `-XHaskell2010` to hlint, whereas the hlint-run action doesn't permit passing any command-line args at all. Not sure what difference that makes in practice. Passing `XHaskell2010` triggered a bug in hlint in the past which is fixed now, and old haskell-ci used to not pass it at all historically, and we never really cared whether it does or doesn't except for the bug it triggered, so it'll probably be fine.

.github/workflows/hlint.yaml
slotThe: It was deliberate, actually, since I think some of hlint's suggestions are a bit too opinionated, but maybe you're right and I should just suffer through and fix the new ones :)

.github/workflows/hlint.yaml
slotThe: > Can we set it to the version we currently use, `3.4.1`? Hopefully there won't be any unexpected warnings then, and if yes, we should investigate why.

Sure, this should be possible.

> One difference betwen the way it's run by the haskell-ci workflow and this action is that haskell-ci detects `default-language:` and passes `-XHaskell2010` to hlint, whereas the hlint-run action doesn't detect anything and permit passing any command-line args at all even if we needed to. Not sure what difference that makes in practice.

hlint already turns on a lot of language extensions by default; we would definitely notice if something went wrong here, as we'd have a lot of files that fail to parse.  Pretty safe to assume that this won't make any difference.

.github/workflows/hlint.yaml
liskin: Now that we're using the correct hlint version, there are no suggestions, and I suspect you could even drop 2c95e27 if you wanted to and there still wouldn't be any. That doesn't change the fact hlint could be opinionated in general, but it no longer produces suggestions that we'd rather not implement…

slotThe: Interesting that we'd get _less_ warnings with a newer hlint version (considering that the suggestions I saw were valid, just perhaps a bit overambitious)

liskin: Yeah, old me would go and find out what's going on… :-)

.github/workflows/hlint.yaml
```diff
@@ -0,0 +1,22 @@
...
+        version: '3.4.1'
+
+    - name: 'Run HLint'
+      uses: haskell/actions/hlint-run@v2
+      with:
+        path: XMonad/
```
liskin, line 21: In bd5b969d9ba24236c0d5ef521c0397390dbc4b37 you hlinted the tests as well, and haskell-ci used to hlint everything that ends up in the tarball. Is this change intentional as well?

liskin (commented):
Almost approved, just one last nit.

.github/workflows/hlint.yaml
slotThe: Ah, no, this was me being stupid :)

liskin: https://github.com/xmonad/xmonad-contrib/commit/bd5b969d9ba24236c0d5ef521c0397390dbc4b37#diff-b376ebbec70caee428a485ef2096195a5f6afe61313e005b15640d8bec912d6d

(sometimes I wonder whether this level of diligence is as annoying as it feels :-))

liskin (approved):
Anyway please feel free to merge once you (decide not to) deal with the last comment re hlinted source paths.

.github/workflows/hlint.yaml
slotThe: Nono, it's definitely appreciated :)

.github/workflows/haskell-ci-hackage.patch
```diff
@@ -45,12 +45,12 @@ set in GitHub repository secrets.
...
              compilerVersion: 8.10.7
-@@ -209,8 +217,80 @@
+@@ -240,8 +240,80 @@
            ${CABAL} -vnormal check
        - name: haddock
          run: |
--          $CABAL v2-haddock $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH all
-+          $CABAL v2-haddock $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH --haddock-for-hackage --builddir $GITHUB_WORKSPACE/haddock all
+-          $CABAL v2-haddock --haddock-all $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH all
++          $CABAL v2-haddock --haddock-all $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH --haddock-for-hackage --builddir $GITHUB_WORKSPACE/haddock all
```
liskin, line 53: This breaks the docs tarball that would be uploaded to Hackage on release :-(

slotThe: Oh.  These were just regenerated with `haskell-ci regenerate`; I guess we need another patch

liskin: Fixed by https://github.com/xmonad/xmonad-contrib/pull/775/commits/688041791176f0b6758012e35a255f9836486629, sorry for the panic :-)

PR #766, opened 2022-10-21
Factor X.A.Cycle* modules; Write Alt+Tab style window switching

LSLeary:
### Description

Note that I haven't tested these changes beyond seeing that they compile without warnings on GHC 8.8.4 and 9.2.4. I intend to do so, but I'll put it up here so others can also test or review it in the meantime, if they so wish. Commits follow:

#### Factor shared logic out of X.A.Cycle* modules

These modules were duplicating a substantial amount of low-level code.

`X.A.CycleWorkspaceByScreen` had already separated most of the implementation details from the logic with `repeatableAction`; all that was left was to generalise it a little further, put it in a suitable place and express the other modules through it.

#### Write new module: X.U.History

(for the below)

#### Write new module: X.A.MostRecentlyUsed

Alt+Tab style window switching. It can be pretty handy, and people are used to having it.

Given the existence of the other repeatable actions, it's quite a glaring omission.

### Checklist

### DIFF
--- a/XMonad/Actions/CycleWorkspaceByScreen.hs
+++ b/XMonad/Actions/CycleWorkspaceByScreen.hs
@@ -25,11 +25,10 @@ module XMonad.Actions.CycleWorkspaceByScreen (
 
 import           Data.IORef
 
-import           Graphics.X11.Xlib.Extras
-
 import           XMonad
 import           XMonad.Prelude
 import           XMonad.Hooks.WorkspaceHistory
+import           XMonad.Actions.Repeatable (repeatable)
 import qualified XMonad.StackSet as W
 
 -- $usage
@@ -53,22 +52,9 @@ import qualified XMonad.StackSet as W
 --
 -- > , ((mod4Mask, xK_slash), cycleWorkspaceOnCurrentScreen [xK_Super_L] xK_slash xK_p)
 
-repeatableAction :: [KeySym] -> (EventType -> KeySym -> X ()) -> X ()
-repeatableAction mods pressHandler = do
-  XConf {theRoot = root, display = d} <- ask
-  let getNextEvent = io $ allocaXEvent $ \p ->
-                 do
-                   maskEvent d (keyPressMask .|. keyReleaseMask) p
-                   KeyEvent {ev_event_type = t, ev_keycode = c} <- getEvent p
-                   s <- io $ keycodeToKeysym d c 0
-                   return (t, s)
-      handleEvent (t, s)
-        | t == keyRelease && s `elem` mods = return ()
-        | otherwise = pressHandler t s >> getNextEvent >>= handleEvent
-
-  io $ grabKeyboard d root False grabModeAsync grabModeAsync currentTime
-  getNextEvent >>= handleEvent
-  io $ ungrabKeyboard d currentTime
+{-# DEPRECATED repeatableAction "Use XMonad.Actions.Repeatable.repeatable" #-}
+repeatableAction :: [KeySym] -> KeySym -> (EventType -> KeySym -> X ()) -> X ()
+repeatableAction = repeatable
 
 handleKeyEvent :: EventType
                -> KeySym
@@ -109,8 +95,7 @@ cycleWorkspaceOnScreen screenId mods nextKey prevKey = workspaceHistoryTransacti
         return $ cycleWorkspaces !! current
       focusIncrement i = io (getAndIncrementWS i) >>= (windows . W.greedyView)
 
-  focusIncrement 1 -- Do the first workspace cycle
-  repeatableAction mods $
+  repeatable mods nextKey $
     runFirst
       [ handleKeyEvent keyPress nextKey $ focusIncrement 1
       , handleKeyEvent keyPress prevKey $ focusIncrement (-1)
--- a/XMonad/Actions/MostRecentlyUsed.hs
+++ b/XMonad/Actions/MostRecentlyUsed.hs
@@ -1,0 +1,227 @@
+{-# LANGUAGE NamedFieldPuns, GeneralizedNewtypeDeriving #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  XMonad.Actions.MostRecentlyUsed
+-- Description :  Tab through windows by recency of use.
+-- Copyright   :  (c) 2022 L. S. Leary
+-- License     :  BSD3-style (see LICENSE)
+--
+-- Maintainer  :  @LSLeary (on github)
+-- Stability   :  unstable
+-- Portability :  unportable
+--
+-- Based on the Alt+Tab behaviour common outside of xmonad.
+--
+-----------------------------------------------------------------------------
+
+-- --< Imports & Exports >-- {{{
+
+module XMonad.Actions.MostRecentlyUsed (
+
+  -- * Usage
+  -- $usage
+
+  -- * Interface
+  configureMRU,
+  mostRecentlyUsed,
+  withMostRecentlyUsed,
+  Location(..),
+
+  ) where
+
+-- base
+import Data.Maybe (fromMaybe)
+import Data.List.NonEmpty (NonEmpty(..), nonEmpty)
+import Data.Monoid (All(..), Any)
+import Data.Foldable (for_)
+import Data.Functor (($>))
+import Data.IORef (newIORef, readIORef, writeIORef, modifyIORef)
+import Control.Applicative (liftA2)
+import Control.Monad (when, unless, join)
+import Control.Monad.IO.Class (MonadIO)
+
+-- mtl
+import Control.Monad.Trans (lift)
+import Control.Monad.State (get, put, gets)
+
+-- containers
+import qualified Data.Map.Strict as M
+
+-- xmonad
+import XMonad
+  ( Window, KeySym, keyPress, io
+  , Event (DestroyWindowEvent, UnmapEvent, ev_send_event, ev_window)
+  )
+import XMonad.Core
+  ( X, XConfig(..), windowset, WorkspaceId, ScreenId
+  , ExtensionClass(..), StateExtension(..)
+  , waitingUnmap
+  )
+import XMonad.Operations (screenWorkspace)
+import qualified XMonad.StackSet as W
+
+-- xmonad-contrib
+import qualified XMonad.Util.ExtensibleConf  as XC
+import qualified XMonad.Util.ExtensibleState as XS
+import XMonad.Util.PureX
+  (handlingRefresh, curScreenId, curTag, greedyView, view, peek, focusWindow)
+import XMonad.Util.History (History, origin, event, erase, ledger)
+import XMonad.Actions.Repeatable (repeatableSt)
+
+-- }}}
+
+-- --< Core Data Types: WindowHistory & Location >-- {{{
+
+data WindowHistory = WinHist
+  { busy :: !Bool
+  , hist :: !(History Window Location)
+  } deriving (Show, Read)
+
+instance ExtensionClass WindowHistory where
+  initialValue = WinHist
+    { busy = False
+    , hist = origin
+    }
+  extensionType = PersistentExtension
+
+data Location = Location
+  { workspace :: !WorkspaceId
+  , screen    :: !ScreenId
+  } deriving (Show, Read, Eq, Ord)
+
+-- }}}
+
+-- --< Interface >-- {{{
+
+-- $usage
+--
+-- 'configureMRU' must be applied to your config in order for 'mostRecentlyUsed'
+-- to work.
+--
+-- > main :: IO ()
+-- > main = xmonad . configureMRU . ... $ def
+-- >   { ...
+-- >   }
+--
+-- Once that's done, it can be used normally in keybinds:
+--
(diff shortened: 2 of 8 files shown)


geekosaur (commented):
Looks reasonable to me, although I haven't tested it.

XMonad/Actions/MostRecentlyUsed.hs
```diff
@@ -0,0 +1,180 @@
...
+-- > , ((mod1Mask, xK_Tab), mostRecentlyUsed [xK_Alt_L, xK_Alt_R] xK_Tab)
+--
+-- N.B.: This example assumes that 'mod1Mask' corresponds to alt, which is not
+-- always the case, depending on how your system is configured.
+
+-- | Configure xmonad to support 'mostRecentlyUsed'.
+configureMRU :: XConfig l -> XConfig l
+configureMRU = XC.once f (MRU ()) where
+  f cnf = cnf
+    { startupHook     = startupHook     cnf <> logWinHist
```
slotThe, line 106: Why is this in the startupHook?  Presumably, before opening any window (i.e., before the logHook could run at least once) pressing the respective keybindings would just make `for_ (nonEmpty $ ledger hist) $ \ne -> …` return `mempty`, so that should still work

slotThe (commented):
I will join the "this looks nice, but I haven't tested it" crowd :)

I'll try to do so this weekend; the general idea and execution look fine to me though!

XMonad/Actions/CycleWorkspaceByScreen.hs
```diff
@@ -53,22 +52,9 @@ import qualified XMonad.StackSet as W
 --
 -- > , ((mod4Mask, xK_slash), cycleWorkspaceOnCurrentScreen [xK_Super_L] xK_slash xK_p)
 
+{-# DEPRECATED repeatableAction "Use XMonad.Actions.Repeatable.repeatable" #-}
```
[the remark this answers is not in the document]
LSLeary: Good catch—it is done.

XMonad/Util/History.hs
```diff
@@ -0,0 +1,71 @@
...
+import Data.Map (Map)
+import qualified Data.Map.Strict as M
+
+-- | A history of unique @k@-events with @a@-annotations.
+data History k a  = History
+  { annals   :: !(IntMap (k, a))
+  , recorded :: !(Map k Int)
+  } deriving (Show, Read, Eq, Ord, Functor, Foldable, Traversable)
+
+-- | O(1). A history of nothing.
```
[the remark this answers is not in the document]
LSLeary: Fair; I'll try the latter for now. The former is a bit of an eyesore for me, since it doesn't render in my local haddocks.

XMonad/Util/History.hs
```diff
@@ -0,0 +1,71 @@
...
+import Data.IntMap (IntMap)
+import qualified Data.IntMap.Strict as I
+import Data.Map (Map)
+import qualified Data.Map.Strict as M
+
+-- | A history of unique @k@-events with @a@-annotations.
+data History k a  = History
+  { annals   :: !(IntMap (k, a))
+  , recorded :: !(Map k Int)
+  } deriving (Show, Read, Eq, Ord, Functor, Foldable, Traversable)
```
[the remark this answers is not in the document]
LSLeary: You're right; the documentation for the module is pretty lacking. I've gone and written some, so hopefully it's clearer now.

XMonad/Actions/MostRecentlyUsed.hs
LSLeary: I put it there without much thought, but it should indeed be quite unnecessary; I've removed it.

XMonad/Actions/MostRecentlyUsed.hs
```diff
@@ -0,0 +1,179 @@
...
+
+    for_ (nonEmpty $ ledger hist) $ \ne -> do
+      mfw <- gets (W.peek . windowset)
+      let iSt = case cycleS ne of
+            (w, _) :~ s | mfw == Just w -> s
+            s                           -> s
+      repeatableSt iSt mods $ \t s ->
+        when (t == keyPress && s == tab) $ do
+          (win, wksp) <- pop
+          lift . windows $ W.focusWindow win . W.greedyView wksp
```
slotThe, line 132: There is an initial "delay" of one key press here when one first starts cycling.  Suppose we have

``` haskell
    ("M-<Tab>", mostRecentlyUsed [xK_Super_L, xK_Super_R] xK_Tab)
```

When `M-<Tab>` is pressed, we essentially come to this event loop and stop, listening for another keypress (e.g. another `M-<Tab>`, to start cycling).  However, intuitively the first `M-<Tab>` should already start the cycling, not requiring another confirmation.

slotThe (approved):
I tried this out now; it works!  Thank you for this, I think it closes an important gap in functionality we didn't know we had.

One thing I will note is that right now we pretty much completely ignore multi-monitor setups (as we don't preserve that kind of information).  When cycling, all workspaces show up on the monitor that currently has focus, not necessarily on the monitor they originally were on.  I don't think it's essential to support this, but we should probably mention this somewhere in the documentation.

XMonad/Actions/MostRecentlyUsed.hs
LSLeary: You're right; all the other repeatable actions do some special handling for the first keypress ... but this is actually a weakness in `repeatableAction`, which I cribbed without thinking too hard about. Rather than opening with `getNextEvent >>= handleEvent`, it should do `handleEvent (keyPress, key)`, then the special handling can be erased.

This requires passing an additional argument to the `repeatable*` functions, but it's a value that's on hand, since it's needed to write the keypress handler anyway. The only problem is in `repeatableAction` itself, which now requires a change of type to be written in terms of `repeatable`. I wonder if it's worth temporarily undeleting the original implementation?

LSLeary:
@slotThe And thank you for reviewing—@geekosaur too.

Re the multi-monitor thing, I'd thought about allowing the user to customise out `greedyView` for `view`, but I hadn't considered saving the screen in the history. It's a good idea, and it's not a difficult change, so I'll try it.

LSLeary:
Due to the need to undo select changes and not just discard them all, it was a bit harder to write than I expected. Nevertheless, I like how it turned out. It might even work!

XMonad/Actions/MostRecentlyUsed.hs
```diff
@@ -116,7 +128,43 @@ mostRecentlyUsed
...
+  where
+    undoer :: (MonadIO m, Monoid a) => m (m a, m a -> m ())
+    undoer = do
+      ref <- io . newIORef $ pure mempty
+      let
+        undo = join (io $ readIORef ref) <* io (writeIORef ref $ pure mempty)
+        toUndo act = io $ modifyIORef ref (liftA2 (<>) act)
+      pure (undo, toUndo)
+    viewScreen :: ScreenId -> X Any
+    viewScreen scr = screenWorkspace scr >>= foldMap view
```
slotThe, line 155: Stupid question: do we actually need to do this here?  I don't think this is quite the same situation as https://github.com/xmonad/xmonad-contrib/pull/543.  We should already be doing the "correct thing" here; that is, not generating windowsets but operating on the current windowset instead, as well as (and this is important) declaring ourselves busy and not inserting new windows into the history in the logHook when cycling.  All state should be preseved in this way, with no ill effects.

XMonad/Actions/MostRecentlyUsed.hs
```diff
@@ -78,6 +85,11 @@ instance ExtensionClass WindowHistory where
     }
   extensionType = PersistentExtension
 
+data Location = Location
+  { workspace :: WorkspaceId
+  , screen    :: ScreenId
```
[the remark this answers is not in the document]
LSLeary: The reason is that I forgot! This kind of history has potential for space-leaks, so I'm trying to keep it strict where I can. The strict `IntMap` & `Map` are probably enough already, but there's no reason to half-arse it.

XMonad/Actions/MostRecentlyUsed.hs
LSLeary: Not really a stupid question; this is more a matter of UX than necessity. I'm considering the intermediate actions to be *previews* while the final action is the *selection*—and a preview that isn't selected should be erased as if it never happened.

I didn't think about this too deeply in the previous draft because I figured `greedyView` would take care of it, but that's not quite right either. With the screen tracking version, it's more obvious that a series of unmitigated previews will leave unrelated screens focused on historic windows we only wanted to tab past. 

What I'd really like to write here is much simpler:
```haskell
mostRecentlyUsed mods key = withWindowSet $ \ws ->
  withMostRecentlyUsed mods key $ \win Location{workspace,screen} ->
    handlingRefresh . void $ do
      modifyWindowSet (const ws)
      screenWorkspace screen >>= foldMap view
      greedyView workspace
      focusWindow win
```
But there, I gather, we really would run into the same problem as #543 when we discard the layout updates performed in each refresh. Hence, I had to do something more complex: build and discharge an "inverse" alongside each preview.

slotThe: > Not really a stupid question; this is more a matter of UX than necessity. I'm considering the intermediate actions to be _previews_ while the final action is the _selection_—and a preview that isn't selected should be erased as if it never happened.
>
> I didn't think about this too deeply in the previous draft because I figured `greedyView` would take care of it, but that's not quite right either. With the screen tracking version, it's more obvious that a series of unmitigated previews will leave unrelated screens focused on historic windows we only wanted to tab past.

Ah, I see; that would explain the use of `PureX` as well.  It's true that this is nicer conceptually.

> What I'd really like to write here is much simpler:
>
> ```haskell
> mostRecentlyUsed mods key = withWindowSet $ \ws ->
>   withMostRecentlyUsed mods key $ \win Location{workspace,screen} ->
>     handlingRefresh . void $ do
>       modifyWindowSet (const ws)
>       screenWorkspace screen >>= foldMap view
>       greedyView workspace
>       focusWindow win
> ```
>
> But there, I gather, we really would run into the same problem as #543 when we discard the layout updates performed in each refresh. Hence, I had to do something more complex: build and discharge an "inverse" alongside each preview.

Isn't it more like we queue up the state mutations, rather than discard them?  If I change the definition of `mostRecentlyUsed` to something like[^1]

``` haskell
mostRecentlyUsed mods key =
  withMostRecentlyUsed mods key $ \win Location{workspace,screen} ->
    handlingRefresh . void $ do
      screenWorkspace screen >>= foldMap view
      greedyView workspace
      focusWindow win
```

then for example X.L.ShowWName (the layout that gave us headaches in #543) seems to work correctly (i.e., correctly update its state).  While it's true that the small popup window that it spawns stays open while cycling windows on the same workspace, it correctly vanishes when either switching workspaces, or once one lets go of the relevant keybinding.

[^1]: I deleted `modifyWindowSet (const ws)` because I'm not quite sure why this is needed here.  Admittedly I'm not well-versed with `PureX`.

LSLeary:
I was rebasing to erase conflicts so I went ahead and squashed the accumulated commits—this doesn't mean it's ready to merge; it still needs testing.

There was also one new commit, but the only actual changes were to make `Location` strict, improve the order and styling in `undoer`, and tweak the docs of `withMostRecentlyUsed` for accuracy.

XMonad/Actions/MostRecentlyUsed.hs
LSLeary: We're not quite on the same page here; let me rewrite that without `PureX`, which I've actually only used because it's *cleaner*. Though I'll warn you, this one isn't type checked:

```haskell
mostRecentlyUsed mods key = withWindowSet $ \ws ->
  withMostRecentlyUsed mods key $ \win Location{workspace,screen} -> windows
    $ W.focusWindow win 
    . W.greedyView workspace
    . maybe id W.view (W.lookupWorkspace screen)
    . const ws
```

The `modifyWindowSet (const ws)`, now `. const ws`, is precisely where we discard prior changes—both unwanted previews and wanted layout updates—it's the key to the simplified but bug-inducing version. I certainly expect the version without it to work without bugs, but also without the UX I talked about.

slotThe: Sorry, I still don't quite get it; maybe I'm not seeing the whole picture here.  To me, `withWindowSet $ \ws -> windows (const ws)` is equivalent to `windows id`; it certainly shouldn't discard any kind of state.  What am I missing?

LSLeary: The windowset `ws` is grabbed only once at the very beginning, while the function `\win Location{...} -> windows $ ... . const ws`, runs repeatedly, once for each time `key` is tapped. The first time it runs, it will indeed be a no-op. The second time, however, it will erase the changes left behind by the first invocation, and so on.

slotThe: Aaaah, right, that make sense.

XMonad/Actions/MostRecentlyUsed.hs
slotThe: Tbh, I consider `repeatableAction` to be pretty "internal": I doubt many users are actually using this and not just a wrapper that'll remain unaffected. 

@LSLeary this is pretty much done, right?  No further complains from my side, so whenever you're ready you can rebase and merge this

slotThe:
Just rebasted on top of `master` and tested everything again—works!

Issue #456, opened 2021-01-30
Previously floating fullscreen windows tiles after leaving fullscreen


toniz4:
### Problem Description
When opening a floating window, entering fullscreen with fullscreenEventHook and leaving fullscreen the window tiles instead of floating again.

### Configuration File

Please include the smallest configuration file that reproduces the
problem you are experiencing:

```haskell
module Main (main) where

import XMonad
import XMonad.Hooks.EwmhDesktops
import qualified XMonad.StackSet as W
import qualified Data.Map        as M

main :: IO ()
main = xmonad def {
  handleEventHook = fullscreenEventHook,
  mouseBindings   = myMouseBindings
  }
  

myMouseBindings :: XConfig l -> M.Map (KeyMask, Button) (Window -> X())
myMouseBindings (XConfig {XMonad.modMask = modm}) = M.fromList $

    [ ((modm, button1), (\w -> focus w >> mouseMoveWindow w
                                       >> windows W.shiftMaster))

    , ((modm, button2), (\w -> focus w >> windows W.shiftMaster))

    , ((modm, button3), (\w -> focus w >> mouseResizeWindow w
                                       >> windows W.shiftMaster))

    ]
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

slotThe:
This is due to [this line](https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Hooks/EwmhDesktops.hs#L331) simply sinking the window when fullscreen is removed; I guess the stackset could have changed in the meantime and for tiled windows trying to restore some sort of "previous position" is practically impossible.  

I wonder if it's worth making an exception for floating windows here; what do other people think?

liskin:
> simply sinking the window
> trying to restore some sort of "previous position" is practically impossible

If the stackset doesn't change, then "simply sinking the window" does restore its original position. When I fullscreen a tiled window and then unfullscreen it, everything is back to where it was. If the window is moved in between, then the stackset is changed and when it sinks, it sinks into the master position or something, but I'd argue that stackset changing during fullscreen isn't all that likely.

But floats… everything related to floats in xmonad is quite a hack, and fullscreening/maximization is a hack on top of a hack. :-(
[...]

Issue #394, opened 2020-10-27
Toggle fullscreen feature


R4stafa:
I use XMonad.Layout.Fullscreen for fullscreen support and the AddFullscreen and RemoveFullscreen messages to manually make windows fullscreen but there is no easy way (at least I can't see one) to toggle the fullscreen state of a window say with a keybinding. So I went ahead and built my own xmonad-contrib where I added a ToggleFullscreen message and it works. I thought this feature should be in the official xmonad-contrib but I wanted to discuss this first before making a PR.

Sorry if I do something the the wrong way I am new to contributing to open source projects.

elkowar:
the whole point of contrib is that it's easy to PR to! So if you already have the code, there is no harm in simply doing the PR, and seeing what happens!

R4stafa:
Ok I created the PR #395

slotThe:
At this point this is basically a duplicate of https://github.com/xmonad/xmonad-contrib/issues/456.  This will be added once https://github.com/xmonad/xmonad-contrib/pull/625 is merged

PR #768, opened 2022-10-23
Add (un)fullscreen hooks and float-restoring toggleFullFloat action

liskin:
#### [X.H.EwmhDesktops: Add (un)fullscreen hooks](../commit/07e82294f073050cf319afd27ce38f1efde1b500)

Fairly straightforward, just add two hooks for (un)fullscreening. There are multiple motivations for this:

* Users are calling for unfullscreened windows to revert back to their original location if they were floating.

* XMonad.Layout.Fullscreen uses some deprecated exports from XMonad.Hooks.EwmhDesktops and reimplements fullscreenEventHook.

This commit only adds the hooks. Neither of the motivations are dealt with yet, so the docs are a bit terse still.

#### [CHANGES: Document X.A.ToggleFullFloat and setEwmhFullscreenHooks](../commit/d19450b6549fd372c521db99622d9bef9ccf7acc)

---

Related: https://github.com/xmonad/xmonad-contrib/issues/456
Related: https://github.com/xmonad/xmonad-contrib/issues/394
Related: https://github.com/xmonad/xmonad-contrib/pull/626

---

The second motivation is still TODO: X.L.Fullscreen should be refactored to use this (probably in a separate PR because there's some legacy code I can't be arsed to actually try to use anytime soon).

### Checklist

  - [X] I've considered how to best test these changes (property, unit, manually, ...) and concluded: I've been running this for a long while.

### DIFF
--- a/XMonad/Actions/ToggleFullFloat.hs
+++ b/XMonad/Actions/ToggleFullFloat.hs
@@ -1,0 +1,122 @@
+-- |
+-- Module      :  XMonad.Actions.ToggleFullFloat
+-- Description :  Fullscreen (float) a window while remembering its original state.
+-- Copyright   :  (c) 2022 Tomáš Janoušek <tomi@nomi.cz>
+-- License     :  BSD3
+-- Maintainer  :  Tomáš Janoušek <tomi@nomi.cz>
+--
+module XMonad.Actions.ToggleFullFloat (
+    -- * Usage
+    -- $usage
+    toggleFullFloatEwmhFullscreen,
+    toggleFullFloat,
+    fullFloat,
+    unFullFloat,
+    gcToggleFullFloat,
+    ) where
+
+import qualified Data.Map.Strict as M
+
+import XMonad
+import XMonad.Prelude
+import XMonad.Hooks.EwmhDesktops (setEwmhFullscreenHooks)
+import XMonad.Hooks.ManageHelpers
+import qualified XMonad.StackSet as W
+import qualified XMonad.Util.ExtensibleState as XS
+
+-- ---------------------------------------------------------------------
+-- $usage
+--
+-- The main use-case is to make 'ewmhFullscreen' (re)store the size and
+-- position of floating windows instead of just unconditionally sinking them
+-- into the floating layer. To enable this, you'll need this in your
+-- @xmonad.hs@:
+--
+-- > import XMonad
+-- > import XMonad.Actions.ToggleFullFloat
+-- > import XMonad.Hooks.EwmhDesktops
+-- >
+-- > main = xmonad $ … . toggleFullFloatEwmhFullscreen . ewmhFullscreen . ewmh . … $ def{…}
+--
+-- Additionally, this "smart" fullscreening can be bound to a key and invoked
+-- manually whenever one needs a larger window temporarily:
+--
+-- >   , ((modMask .|. shiftMask, xK_t), withFocused toggleFullFloat)
+
+newtype ToggleFullFloat = ToggleFullFloat{ fromToggleFullFloat :: M.Map Window (Maybe W.RationalRect) }
+    deriving (Show, Read)
+
+instance ExtensionClass ToggleFullFloat where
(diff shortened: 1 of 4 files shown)


PRESFIL:
Interesting. We wait `XMonad.Actions.ToggleFullFloat` to restore floating
states of windows?

I remember `X.L.PositionStoreFloat` and `X.H.PositionStoreHook` which
are both uses `X.U.PositionStore`. Does `ToggleFullFloat` also uses it?

liskin:
> We wait `XMonad.Actions.ToggleFullFloat` to restore floating states of windows?

Yeah, I'll get it finished and pushed here soon-ish.

> I remember `X.L.PositionStoreFloat` and `X.H.PositionStoreHook` which are both uses `X.U.PositionStore`. Does `ToggleFullFloat` also uses it?

I'm aware of that one but didn't use it as a storage backend to avoid conflicts. Although I must admit I can't think of a specific example of how that conflict would actually behave in practice, off the top of my head.

(It's a bit unfortunate that this PositionStore* stuff is split into three modules. It kinda looks like it may be reused by other modules, but X.U.PositionStore is just an implementation detail of X.L.PositionStoreFloat.)

liskin:
soon-ish™ :-)

Anyway, I've done the docs today and pushed it for review. Been running this for months with no trouble. Sorry for the wait.

XMonad/Actions/ToggleFullFloat.hs
```diff
@@ -0,0 +1,122 @@
...
+import XMonad.Hooks.EwmhDesktops (setEwmhFullscreenHooks)
+import XMonad.Hooks.ManageHelpers
+import qualified XMonad.StackSet as W
+import qualified XMonad.Util.ExtensibleState as XS
+
+-- ---------------------------------------------------------------------
+-- $usage
+--
+-- The main use-case is to make 'ewmhFullscreen' (re)store the size and
+-- position of floating windows instead of just uncontitionally sinking them
```
slotThe, line 31:
``` suggestion
-- position of floating windows instead of just unconditionally sinking them
```

XMonad/Actions/ToggleFullFloat.hs
```diff
@@ -0,0 +1,122 @@
...
+-- into the floating layer. To enable this, you'll need this in your
+-- @~\/.xmonad\/xmonad.hs@:
```
slotThe, line 33: I know that this is almost canonical at this point, but with the advent of proper XDG support perhaps we could just speak of an `@xmonad.hs@` here?

XMonad/Actions/ToggleFullFloat.hs
```diff
@@ -0,0 +1,122 @@
...
+        Just Nothing -> W.sink w      -- was tiled before
+        Nothing -> W.sink w           -- fallback when not found in ToggleFullFloat
+
+-- | Install ToggleFullFloat garbage collection hooks.
+--
+-- Note: This is included in 'toggleFullFloatEwmhFullscreen', only needed if
+-- using the 'toggleFullFloat' separately from the EWMH hook.
+gcToggleFullFloat :: XConfig a -> XConfig a
+gcToggleFullFloat c = c { startupHook     = startupHook c <> gcToggleFullFloatStartupHook
+                    , handleEventHook = handleEventHook c <> gcToggleFullFloatEventHook }
```
slotThe, line 99: This seems like an indentation misake?

slotThe (approved):
This look fine to me; thank you for getting to it :tada:

XMonad/Actions/ToggleFullFloat.hs
liskin: Oh yes, indeed! Changed here. Feel free to push a change across the entire repo if/when you feel like it. :-)

XMonad/Actions/ToggleFullFloat.hs
liskin: Yeah, that's what you get if you `sed` and forget that we're formatting stuff manually here :-)

Issue #771, opened 2022-10-29
The currently focused window is not last in `_NET_CLIENT_LIST_STACKING` 


ilya-bobyr:
### Problem Description

[Freedesktop spec on `_NET_CLIENT_LIST_STACKING`](https://specifications.freedesktop.org/wm-spec/latest/ar01s03.html#idm45374033259392) says:

> _NET_CLIENT_LIST_STACKING has bottom-to-top stacking order.

But, currently, `_NET_CLIENT_LIST_STACKING` is using workspace order for windows, only ordering windows within a workspace subgroup in the focus order.  Though, even that ordering is not completely bottom-to-top.

I have an application that uses `_NET_CLIENT_LIST_STACKING` in order to determine the currently focused window.
Yes, it should have used `_NET_ACTIVE_WINDOW`, but it is trying to work around some kind of bug.

### Steps to Reproduce

Switching the currently active window in a workspace that is not the very last one, I see this:
```fish
❯ while true; xprop -root _NET_CLIENT_LIST_STACKING ; sleep 1s; end
```
```
_NET_CLIENT_LIST_STACKING(WINDOW): window id # 0x4000003, 0x2a00006, 0x2c00008, 0x4e00002, 0x260002e, 0x1a00002, 0x260005a, 0x5a00002, 0x2600062, 0x260004a, 0x2600042, 0x2600052
_NET_CLIENT_LIST_STACKING(WINDOW): window id # 0x4000003, 0x2a00006, 0x2c00008, 0x4e00002, 0x260002e, 0x5a00002, 0x1a00002, 0x260005a, 0x2600062, 0x260004a, 0x2600042, 0x2600052
```

Notice how the change happens in the middle of the list.  Rather than in the last element.

### Configuration File

Please include the smallest _full_ configuration file that reproduces
the problem you are experiencing:

```haskell
module Main (main) where

import XMonad
import XMonad.Hooks.EwmhDesktops (ewmh, ewmhFullscreen)

main :: IO ()
main =
  xmonad
    $ ewmhFullscreen
    $ ewmh
    def
```

### Checklist
  - [X] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)
  - I tested my configuration
    - [X] With `xmonad` commit `a2259bb3091429f4293d371e13fd5a2ee8da30db` Sun Sep 18 16:27:14 2022 +0200
[...]

PR #772, opened 2022-10-29
EwmhDesktops: _NET_CLIENT_LIST_STACKING: In focus order

ilya-bobyr:
Order workspaces based on the visibility, before collecting windows. `_NET_CLIENT_LIST_STACKING` is supposed to be in the focus order.

### Description

A fix for https://github.com/xmonad/xmonad-contrib/issues/771.  Issue contains a description of the problem.

### Checklist
  - [X] I've considered how to best test these changes (property, unit, manually, ...) and concluded: manually

### DIFF
--- a/XMonad/Hooks/EwmhDesktops.hs
+++ b/XMonad/Hooks/EwmhDesktops.hs
@@ -320,10 +320,33 @@ ewmhDesktopsLogHook' EwmhDesktopsConfig{workspaceSort, workspaceRename} = withWi
     let clientList = nub . concatMap (W.integrate' . W.stack) $ ws
     whenChanged (ClientList clientList) $ setClientList clientList
 
-    -- Set stacking client list which should have bottom-to-top
-    -- stacking order, i.e. focused window should be last
-    let clientListStacking = nub . concatMap (maybe [] (\(W.Stack x l r) -> reverse l ++ r ++ [x]) . W.stack) $ ws
-    whenChanged (ClientListStacking clientListStacking) $ setClientListStacking clientListStacking
+    -- @ws@ is sorted in the "workspace order", which, by default, is
+    -- the lexicographical sorting on @WorkspaceId@.
+    -- @_NET_CLIENT_LIST_STACKING@ is expected to be in the "bottom-to-top
+    -- stacking order".  It is unclear what that would mean for windows on
+    -- invisible workspaces, but it seems reasonable to assume that windows on
+    -- the current workspace should be "at the top".  With the focused window to
+    -- be the top most, meaning the last.
+    --
+    -- There has been a number of discussions on the order of windows within a
+    -- workspace.  See:
+    --
+    --   https://github.com/xmonad/xmonad-contrib/issues/567
+    --   https://github.com/xmonad/xmonad-contrib/pull/568
+    --   https://github.com/xmonad/xmonad-contrib/pull/772
+    let clientListStacking =
+          let wsInFocusOrder = W.hidden s
+                               ++ (map W.workspace . W.visible) s
+                               ++ [W.workspace $ W.current s]
+              stackWindows (W.Stack cur up down) = reverse up ++ down ++ [cur]
+              workspaceWindows = maybe [] stackWindows . W.stack
+              -- In case a window is a member of multiple workspaces, we keep
+              -- only the last occurrence in the list.  One that is closer to
+              -- the top in the focus order.
+              uniqueKeepLast = reverse . nub . reverse
+           in uniqueKeepLast $ concatMap workspaceWindows wsInFocusOrder
+    whenChanged (ClientListStacking clientListStacking) $
+      setClientListStacking clientListStacking
 
     -- Set current desktop number
     let current = W.currentTag s `elemIndex` map W.tag ws
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,6 +55,14 @@
     - Deprecated the entire module, use `XMonad.Actions.WithAll`
       instead.
 
+  * `XMonad.Hooks.EwmhDesktops`
+
+    - `_NET_CLIENT_LIST_STACKING` puts windows in the current workspace at the
+      top in bottom-to-top order, followed by visible workspaces, followed by
+      invisible workspaces.  Within visible and invisible groups, workspaces are
+      ordered lexicographically, as before.  Currently focused window will
+      always be the topmost, meaning the last in the list.
+
 ### New Modules
 
 * `XMonad.Hooks.OnPropertyChange`:


XMonad/Hooks/EwmhDesktops.hs
```diff
@@ -320,10 +320,28 @@ ewmhDesktopsLogHook' EwmhDesktopsConfig{workspaceSort, workspaceRename} = withWi
...
+    -- @_NET_CLIENT_LIST_STACKING@ is expected to be in the "bottom-to-top
+    -- stacking order".  It is unclear what that would mean for windows on
+    -- invisible workspaces, but it seems reasonable to assume that windows on
+    -- the current workspace should be "at the top".  With the focused window to
+    -- be the top most, meaning the last.
+    let clientListStacking =
+          let wsInFocusOrder = W.hidden s
+                               ++ (map W.workspace . W.visible) s
+                               ++ [W.workspace $ W.current s]
+              stackWindows (W.Stack cur up down) = down ++ reverse up ++ [cur]
```
TheMC47, line 334: Based on the previous code, shouldn't this be like this? Or was there a mistake there too?
```haskell
              stackWindows (W.Stack cur up down) = reverse up ++ down ++ [cur]
```

ilya-bobyr: When I was writing this, I thought that the previous code incorrect.
But now that you ask, I am not so sure :)

"Bottom-to-top focus order" means that things that are higher in the focus order should be closer to the end of the list.
But I actually do not know if `up` or `down` is related to the visual stacking.
I think it depends on the layout used for the workspace.
It seems reasonable to assume that either the first element of `up` (because it is reversed) or the first element of `down` would be next to `cur` in the focus order?

Meaning that it could be either of:
```haskell
stackWindows (W.Stack cur up down) = reverse down ++ up ++ [cur]
```
or
```haskell
stackWindows (W.Stack cur up down) = up ++ reverse down ++ [cur]
```

What do you think?

geekosaur: The problem with this is that in a tiling window manager there is no actual "visual stacking", and in xmonad the usual focus changing primitives use `up` or `down` (pushing `focus` onto one and then popping the other into `focus`, unless this causes wraparound).

geekosaur: Also, for those layouts that don't actually tile as such, there's no guarantee that what the layout does matches what the window stack suggests, and it might be necessary to duplicate `windows`'s zipping the stack against the result of `runLayout` (a list of rectangles in hopefully stacking order, if a meaningful one exists) to get an accurate stacking list.

ilya-bobyr: A complete solution might be more complex indeed.
I only care for the focused window to be the last on the list.
Other than that, I can code something that is simple for the rest of the windows, if there is a good suggestion.
But I am not sure that I am up to coding an algorithm that reconstructs stacking order from the rectangles.

TheMC47: > But I actually do not know if up or down is related to the visual stacking.

I don't think this is relevant in the general case, except if some layout stacks. I don't see why we should go through the trouble of implementing this accurately. I even think the previous implementation (trivially) fulfills the spec because we can assume the layouts don't stack. That said, this is technically a breaking change (not that I think that's problematic). 

I'm not familiar with the spec by any chance, it's just that the change seemed unjustified to me.

This was discussed previously in https://github.com/xmonad/xmonad-contrib/pull/568 and https://github.com/xmonad/xmonad-contrib/issues/567. Any thoughts @liskin  @slotThe ?

XMonad/Hooks/EwmhDesktops.hs
```diff
@@ -320,10 +320,28 @@ ewmhDesktopsLogHook' EwmhDesktopsConfig{workspaceSort, workspaceRename} = withWi
...
+              workspaceWindows = maybe [] stackWindows . W.stack
+              -- In case a window is a member of multiple workspaces, we keep
+              -- only the last occurrence in the list.  One that is closer to
+              -- the top in the focus order.
+              uniqueKeepLast ls = [ e | (i, e) <- zip [0..] ls
+                                      , notElem e $ take i ls
+                                  ]
```
TheMC47, line 341: This is not correctly implemented. 
```haskell
ghci> uniqueKeepLast [1,3,1]
[1,3]
```


```suggestion
              uniqueKeepLast = reverse . nub . reverse
```

ilya-bobyr: You are right.  I should have checked it.
Changed to

```haskell
              uniqueKeepLast ls = [ e | (i, e) <- zip [0..] ls
                                      , notElem e $ drop (i + 1) ls
                                  ]
```

But if you think that `reverse . nub . reverse` is a better choice, I'll change it.

XMonad/Hooks/EwmhDesktops.hs
ilya-bobyr: I might have made a mistake, changing the order of the windows within a given workspace span.
My main focus was making sure that the currently focused window is the last in the list.
After reading both #568 and #567, it seems it is a consensus that it should be the case.
All the other of my changes are, hopefully, readability improvements.

I can change the PR to only affect the order of workspace spans within `_NET_CLIENT_LIST_STACKING`.

In my defense, it seems I am not the only one thinking that the current ordering is reversed.
I think this comment also suggested:

```haskell
stackWindows (W.Stack cur up down) = up ++ reverse down ++ [cur]
```
https://github.com/xmonad/xmonad-contrib/issues/567#issuecomment-870855824

And I've made this change, thinking it was just a typo.

XMonad/Hooks/EwmhDesktops.hs
liskin: `reverse . nub . reverse` is more readable for most Haskellers, and not any slower, so I'd say it's preferable.

XMonad/Hooks/EwmhDesktops.hs
liskin: I think the order is unpredictable to the point it simply does not and can not matter. If I'm reading the code in X.L.Magnifier correctly, in addition to making sure the magnified window is on top, it reverses the order of windowrects that came from the underlying layout:
https://github.com/xmonad/xmonad-contrib/blob/3d65a6bf72236a0b54b63b6c2c71854eb17ed3d3/XMonad/Layout/Magnifier.hs#L226-L231

```haskell
> let mag w ws | w == 3 = ws ++ [w] | otherwise = w : ws in reverse (foldr mag [] [1,2,3,4,5])
[3,5,4,2,1]
```

So we have various layouts returning windowrects in various orders, and then we have modifiers that just reverse the order when used… It's simply not worth trying to find a "best" order here. :-)

XMonad/Hooks/EwmhDesktops.hs
ilya-bobyr: Got it.  Changed as you suggested.

XMonad/Hooks/EwmhDesktops.hs
ilya-bobyr: I've restored the order within a workspace as it was before.
Now the only functional change should be in how the workspace spans are ordered.
The current workspace will be the last, and the currently focused window will be the last as well.

Added a link to this and the other two discussion in a comment.
Could help the next person looking at this code, I think.

XMonad/Hooks/EwmhDesktops.hs
slotThe: Not to bikeshed this too much, but since we are talking about _all_ windows here, it may be advisable to use [nubOrd](https://hackage.haskell.org/package/containers-0.6.6/docs/Data-Containers-ListUtils.html#v:nubOrd)

liskin: Oh, didn't even realize that one is in the standard library! (But it probably won't be for the entire range of stackage versions we support...)

slotThe: Ah, indeed, GHC 8.4.4 still ships with `containers` version 0.5.11.0, which is the last version that does not have this...

liskin: We could possibly stop supporting 8.4, as Debian stable has 8.8 and Ubuntu 20.04 has 8.6, and those were the benchmark I recall.

geekosaur: Or just define it ourselves, since it's dead simple (`Set.toList . Set.fromList`).

TheMC47 (approved):
Thanks for your contribution :tada:  I think the code looks fine now. It's not worth it to make the logic here too complicated, and the property itself doesn't make much sense for tiling window managers. The issues you linked should be helpful for whoever wants to pick this up in the future.

XMonad/Hooks/EwmhDesktops.hs
slotThe: That would destroy the order of things, right?  `nubOrd` in containers is defined in a way so that doesn't happen; something like

``` haskell
nubOrd :: Ord b => [a] -> [a]
nubOrd = go Set.empty
 where
  go _ [] = []
  go s !(!x : xs)
    | x `Set.member` s =     go s                 xs
    | otherwise        = x : go (Set.insert fx s) xs
```

Maybe `nub` is fine here and this really is bikeshedding, I don't know (I remember seeing improvements starting only at about 20 or 30 elements)

geekosaur: Pretty much, yes. I know I've told people that our main use of `nub` is fine because if you have enough screens for `nubOrd` to show an improvement then you have bigger problems than `nub`…

ilya-bobyr: I do not have an opinion here, but I can make the change, if `nubOrd` is desired.
I could either remove `8.4.4` from the tested list, and bump both `base` and `containers` minimal version.
Or I can put `nubOrd` implementation somewhere.
It seems that the library implementation is a bit more complex, as it supports fusion.
But I can also implement it just for this case at the use site, calling it `uniqueKeepLast`.

slotThe (approved):
It's probably fine this way, I agree.

TheMC47:
Will merge later if no one obejcts

PR #774, opened 2022-10-30
X.U.NamedScratchpad: Add exclusive scratchpad capabilities

slotThe:
Getting rid of another scratchpad module by introducing its functionality into NamedScratchpad.

It _works_, but sadly the implementation isn't quite as smooth as one would like it to be, so I'm posting it here as a draft while I think of more sensible ways to do this.

TODOs:

  - [x] Documentation
    - [x] In the introduction 
  - [x] Feature parity with XMonad.Util.ExclusiveScratchpads
    - [x] Deprecate the above
  - [x] `CHANGES.md`

### Checklist

        manually, ...) and concluded: Tested them manually for now—seems to work.  I don't know anyone who actually uses ExclusiveScratchpads though, so not sure this is enough.

### DIFF
--- a/XMonad/Util/NamedScratchpad.hs
+++ b/XMonad/Util/NamedScratchpad.hs
@@ -1,5 +1,6 @@
-{-# LANGUAGE InstanceSigs     #-}
-{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE InstanceSigs   #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ViewPatterns   #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Util.NamedScratchpad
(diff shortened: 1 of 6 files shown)


XMonad/Util/NamedScratchpad.hs
```diff
@@ -124,23 +127,51 @@ data NamedScratchpad = NS { name   :: String      -- ^ Scratchpad name
...
+--
+--   (ii) Replace possibly dummy scratchpads in @nspExclusives@ with
+--        proper values.  For convenience, the user may specify
+--        exclusive scratchpads by name in the startup hook.  However,
+--        we don't necessarily have all information then to immediately
+--        turn these into proper NamedScratchpads.  As such, we thinly
+--        wrap the names into an NSP skeleton, to be filled in later.
+--        This function, to be executed _before_
+--        'someNamedScratchpadAction' is the (latest) point where that
+--        happens.
```
slotThe, line 158: This is obviously a bit... sad.  In order to stay backwards compatible (and NSP is a module where we really want this) we have to probably do _something_ with lots of hidden assumptions, but hopefully it won't have to be this extreme.  I was playing around a bit with GADTs (having something like `NSPState 'Filled` and `NSPState 'Incomplete`) but extensible state doesn't really like this.

XMonad/Util/NamedScratchpad.hs
```diff
@@ -251,9 +282,17 @@ someNamedScratchpadAction f runApp _ns scratchpadName = do
 
             case NE.nonEmpty matchingOnCurrent of
                 -- no matching window on the current workspace -> scratchpad not running or in background
-                Nothing -> case NE.nonEmpty matchingOnAll of
-                    Nothing   -> runApp conf
-                    Just wins -> f (windows . W.shiftWin (W.currentTag winSet)) wins
+                Nothing -> do
+                    -- before we do anything, check for exclusive scratchpads to hide
+                    whenJust (nspExclusives !? name conf) $ \unwanted -> do
+                        let isUnwanted = foldl' (\qs nsp -> qs <||> query nsp) (pure False) unwanted
```
slotThe, line 288: Since we really only want the queries of the exclusive scratchpads, it may or may not be preferrable to just store a `Map String (Query Bool)` instead.

PR #777, opened 2022-11-12
XMonad.Actions.Search: Added search engines

Luc-Saccoccio:
### Description

I added the search engines for the AUR, Flora, nCatLab, ProtonDB, RosettaCode, Sourcehut, Steam, and the Void Linux Packages. I also reorganized the list alphabetically and added the entry for `wiktionary` which wasn't mentioned.

Those changes were motivated by
* The accessibility of diverse search engines, not only limited to standard one (as it was already the case with IMDb being on the list), and with a bias toward mathematics, coding and games.
* The appearance of other packages databases (here, the Arch and Void ones).
* The mention of the undocumented `wiktionary` function.

### Checklist

        manually, ...) and concluded: it builds, it passes the tests and it works.

### DIFF
--- a/XMonad/Actions/Search.hs
+++ b/XMonad/Actions/Search.hs
@@ -31,14 +31,17 @@ module XMonad.Actions.Search (   -- * Usage
                                  prefixAware,
                                  namedEngine,
 
-                                 amazon,
                                  alpha,
+                                 amazon,
+                                 aur,
                                  codesearch,
                                  deb,
                                  debbts,
                                  debpts,
                                  dictionary,
+                                 duckduckgo,
                                  ebay,
+                                 flora,
                                  github,
                                  google,
                                  hackage,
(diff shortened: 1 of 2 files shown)


geekosaur:
Compile with warnings is complaining about your new search engine definitions missing type signatures, and in addition complains that `voidpackages` is defined but not used.

Luc-Saccoccio:
I'm really sorry, it should be fixed now :/

geekosaur:
I should also note that git can behave oddly if the first line of the commit message is longer than 50 characters or the second line is nonempty (since it gets combined with the first and usually pushes it over that limit). This is usually only cosmetic, though. See for example the result of listing branches verbosely.

geekosaur:
Thanks!

XMonad/Actions/Search.hs
```diff
@@ -306,19 +335,24 @@ imdb          = searchEngine "imdb"          "https://www.imdb.com/find?s=all&q=
...
 openstreetmap = searchEngine "openstreetmap" "https://www.openstreetmap.org/search?query="
+protondb      = searchEngine "protondb"      "https://www.protondb.com/search?q="
+rosettacode   = searchEngine "rosettacode"  "https://rosettacode.org/w/index.php?search="
 scholar       = searchEngine "scholar"       "https://scholar.google.com/scholar?q="
+sourcehut     = searchEngine "sourcehut"     "https://sr.ht/projects?search="
 stackage      = searchEngine "stackage"      "https://www.stackage.org/lts/hoogle?q="
+steam         = searchEngine "steam"         "https://store.steampowered.com/search/?term="
 thesaurus     = searchEngine "thesaurus"     "https://thesaurus.com/browse/"
+vocabulary    = searchEngine "vocabulary"    "https://www.vocabulary.com/search?q="
+voidpackages  = searchEngine "voidpackages"  "https://voidlinux.org/packages/?arch=x86_64&q="
```
slotThe, line 348: What if I want to search for musl packages? :)  Maybe adding at least one variant would be good here (no big deal, though, I will push something tomorrow)

PR #778, opened 2022-11-19
Drop GHC 8.4

slotThe:
### Description

Now that Debian has 8.6, it's time to drop 8.4.

Contrib version of https://github.com/xmonad/xmonad/pull/430

##### 0f69215f ci: Update to GHC 9.2.5

+ A new stackage LTS is out with GHC 9.2.5, so test for this.
+ Update the haskell-ci jobs accordingly from 9.2.4.

Many instance declarations can now be derived either by DerivingVia or
GeneralizedNewtypeDeriving.

##### 30d35948 ci: Drop support for GHC 8.4

Debian stable is not on 8.6, which was always our guide as to which GHC
versions we want to support.

### Checklist

        manually, ...) and concluded: compiles :)

### DIFF
--- a/XMonad/Prompt/OrgMode.hs
+++ b/XMonad/Prompt/OrgMode.hs
@@ -67,7 +67,12 @@ import XMonad.Util.XSelection (getSelection)
 import XMonad.Util.Run
 
 import Control.DeepSeq (deepseq)
-import Data.Time (Day (ModifiedJulianDay), NominalDiffTime, UTCTime (utctDay), addUTCTime, defaultTimeLocale, formatTime, fromGregorian, getCurrentTime, iso8601DateFormat, nominalDay, toGregorian)
+import Data.Time (Day (ModifiedJulianDay), NominalDiffTime, UTCTime (utctDay), addUTCTime, fromGregorian, getCurrentTime, nominalDay, toGregorian)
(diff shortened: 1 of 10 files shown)


XMonad/Prompt/OrgMode.hs
```diff
@@ -382,7 +387,11 @@ toOrgFmt tod day =
   mconcat ["<", isoDay, " ", take 3 $ show (dayOfWeek day), time, ">"]
  where
   time   :: String = maybe "" ((' ' :) . show) tod
+#if MIN_VERSION_time(1, 9, 0)
+  isoDay :: String = iso8601Show day
+#else
   isoDay :: String = formatTime defaultTimeLocale (iso8601DateFormat Nothing) day
+#endif
```
slotThe, line 394: I guess this is one vote for dropping 8.6 as well in the near-ish future, but not sure if it's enough (the module has to use `CPP` anyways for the exports when testing so...)

liskin:
`30d35948148d369be12634255b3d079d7ca8031b`'s commit message has a typo ("not" instead of "now") and also isn't true—Debian stable has 8.8, it's Ubuntu 20.04 that only has 8.6.

liskin (approved):

PR #418, opened 2020-12-09
X.L.TrackFloating: Clean up and simplify

liskin:
### Description

https://github.com/xmonad/xmonad-contrib/pull/253 added a similar module using a somewhat simpler implementation. Borrow some ideas and share some code, so that we can possibly deprecate one in favor of the other.

### Checklist

  - [X] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

  - [ ] I updated the `CHANGES.md` file

  - n/a I updated the `XMonad.Doc.Extending` file (if appropriate)

PR #783, opened 2022-12-04
Change X.L.StateFull semantics for multi-monitor

LSLeary:
### Description

> The "last true focus" previously tracked by `X.L.StateFull.FocusTracking` was the `WindowSet` focus. With the intention of improving the multi-monitor experience, it now tracks `WindowSpace` focus instead.
> 
> That said, I was actually using two or three monitors when I originally wrote this module, so it may not be mere oversight—I have a vague feeling I used the `WindowSet` focus for a reason. If so, that reason is lost to oblivion; we'll have to rediscover it.

This change is a bit too trivial for a PR, but on account of the above, I thought I'd give others the chance to have a look at it before merging.

N.B.: Due to the variety of related messes I've collected in my local xmonad/-contrib branches and config, it's currently a PITA to test anything, so I haven't.

### Checklist

        manually, ...) and concluded: ... the CI passes.

      - Subtle behavioural changes do not merit changelog entries.

### DIFF
--- a/XMonad/Layout/StateFull.hs
+++ b/XMonad/Layout/StateFull.hs
@@ -72,12 +72,12 @@ instance LayoutClass l Window => LayoutClass (FocusTracking l) Window where
   description (FocusTracking _ child)
     | chDesc == "Full"  = "StateFull"
     | ' ' `elem` chDesc = "FocusTracking (" ++ chDesc ++ ")"
-    | otherwise           = "FocusTracking " ++ chDesc
+    | otherwise         = "FocusTracking "  ++ chDesc
     where chDesc = description child
 
   runLayout (W.Workspace i (FocusTracking mOldFoc childL) mSt) sr = do
 
-    mRealFoc <- gets (W.peek . windowset)
+    mRealFoc <- gets (W.peek . W.view i . windowset)
     let mGivenFoc = W.focus <$> mSt
         passedMSt = if mRealFoc == mGivenFoc then mSt
                     else (mOldFoc >>= \oF -> findZ (==oF) mSt) <|> mSt


slotThe (approved):
No objections from my side (but I also don't use X.L.StateFull so... :)

liskin:
This is actually the same as https://github.com/xmonad/xmonad-contrib/pull/468, which I should've ported to X.L.StateFull but forgot. Its PR description and commit message have more details than this PR, and there's also a changelog entry, so feel free to steal anything from there to make this commit/PR more of a "fix bug" rather than "change semantics but dunno what it does really". Or just merge it as it is, whatever. :-)

Anyway, a little additional context: in https://github.com/xmonad/xmonad-contrib/pull/418 I noticed that X.L.StateFull and X.L.TrackFloating are the same thing and used your simpler (easier to understand, reason about) implementation in X.L.TrackFloating, with the intention of eventually deprecating X.L.StateFull — it was merged 8 years after X.L.TrackFloating appeared, doing exactly the same thing. Unfortunately this got buried under a pile of other TODOs, and forgotten. And when I did https://github.com/xmonad/xmonad-contrib/pull/468, I forgot StateFull is still a thing…

Perhaps the reason I never did the deprecation was because I thought I'd drop X.L.StateFull's code, reexport X.L.TrackFloating with a backward compatible interface, and drop it after a few releases, but one is a LayoutClass and the other a LayoutModifier, so full backwards compat isn't possible. And it wasn't worth spending any more time thinking about this. But really I should have just slapped a DEPRECATED pragma on it and it would've been resolved by now, and also it'd be clear that a fix in one needs to be ported to the other.

LSLeary:
Closed in favour of #784.

PR #784, opened 2022-12-05
Extract redundant layouts into new module: X.L.FocusTracking

LSLeary:
### Description

> `X.L.StateFull`'s `FocusTracking` and the eponymous `TrackFloating` were redundant, hence the former was slated for deprecation. However, the latter and its host module are somewhat poorly named; the layout modifier has little relation to floats. As such, it's renamed and rehosted, becoming the eponymous `FocusTracking`.
> 
> The redundant offerings in the original modules are redefined in terms of the new module where possible, and deprecated.
> 
> See: #418; comments on #253, #783.

The situation is a bit of a mess. This seemed like the best way to resolve it, but I'm open to other options.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@
 
 ### Breaking Changes
 
+  * Deprecated the `XMonad.Layout.StateFull` module and
+    `XMonad.Layout.TrackFloating.(t|T)rackFloating` in favour of
+    `XMonad.Layout.FocusTracking`.
+
   * Dropped support for GHC 8.4.
 
   * `XMonad.Util.ExclusiveScratchpads`
@@ -76,6 +80,10 @@
       always be the topmost, meaning the last in the list.
 
 ### New Modules
+
+  * `XMonad.Layout.FocusTracking`.
+
+    - Replaces `X.L.StateFull` and half of `X.L.TrackFloating`.
 
   * `XMonad.Actions.MostRecentlyUsed`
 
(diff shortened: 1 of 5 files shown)


LSLeary:
@geekosaur The closest we could get to doing it "right" in contrib would be a layout modifier that tracks a stack for the underlying layout and manually handles the changes it sees every time the layout runs, interpreting insertions and deletions to shift focus accordingly. I don't want to write it, though—I'm tired of spending my efforts on hacky code to work around xmonad's weaknesses. Better to fix problems at the source.

liskin (approved):
Yeah, looks legit.

The docs could be improved a bit I think.

TrackFloating docs say this:

> Layout modifier that tracks focus in the tiled layer while the floating layer
or another sublayout is in use. This is particularly helpful for tiled layouts
where the focus determines what is visible. It can also be used to improve the
behaviour of a child layout that has not been given the focused window.

I think FocusTracking should mention the floating layer, it's an important usecase, possibly even link to the two old bugs on code.google.com as those links still somehow work. :-)
And TrackFloating module's purpose is now just UseTransientFor, so the module doc might need updating anyway.

slotThe (approved):
@LSLeary feel free to merge after the docs have been updated

LSLeary:
Sorry about my sloth. I picked up a couple of other projects over the holidays and got stuck on one of them.

I've made the changes to the docs, but I didn't touch the rest of `X.L.TrackFloating`. The docs in that module have a lot of cruft—if I were to rewrite them, I'd mostly just be stripping them away. I don't mind doing that, but perhaps someone else wants to edit them a bit more selectively.

PR #794, opened 2023-01-20
X.U.EZConfig: Add remapKeysP

slotThe:
This was discussed on IRC, and I thought it'd be a neat addition.

The obvious caveat (also mentioned in the documentation) is that this does not work for submap keys, as these are not transparently bound.  One could provide something like

``` haskell
additionalRemapKeysP :: [(String, String)] -> [(String, X ())] -> XConfig l -> XConfig l
```

which could sort of slurp some keys from the second argument, but this seems a bit patchworky to me, so I've left it at `remapKeysP` for now.

### Checklist

        manually, ...) and concluded: manually tested this; don't think we can easily write property test for this.

### DIFF
--- a/XMonad/Util/EZConfig.hs
+++ b/XMonad/Util/EZConfig.hs
@@ -24,6 +24,7 @@ module XMonad.Util.EZConfig (
                              -- * Adding or removing keybindings
 
                              additionalKeys, additionalKeysP,
+                             remapKeysP,
                              removeKeys, removeKeysP,
                              additionalMouseBindings, removeMouseBindings,
 
@@ -105,6 +106,40 @@ additionalKeys conf keyList =
 additionalKeysP :: XConfig l -> [(String, X ())] -> XConfig l
 additionalKeysP conf keyList =
     conf { keys = \cnf -> M.union (mkKeymap cnf keyList) (keys conf cnf) }
+
+-- |
+-- Remap keybindings from one binding to another.  More precisely, the
+-- input list contains pairs of the form @(TO, FROM)@, and maps the
+-- action bound to @FROM@ to the key @TO@.  For example, the following
+-- would bind @"M-m"@ to what's bound to @"M-c"@ (which is to close the
+-- focused window, in this case):
+--
+-- > main :: IO ()
+-- > main = xmonad $ def `remapKeysP` [("M-m", "M-c")]
+--
+-- NOTE: Submaps are not transparent, and thus these keys can't be
+-- accessed in this way: more explicitly, the @FROM@ string may **not**
+-- be a submap.  However, the @TO@ can be a submap without problems.
+-- This means that
+--
+-- > xmonad $ def `remapKeysP` [("M-m", "M-c a")]
+--
+-- is illegal (and indeed will just disregard the binding altogether),
+-- while
+--
+-- > xmonad $ def `remapKeysP` [("M-c a", "M-m")]
+--
+-- is totally fine.
+remapKeysP :: XConfig l -> [(String, String)] -> XConfig l
+remapKeysP conf keyList =
+    conf { keys = \cnf -> mkKeymap cnf (keyList' cnf) <> keys conf cnf }
+  where
+   keyList' :: XConfig Layout -> [(String, X ())]
+   keyList' cnf =
+     mapMaybe (traverse (\s -> case readKeySequence cnf s of
+                                 Just [ks] -> keys conf cnf M.!? ks
+                                 _         -> Nothing))
+              keyList
 
 -- |
 -- Remove standard keybindings you're not using. Example use:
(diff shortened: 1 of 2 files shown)


XMonad/Util/EZConfig.hs
```diff
@@ -106,6 +107,37 @@ additionalKeysP :: XConfig l -> [(String, X ())] -> XConfig l
 additionalKeysP conf keyList =
     conf { keys = \cnf -> M.union (mkKeymap cnf keyList) (keys conf cnf) }
 
+-- |
+-- Remap keybindings from one binding to another.  For precisely, the
```
geekosaur, line 111: "More precisely", perhaps?

XMonad/Util/EZConfig.hs
```diff
@@ -106,6 +107,37 @@ additionalKeysP :: XConfig l -> [(String, X ())] -> XConfig l
...
+-- > main = xmonad $ def `remapKeysP` [("M-m", "M-c")]
+--
+-- NOTE: Submaps are not transparent, and thus these keys can't be
+-- accessed in this way: more explicitly, the @FROM@ string may **not**
+-- be a submap.  However, the @TO@ can be a submap without problems.
+-- This means that
+--
+-- > xmonad $ def `remapKeysP` [("M-m", "M-c a")]
+--
+-- is illegal (and indeed will bind @"M-m"@ to @"M-c"@!), while
```
geekosaur, line 127: Should this perhaps catch and report an error (presumably to `stderr` or via `xmessage`) instead of doing something unexpected?

slotThe: We probably should, but it's harder than it looks.  We can't really use `xmessage` and friends because `remapKeysP` is not in `X` (and using `error` and such would obviously result in runtime crashes).  This is why, for the rest of EZConfig, the hack with `checkKeymap` exists.  I don't think it's worth having a whole second `checkRemapKeymap` for this, tbh, so I'm not sure what to do here.

geekosaur: Better to ignore it entirely, then, instead of applying only the first part of the key sequence.

slotThe: Good point, these keybindings should now just be ignored

XMonad/Util/EZConfig.hs
```diff
@@ -106,6 +107,40 @@ additionalKeysP :: XConfig l -> [(String, X ())] -> XConfig l
...
+-- action bound to @FROM@ to the key @TO@.  For example, the following
+-- would bind @"M-m"@ to what's bound to @"M-c"@ (which is to close the
+-- focused window, in this case):
+--
+-- > main :: IO ()
+-- > main = xmonad $ def `remapKeysP` [("M-m", "M-c")]
+--
+-- NOTE: Submaps are not transparent, and thus these keys can't be
+-- accessed in this way: more explicitly, the @FROM@ string may **not**
+-- be a submap.  However, the @TO@ can be a submap without problems.
```
TheMC47, line 122: Could you explain why? I don't get it

geekosaur: To parse `@FROM@` as a submap, you'd have to be able to peek inside the compiled code for `Submap`.

geekosaur (approved):
LGTM!

XMonad/Util/EZConfig.hs
TheMC47: Oh right :sweat_smile:

TheMC47 (approved):

XMonad/Util/EZConfig.hs
geekosaur: Actually, it occurs to me that `@TO@` has submap issues as well: since we can't look inside a submap, if you had two bindings attempting to use the same submap (including one in `RemapP` and one in `additionalKeysP`), one would be overwritten.

slotThe: I think this is consistent with how `additionalKeysP` "allows" you to do this (rather, it doens't—maybe can't—have a mechanism to give you a compile-time error).  Plus, we also can't check whether other keybindigs are already taken, so disallowing `TO` from being a submap would seem a bit arbitrary to me

geekosaur: It wasn't a problem with `additionalKeysP` because it builds the submaps first, so there's no overwriting unless you use it twice for some reason. With `remapKeysP` you now have two ways to get submaps that aren't aware of each other.

Not sure how to fix this.

slotThe: I think the end result is the same—keys are overwritten (which is why `checkKeymap` checks for exactly this).  I don't think it's a big problem, tbh, so I'd just leave it as-is, unless you have strong opinions about this

PR #795, opened 2023-01-24
X.H.StatusBar: Added startAllStatusBars

TheMC47:
### Description

A tiny PR to add a function that starts status bars. My use case is to have a "Zen" mode where I hide all status bars (since even the percentages of CPU usage annoy me when I'm trying to focus)

### Checklist

  - [x] I've considered how to best test these changes: Manually


slotThe (approved):
This looks fine to me, though maybe we should add a note about how this is usually not needed and only for certain use-cases

PR #796, opened 2023-01-27
X.U.NamedActions: Do not discard all keybindings in addDescrKeys

slotThe:
Discarding keybindings may yield unintended behaviour when keybindings are added via combinators, instead of the `keys` field.  This makes the
combinators noncommutative, which is very counterintuitive.

Related: https://old.reddit.com/r/xmonad/comments/10g1v1r/deftogglestrutskey_not_working/

---

This is a PR because I don't use NamedActions and, though I've verified that this doesn't immediately break things with the simplest config, someone else should look at ti

---

### Checklist

        manually, ...) and concluded: tested manually; seems to do the trick.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -173,7 +173,13 @@
       [polybar issue](https://github.com/polybar/polybar/issues/2603)).
 
   * `XMonad.Hooks.StatusBar`
+
     - Added `startAllStatusBars` to start the configured status bars.
+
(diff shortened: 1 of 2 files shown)


geekosaur (approved):
LGTM!

TheMC47 (approved):

slotThe:
Let's see, then :)

PR #799, opened 2023-02-08
Config modifiers `infixl 4` so you can use `(++)`

geekosaur:
### Description

Per a discussion we had last November, this PR changes the fixity of `additionalKeys` and friends from the default (`infixl 9`) to `infixl 4`. This makes it possible to use `(++)` to concatenate lists within them without requiring extra parentheses, since `(++)` is `infixl 5`.

This should not affect existing configs.

### Checklist

        manually, ...) and concluded: tested locally

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -105,6 +105,10 @@
     - Added `remapKeysP`, which remaps keybindings from one binding to
       another.
 
+    - Made `additionalKeys{,P}`, `removeKeys{,P}`, `remapKeysP`, and
+      `{additional,remove}MouseBindings` `infixl 4` so they can more easily
+      be concatenated with `(++)`.
+
   * `XMonad.Util.NamedScratchpad`
 
     - Added `addExclusives`, `resetFocusedNSP`, `setNoexclusive`,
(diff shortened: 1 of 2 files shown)


slotThe (approved):
LGTM.  I normally don't like explicit fixities for functions, but I agree that in this case it's the most ergonomic thing to do.

For future reference, a user posted a configuration akin to

``` haskell
main :: IO ()
main = xmonad $ def { modMask = mod4Mask }
  `additionalKeysP`
       [("M-<Tab>", sendMessage NextLayout)]
    ++ [("M-s " ++ k, S.promptSearchBrowser myPromptConfig myBrowser' f) | (k,f) <- searchList ]
    ++ [("M-S-s " ++ k, S.selectSearch f) | (k,f) <- searchList ]
```

which doesn't compile because of the default fixity levels.  Here is the—slightly edited—relevant part of the ensuing conversation:

```
<geekosaur>  you need parentheses around the lists you're appending
<geekosaur>  otherwise you're trying to append the lists to your config,
             which doesn't make sense as the error message says
<geekosaur>  this depends on fixity and I think we don't declare a useful
             fixity for `additionalKeysP`. if we did it would allow (++) to
             associate tighter
<Solid>      we should probably do that, actually
<geekosaur>  actually I suspect it has a fixity declaration, at minimum so it can be
             combined with removeKeys{,P}
<geekosaur>  so it may require some thought
<geekosaur>  since we can't change the fixity of (++)
<Solid>      it does not, actually
<Solid>      nothing in EZConfig does
<Solid>      combining them with removeKeys still works beause backticks are
             left associative by default
<geekosaur>  oh, for some reason I thought they were nonassoc by default
             (just checked the Report)
<geekosaur>  so we just need them to have a fixity lower than 5
<Solid>      yeah
<Solid>      make all of them `infixl {1..4}`
```

PR #800, opened 2023-02-11
Add loggers for window classname

1in1:
### Description

This PR adds loggers to expose the X window classname of the current window. It adds the following, which each have an existing counterpart for logging window titles:
- `logClassname`
- `logClassnames`
- `logClassnames'`
- `logClassnameOnScreen`
- `logClassnamesOnScreen`
- `logClassnamesOnScreen'`
- `ClassnamesFormat(..)`

This is motivated by wanting to create statusbar output which provides the classname of the current window. `XMonad.Util.NamedWindows` provides both:
- `getName`, which provides the title of a window (e.g., "Home - BBC News"), and
- `getNameWMClass`, which provides the classname of the window (e.g., "chromium").

There are currently methods for fetching the former, but not the latter. The implementation here is identical to the existing code for the window title.

Note: While it looks like there ought to be more opportunity to share code, a lot of the existing methods are exported and have non-generic signatures. I figured it easier to just rewrite.

Note: Firefox comes through as `Navigator` for me, so a common requirement in a formatter might be a string replace on that word. I'm not adding that here as I feel that's an end-user task.

### Checklist

        manually, ...) and concluded: Sufficiently small change that manual testing seems to be appropriate, looking at the existing tests. I've manually tested all of the above loggers separately.

### DIFF
--- a/XMonad/Util/Loggers.hs
+++ b/XMonad/Util/Loggers.hs
@@ -35,13 +35,16 @@ module XMonad.Util.Loggers (
     -- $xmonad
     , logCurrent, logLayout
     , logTitle, logTitles, logTitles'
+    , logClassname, logClassnames, logClassnames'
     , logConst, logDefault, (.|)
     -- * XMonad: Screen-specific Loggers
     -- $xmonad-screen
     , logCurrentOnScreen, logLayoutOnScreen
-    , logTitleOnScreen, logWhenActive
+    , logTitleOnScreen, logClassnameOnScreen, logWhenActive
     , logTitlesOnScreen, logTitlesOnScreen'
+    , logClassnamesOnScreen, logClassnamesOnScreen'
     , TitlesFormat(..)
+    , ClassnamesFormat(..)
     -- * Formatting Utilities
     -- $format
     , onLogger
@@ -58,7 +61,7 @@ import qualified XMonad.StackSet as W
 import XMonad.Hooks.StatusBar.PP
 import XMonad.Hooks.UrgencyHook (readUrgents)
 import XMonad.Util.Font (Align (..))
-import XMonad.Util.NamedWindows (getName)
+import XMonad.Util.NamedWindows (getName, getNameWMClass)
 
 import Control.Exception as E
 import XMonad.Prelude (find, fromMaybe, isPrefixOf, isSuffixOf, WindowScreen)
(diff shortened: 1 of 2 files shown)


TheMC47:
Thank you for your contribution🎉

I'm not sure about adding another PP field at this point. Maybe adding this functionality as a logger instead is better?

1in1:
That's reasonable! I'll take a look :)

geekosaur:
It's not exactly personal preference: `appName` and `className` are intended for programmatic use, while `title` is more user-oriented. That said, I can see it being useful in some cases.

XMonad/Util/Loggers.hs
```diff
@@ -251,6 +254,70 @@ instance Default TitlesFormat where
     , urgentFormat    = wrap "!" "!" . xmobarRaw . shorten 30 . xmobarStrip
     }
```
slotThe, line 256: Could we perhaps make this a little bit more generic?  We're duplicating a lot of code right now

XMonad/Util/Loggers.hs
```diff
@@ -251,6 +254,70 @@ instance Default TitlesFormat where
...
+
+-- | How to format these classnames by default when using 'logClassnames'' and
+-- 'logClassnamesOnScreen''.
+instance Default ClassnamesFormat where
+  def = ClassnamesFormat
+    { focusedFormatClassname   = wrap "[" "]" . xmobarRaw . shorten 30 . xmobarStrip
+    , unfocusedFormatClassname =                xmobarRaw . shorten 30 . xmobarStrip
+    , urgentFormatClassname    = wrap "!" "!" . xmobarRaw . shorten 30 . xmobarStrip
+    }
+
```
TheMC47, line 320: Similar to what @slotThe, maybe we use the same record `TitlesFormat` for this as well? Maybe

```
type ClassnamesFormat = TitlesFormat
```

1in1: My only worry with this is it leaves the constructors feeling a bit weird. I.e., building a `ClassnamesFormat` with a function `TitlesFormat` feels off to me. I don't consider myself a god of haskell, is there a neat way to get around this? Or, I'm happy to go with it anyway, if you feel it won't be an issue?

TheMC47: That's a valid concern. I'm not sure how big of an issue this actually is though:
- People use `def` in their configs
- There's a single function that's explicitly "deconstructing"  `ClassnamesFormat`

Let's see what others think though, I don't feel strongly about it.

geekosaur: Perhaps just refactor a bit?
```
xmobarWsFormat, xmobarFocusedFormat, xmobarUrgentFormat :: String -> String
xmobarWsFormat = xmobarRaw . shorten 30 . xmobarStrip
xmobarFocusedFormat = wrap "[" "]" . xmobarWsFormat
xmobarUrgentFormat = wrap "!" "!" . xmobarWsFormat
```

TheMC47 (approved):
LGTM, good work :tada: If you could squash the commits, that would be great. If not, I'll do that and merge (if no one else objects)

1in1:
Squashed and over to you guys when ready :) thanks!

TheMC47:
@1in1 congrats on getting your contribution merged! Looking forward to what else you have to add 😀

Issue #801, opened 2023-02-11
`XMonad.Hooks.Modal.logMode` used in `XMonad.Hooks.StatusBar.statusBarProp` issue


sagittarius-a:
### Problem Description

When I use the `XMonad.Hooks.Modal.logMode` in the `ppExtras` field of xmobar pretty-printer,
it works flawlessly.

However, if I use `ppOrder`:

- the ordering is properly applied
- it works as soon as I enter the new mode, the mode string is printed to xmobar
- I can perform actions in the current mode

But:
- But I cannot exit the mode anymore
- The mode string is still in xmobar (though I can't tell if it is a cache issue in xmobar)
- No other Xmonad keybindings works

The only way to solve the issue is to restart Xmonad. It can be very tedious if there is no way to do it without using any Xmonad keybinding.

### Steps to Reproduce

- [Optional] Open a terminal (to be able to restart Xmonad easily)
- Press `M-g` to enter the example mode
- Press `Escape` to exit the example mode

No Xmonad keybinding should be working at this point.

### Configuration File

```haskell
import XMonad
import XMonad.Hooks.StatusBar
import XMonad.Hooks.StatusBar.PP
import XMonad.Util.EZConfig
import XMonad.Hooks.Modal

main :: IO ()
main = xmonad
     . modal [noModMode, sayHelloMode]
     . withEasySB (statusBarProp "xmobar" (pure myXmobarPP)) defToggleStrutsKey
     $ def
  `additionalKeysP`
    [ ("M-g"  , setMode "Hello")   ]

myXmobarPP :: PP
myXmobarPP = def
    { ppOrder           = \[ws, l, title, mode] -> [ws, l, mode, title]
    , ppExtras          = [logMode]
    }

    -- { ppExtras          = [logMode] }

sayHelloMode :: Mode
sayHelloMode = mode "Hello" $ mkKeysEz
  [ ("g", xmessage "Hello, World!") ]

```

### Checklist

  - [X] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [X] With `xmonad` commit `391c0fc`
    - [X] With `xmonad-contrib` commit `29f0e032`

LSLeary:
A sequence of events:

 1. You run `exitMode`.
 2. There's no longer an active mode, so `logMode` produces `Nothing`.
 3. Hence the list that `ppOrder` takes shrinks to length 3.
[...]

PR #802, opened 2023-02-20
`dynamicLogString` forces its result and recovers

geekosaur:
### Description

Originally, `dynamicLogString` could have a bottom hidden in it and thereby crash the `logHook`. Under some circumstances (see #801) this could cause xmonad to get stuck. We now force the result, and `dynamicLogString` catches the exception and substitutes a message (currently `"_|_"`). Use `dynamicLogString'` for the old behavior.

Closes: #801 

### Checklist

### DIFF
--- a/XMonad/Hooks/StatusBar/PP.hs
+++ b/XMonad/Hooks/StatusBar/PP.hs
@@ -32,6 +32,7 @@ module XMonad.Hooks.StatusBar.PP (
     -- * Build your own formatter
     PP(..), def,
     dynamicLogString,
+    dynamicLogString',
     dynamicLogWithPP,
 
     -- * Predicates and formatters
@@ -55,6 +56,7 @@ module XMonad.Hooks.StatusBar.PP (
     ) where
 
 import Control.Monad.Reader
+import Control.DeepSeq
 
 import XMonad
 import XMonad.Prelude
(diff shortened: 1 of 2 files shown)


XMonad/Hooks/StatusBar/PP.hs
```diff
@@ -199,12 +209,12 @@ dynamicLogString pp = do
     -- window title
     wt <- maybe (pure "") (fmap show . getName) . S.peek $ winset
 
-    return $ sepBy (ppSep pp) . ppOrder pp $
+    return $! force (sepBy (ppSep pp) . ppOrder pp $
```
liskin, line 212: Why not `return $! force $ sepBy …` ?

XMonad/Hooks/StatusBar/PP.hs
```diff
@@ -64,6 +65,8 @@ import XMonad.Util.NamedWindows
 import XMonad.Util.WorkspaceCompare
 import XMonad.Hooks.UrgencyHook
 
+import Control.DeepSeq
```
liskin, line 68: Move this next to `import Control.Monad.Reader`?

XMonad/Hooks/StatusBar/PP.hs
geekosaur: Habit because there's already a `$` and I didn't feel like figuring out how they'd interact. I have a love-hate relationship with `$`.

PR #810, opened 2023-04-19
Put a cap on window names

### LINKED ISSUE #809: qutebrowser launch hangs xmonad w/ Tabbed
### Problem Description

xmonad hangs for good when qutebrowser is being launched in a Tabbed layout. The tab decoration gets garbled. xmonad works fine if qutebrowser is launced in a non-Tabbed layout and then brought to a Tabbed one.
screenshot

### Configuration File

```haskell
import XMonad
import XMonad.Layout.Tabbed

main = xmonad def { layoutHook = simpleTabbed }
```

### Checklist

  - I tested my configuration
    - [x] With `xmonad` version 0.17.1
    - [x] With `xmonad-contrib` version 0.17.1

geekosaur:
### Description

It turns out qutebrowser will place an entire `data:` URL in `_NET_WM_NAME`, up to at least 389K characters including newlines and possibly binary characters.

Clamping window titles to the first line, and to 2K chars because `defaultShrinker` is quadratic in the title length.

Closes #809 

### Checklist

        manually, ...) and concluded: tested in my harness

  - [ ] I updated the `CHANGES.md` file — shouldn't be necessary

### DIFF
--- a/XMonad/Layout/Decoration.hs
+++ b/XMonad/Layout/Decoration.hs
@@ -400,7 +400,11 @@ updateDecos s t f = mapM_ $ updateDeco s t f
 -- structure and the needed 'Rectangle's
 updateDeco :: Shrinker s => s -> Theme -> XMonadFont -> (OrigWin,DecoWin) -> X ()
 updateDeco sh t fs ((w,_),(Just dw,Just (Rectangle _ _ wh ht))) = do
-  nw  <- getName w
+  -- xmonad-contrib #809
+  -- qutebrowser will happily shovel a 389K multiline string into @_NET_WM_NAME@
+  -- and the 'defaultShrinker' (a) doesn't handle multiline strings well (b) is
+  -- quadratic due to using 'init'
+  nw  <- fmap (take 2048 . takeWhile (/= '\n') . show) (getName w)
   ur  <- readUrgents
   dpy <- asks display
   let focusColor win ic ac uc = maybe ic (\focusw -> case () of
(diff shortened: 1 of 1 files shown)


geekosaur:
```
[19 15:36:26] <L29Ah> hmm xmonad hangs for a few secs even with this patch
[19 15:36:50] <L29Ah> but that's good enough i guess
[19 15:44:27] <geekosaur> seemed pretty quick here
[19 15:45:21] <L29Ah> i7-8550U laptop here
[19 15:45:35] <geekosaur> mine is an older i5
[19 15:45:50] <geekosaur> and I'm running it in a xephyr in a test harness
[19 15:47:57] <L29Ah> i guess that might be due to the fact i have several such tabs in my qutebrowser
[19 15:48:09] <L29Ah> and it blinks them
[19 16:06:42] <geekosaur> what I'm actually seeing is qutebrowser hangs for a second or two. that's not under xmonad's control; drawing is solely between the client and the X server, xmonad can't even know it's happening without e.g. taking a couple of screenshots and comparing them
```

XMonad/Layout/Decoration.hs
```diff
@@ -400,7 +400,11 @@ updateDecos s t f = mapM_ $ updateDeco s t f
 -- structure and the needed 'Rectangle's
 updateDeco :: Shrinker s => s -> Theme -> XMonadFont -> (OrigWin,DecoWin) -> X ()
 updateDeco sh t fs ((w,_),(Just dw,Just (Rectangle _ _ wh ht))) = do
-  nw  <- getName w
+  -- xmonad-contrib #809
+  -- qutebrowser will happily shovel a 389K multiline string into @_NET_WM_NAME@
+  -- and the 'defaultShrinker' (a) doesn't handle multiline strings well (b) is
+  -- quadratic due to using 'init'
```
slotThe, line 406: I didn't completely follow all of this, but is there any reason why `defaultShrinker` *should* be quadratic or is that a bug we ought to fix as well?

geekosaur: It shrinks by a character at a time until the rendered version fits inside the `Decoration`. Unfortunately, you can't run that one backwards and determine the proper length of the `String` from what will fit inside the `Decoration` directly, because even if you can ensure that a monospaced font was used (we can't; that's only the default) there are Unicode glyphs that take up the space of two monospaced characters.

slotThe: But can't you do like a binary search over the length of the string? I don't know how often you only have to chop of one or two characters; I feel like when you have to chop, you have to chop quite a lot

geekosaur:
Probably. It would also partially avoid problems like splitting ligatures (flags, etc.). But I feel like that's beyond me.

XMonad/Layout/Decoration.hs
geekosaur: Probably. It would also partially avoid problems like splitting ligatures (flags, etc.). But I feel like that's beyond me.

PR #813, opened 2023-05-03
Actual PerScreen implementation (WIP)

geekosaur:
### Description

I have implemented an actual PerScreen layout (`onScreen` and `onScreens`). The old `ifWidth` is moved to a new module `XMonad.Layout.ByWidth`, and re-exported with a deprecation warning pointing to the new module.

`CHANGES.md` not updated yet, and marked as draft if I can after I open it: I'd like someone who can to test a corner case with respect to `rescreen`. I _think_ it will usually work, but may glitch if you rapidly plug and unplug an external monitor to which this layout would apply.

### Checklist

  - [ ] I updated the `CHANGES.md` file (see above)


geekosaur:
I squashed it down a bit.

XMonad/Layout/PerScreen.hs
```diff
@@ -1,75 +1,119 @@
...
+         => ScreenId -> l1 a -> l2 a -> OnScreen l1 l2 a
+onScreen s = onScreens [s]
+
+-- | Specify a layout to run on a list of screens.
+--   Note that this works by 'ScreenId'. It has a 'Num' instance, so literal
+--   screen numbers will work as expected, but if you use a binding you need
+--   to use the 'S' constructor.
+onScreens :: (LayoutClass l1 a, LayoutClass l2 a)
+          => [ScreenId] -> l1 a -> l2 a -> OnScreen l1 l2 a
+onScreens ss l1 l2 = OnScreen ss l1 l2 False -- @@@ is this right?
```
TheMC47, line 66: > @@@ is this right?

Are you asking about the flag? I think so since as soon as runLayout is executed the flag will be set correctly

XMonad/Layout/PerScreen.hs
```diff
@@ -1,75 +1,119 @@
...
+        -- It'd be easier if 'runLayout' were passed the whole 'StackSet' with
+        -- 'windows's idea of the current screen rects, but then it wouldn't know
+        -- which 'Workspace' to work on…
+        which <- withDisplay getCleanedScreenInfo >>= \case
+            [] -> trace "OnScreen: no screens?" >> return False
+            rs -> return $ maybe 0 fst (find (f r) (zip [0..] rs)) `elem` ss
+                where f lr (_,sr) = rect_x lr >= rect_x sr &&
+                                    rect_x lr < rect_x sr + fi (rect_width sr) &&
+                                    rect_y lr >= rect_y sr &&
+                                    rect_y lr < rect_y sr + fi (rect_height sr)
```
TheMC47, line 87: I agree, this does indeed suck. Maybe we hide it in `XMonad.Prelude.deriveActiveScreen`? It would make the code easier to read

TheMC47 (approved):
LGTM, one minor nitpick.

XMonad/Layout/PerScreen.hs
geekosaur: That was what I figured but I can't recall if there are special cases where the default might be used inappropriately. But even if there were it would be on an inactive screen by definition, which is why I used `False`.

XMonad/Layout/PerScreen.hs
geekosaur: Most of the ugliness there is just testing if the (top left corner of the) current screen rectangle is contained in a screen, no? The rest is the `zip` which is a bit deeply embedded to be yanked out.

geekosaur: And I'd have a different kind of ugliness if I were building a list of screen rectangles from `ScreenDetail`s anyway.

geekosaur:
In case anyone's wondering, @liskin expressed a concern and plans to review this within the next few days, so don't rush into committing it just yet,

XMonad/Layout/ByWidth.hs
```diff
@@ -0,0 +1,77 @@
...
+mkNewPerScreenF :: PerScreen l1 l2 a -> Maybe (l2 a) ->
+                 PerScreen l1 l2 a
+mkNewPerScreenF (PerScreen w _ lt lf) mlf' =
+    PerScreen w False lt $ fromMaybe lf mlf'
+
+instance (LayoutClass l1 a, LayoutClass l2 a, Show a) => LayoutClass (PerScreen l1 l2) a where
+    runLayout (W.Workspace i p@(PerScreen w _ lt lf) ms) r
+        | rect_width r > w    = do (wrs, mlt') <- runLayout (W.Workspace i lt ms) r
+                                   return (wrs, Just $ mkNewPerScreenT p mlt')
+        | otherwise           = do (wrs, mlt') <- runLayout (W.Workspace i lf ms) r
```
liskin, line 69: Just doing `runLayout` with the other layout whenever the condition changes will leave decorations (and other stuff like this) around. In other words, this suffers from https://github.com/xmonad/xmonad-contrib/issues/75. You need to send `Hide` to the last layout first before `runLayout`ing the new one, like this: https://github.com/xmonad/xmonad-contrib/pull/582#issuecomment-898495845

Also, you need to handle `ReleaseResources` and pass it on to both layouts.

XMonad/Layout/PerScreen.hs
```diff
@@ -1,75 +1,119 @@
...
+onScreens ss l1 l2 = OnScreen ss l1 l2 False -- @@@ is this right?
+
+data OnScreen l1 l2 a = OnScreen [ScreenId] (l1 a) (l2 a) Bool
+    deriving (Read, Show)
+
+instance (LayoutClass l1 a, LayoutClass l2 a, Show a) => LayoutClass (OnScreen l1 l2) a where
+    runLayout (W.Workspace i p@(OnScreen ss l1 l2 _) ms) r = do
+        -- this sucks.
+        -- You might think we could use the existing screen detail, but we may be
+        -- invoked from 'rescreen' in which case only 'rescreen' and 'windows'
```
liskin, line 75: But `windows` does `modify (\s -> s { windowset = ws })` pretty early (definitely before running any layouts) so why can't you just `withWindowSet` like X.L.NoBorders does (it also has logic to detect which screen is which based on the rect passed)?

liskin: By not using the windowset this won't work with XMonad.Layout.LayoutScreens.

XMonad/Layout/ByWidth.hs
liskin: Oh, now I see you're just moving stuff around and this is still the same code that was there. So I guess that's an argument for not fixing it in the scope of this PR.

That then leads me to a thought whether it wouldn't be better to not hijack the `PerScreen` name and just add an `X.L.OnScreen` module… Deprecating `PerScreen` in favor of `ByWidth` or `IfWidth` can stay, of course.

Oh, and related to my previous comment about #75 - the plan was to implement #582 generic enough so that `IfWidth` can be implemented using it and thus get #75 fixed for free. So we may want to limit the exports a bit… although we can always use `pattern` for backwards compat.

XMonad/Layout/PerScreen.hs
liskin: I don't think there are cases where it might be used inappropriately. **But** we're also not using the flag enough. We need to `Hide` whenever the flag flips, and dispatch `ReleaseResources` to both layouts.

(Might be better to finish #582 and get #75 fixed for free, like I mentioned in another comment.)

XMonad/Layout/PerScreen.hs
geekosaur: I think `LayoutScreens` should be deprecated in favor of using `xrandr`: applications can't ever work with `LayoutScreens`.

XMonad/Layout/ByWidth.hs
geekosaur: The main reason I reused `PerScreen` is consistency with `PerWorkspace` and friends.

And yes, the only changes I made to `ByWidth` were renames; everything else is the same, including the weirdnesses @TheMC47 noticed. I was actually somewhat afraid to touch them, because @ezyang is more of a Haskeller than I am and I tend to assume "fancy" code has some reason not visible to me.

geekosaur: Re `Hide` and `ReleaseResources`: I did think about that, but thought we had a more global plan to deal with those somewhere. I can add that. What worries _me_ is I don't think it scales; are there, or will there at some point be, more "global" messages that need to be dealt with by individual layouts?

XMonad/Layout/PerScreen.hs
geekosaur: If we receive `Hide`, do I need to relay it to both layouts or just the active one?

geekosaur: This should be done, although I cheated a bit on `ReleaseResources`: should I check for a new constructor being produced by the forwarded message?

Also, if we receive `Hide`, should it be relayed to both sublayouts or only the active one?

geekosaur: I now have both fully implemented `ReleaseResources` and a direct implementation of `Hide` which sends to both sublayouts because it's a PITA to send only to the active one. Also I'm relying in a few places on `Hide` being idempotent, but we may have other issues if it's not anyway.

geekosaur:
I renamed `ByWidth.hs` to `IfWidth.hs` for consistency with `IfMax.hs` and with its `ifWidth` function, and squashed everything in the process. AFAIK this is just waiting on someone to confirm that I'm handling the `Hide` and `ReleaseResources` messages properly.

PR #821, opened 2023-07-28
Export 'Mag' and 'Toggle' from Magnifier

Zawaken:
### Description

behind them.

I've made XMonad.Layout.Magnifier export All constructors under 'Magnifier' and 'Toggle', so that it is possible to make custom magnifiers, with control over size both horizontally and vertically.

Example:
```hs
-- maximizeVertical defaulting to On instead of Off
custommaximizeVertical :: l a -> ModifiedLayout Magnifier l a
custommaximizeVertical = ModifiedLayout (Mag 1 (1, 1000) On (AllWins 1))
```

### Checklist

        manually, ...) and concluded: This works exactly how I want it to work

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -120,8 +120,8 @@
 ### Bug Fixes and Minor Changes
 
   * `XMonad.Layout.Magnifier`
-    - Added `Toggle(..)` and changed `Magnifier` to `Magnifier(..)` in the exports
-      for Magnifier.hs, to allow for custom horizontal and vertical sizes for magnifiers.
+    - Exposed `Toggle(..)` and the `Mag` constructor, to allow for custom
+      horizontal and vertical sizes for magnifiers.
 
   * `XMonad.Util.Loggers`
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -119,6 +119,10 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Layout.Magnifier`
+    - Added `Toggle(..)` and changed `Magnifier` to `Magnifier(..)` in the exports
+      for Magnifier.hs, to allow for custom horizontal and vertical sizes for magnifiers.
+
   * `XMonad.Util.Loggers`
 
     - Added `logClassname`, `logClassnames`, `logClassnames'`,
--- a/XMonad/Layout/Magnifier.hs
+++ b/XMonad/Layout/Magnifier.hs
@@ -43,7 +43,8 @@ module XMonad.Layout.Magnifier
       -- * Messages and Types
       MagnifyMsg (..),
       MagnifyThis(..),
-      Magnifier,
+      Magnifier(..),
+      Toggle(..),
     ) where
 
 import Numeric.Natural (Natural)


geekosaur (approved):
LGTM, thanks!

geekosaur:
Hm, strictly this breaks PVP so should have a major version bump. But I think we're headed that way anyway, whenever we get around to the next release. (0.17.2 didn't contain any of these changes, just what was needed for GHC 9.6 and its bundled libraries.)

slotThe:
Duplicate of https://github.com/xmonad/xmonad-contrib/pull/583

Further, the `custommaximizeVertical` can already be done with the `magnify` combinator, as far as I can tell. Is there really anything that actually requires us to expose these internals?

geekosaur:
`magnify`'s documentation (and implementation; I checked) specifically expands the selected window in both directions, unlike `maximizeVertically`.

slotThe:
Ah, right, we use the same number twice… Well, looks like we just need a more general combinator then?

Zawaken:
I didn't realize this was a duplicate, but I see now that it is.

I have tested doing some [changes in a branch in my fork](https://github.com/xmonad/xmonad-contrib/compare/master...Zawaken:xmonad-contrib:magnifiercxy) that might be better, where I have changed `magnify` to allow for custom horizontal and vertical values. (the magnifiercxy name is kind of a placeholder)

Is this more in line with what you were thinking @slotThe, if so I can create a Pull request for the new branch, or merge it into this pull request.

slotThe:
> I have tested doing some [changes in a branch in my fork](https://github.com/xmonad/xmonad-contrib/compare/master...Zawaken:xmonad-contrib:magnifiercxy) that might be better, where I have changed `magnify` to allow for custom horizontal and vertical values. (the magnifiercxy name is kind of a placeholder, and to show how this would work)
>
> Is this more in line with what you were thinking @slotThe, if so I can create a Pull request for the new branch, or merge it into this pull request.

Mh, I was more thinking of creating another (yet another, I know…) general combinator; I'm not sure a breaking change is warranted here

geekosaur:
What's "a bit dirty" about it? If anything, I'd say the one that requires users to use `ModifiedLayout` directly is the dirty one; it's intended to be internal.

Zawaken:
I mean compared to my [previous proposed solution](https://github.com/xmonad/xmonad-contrib/commit/2ffca4467776f7f9b0e636ce5bec18fdf9c9be15), changing `magnify` to accept two Rationals worked like a charm, and didn't seem to break anything, so the new one in the `magnifiercxy` feels a bit dirtier if that makes sense.

But both of them work well.

It is like you say, way less dirty than using ModifiedLayout.

geekosaur:
Right, but changing `magnify` is non-backward-compatible. That's why @slotThe wants a new combinator instead.

Zawaken:
Aha, that makes more sense then, thanks.

I will see if I can figure out some better names for the combinators and create a PR for these changes.

slotThe:
Closed by #822

PR #822, opened 2023-08-01
Add magnifiyxy combinator to XMonad.Layout.Magnifier

Zawaken:
### Description

behind them.

I've added another general combinator to XMonad.Layout.Magnifier, that allows setting up custom size both horizontally and vertically. `magnifyxy` is the new combinator, and the new implementations of magnifiers are `magnifierxy`, `magnifierxy'`, `magnifierxyOff`, and `magnifierxyOff'`.

Not sure if I am 100% on the names of the combinators.

This was mostly done to make it possible to make a `maximizeVertical` that defaults to `On`.
```hs
magnifierxy 1 1000 (Tall 1 (3/100) (1/2))
```

### Checklist

        manually, ...) and concluded: The combinators work, and the MagnifyMore+MagnifyLess keybinds still work with magnifierxy.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -119,6 +119,13 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Layout.Magnifier`
+
+    - Added `magnifyxy` to allow for different magnification in the
+      horizontal and vertical directions. Added `magnifierxy`,
+      `magnifierxy'`, `magnifierxyOff`, and `magnifierxyOff'` as
+      particular combinators.
+
   * `XMonad.Util.Loggers`
 
     - Added `logClassname`, `logClassnames`, `logClassnames'`,
--- a/XMonad/Layout/Magnifier.hs
+++ b/XMonad/Layout/Magnifier.hs
@@ -26,12 +26,15 @@ module XMonad.Layout.Magnifier
 
       -- * General combinators
       magnify,
+      magnifyxy,
 
       -- * Magnify Everything
       magnifier,
       magnifierOff,
       magnifiercz,
       magnifierczOff,
+      magnifierxy,
+      magnifierxyOff,
       maxMagnifierOff,
       maximizeVertical,
 
@@ -39,6 +42,8 @@ module XMonad.Layout.Magnifier
       magnifier',
       magnifiercz',
       magnifierczOff',
+      magnifierxy',
+      magnifierxyOff',
 
       -- * Messages and Types
       MagnifyMsg (..),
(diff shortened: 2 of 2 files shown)


geekosaur:
FWIW when making a modified version of an existing combinator, the Haskell convention is to append a prime `'` to the name. There are several examples already in the module.

Zawaken:
Yeah, I saw [`magnify'` was already used](https://github.com/Zawaken/xmonad-contrib/blob/ed1bd57c55c7599adfc3e5ea2b95b55491be20e1/XMonad/Layout/Magnifier.hs#L234-L239), so not sure what to then call it.

If you have any idea that is better than what I have put here, I would gladly take them. 😄

CHANGES.md
```diff
@@ -119,6 +119,10 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Layout.Magnifier`
+    - Added `magnifyxy` to allow for custom magnification in both directions,
```
slotThe, line 123:
```suggestion
  * `XMonad.Layout.Magnifier`
  
    - Added `magnifyxy` to allow for custom magnification in both directions,
```

XMonad/Layout/Magnifier.hs
```diff
@@ -119,6 +124,17 @@ magnify
 magnify cz mt start = ModifiedLayout $
     Mag 1 (fromRational cz, fromRational cz) (bool Off On start) mt
 
+magnifyxy
```
slotThe, line 127: Maybe add some top-level docs here along the lines of "Like 'magnify', but with the ability to specify different amounts of horizontal and vertical magnification"

slotThe (approved):
> Yeah, I saw [`magnify'` was already used](https://github.com/Zawaken/xmonad-contrib/blob/ed1bd57c55c7599adfc3e5ea2b95b55491be20e1/XMonad/Layout/Magnifier.hs#L234-L239), so not sure what to then call it.

That's internal, though, so there's no harm in renaming this here. However, FWIW, I like `magnifyxy` (or some variant thereof) much more than `magnify'`, because the former already tells you a bit about what it does.

slotThe:
(If you want, I can also make these changes, and squash everything together nicely; just say so)

XMonad/Layout/Magnifier.hs
Zawaken: Is top-level docs like how it is for "magnify"?

As in `-- | Like magnify.....`?

Zawaken:
Yeah I'd love that @slotThe 😄

slotThe:
Thank you for the contribution! :tada:

PR #825, opened 2023-08-23
New module: X.L.ConditionalLayout

slotThe:
Supersedes https://github.com/xmonad/xmonad-contrib/pull/582

What I haven't done, yet we probably should (not necessarily in this PR):

 - Rewrite `IfMax`, `PreScreen`, and `PerWorkspace` with this, which would close https://github.com/xmonad/xmonad-contrib/issues/75
 - Provide some QOL functions so that it's easier to write predicates. E.g., I reckon lots of people will want to use a variant of `gets $ find ((t ==) . W.tag) . W.workspaces . windowset`.

### [New module: X.L.ConditionalLayout](59054839f17a5fbb186f7e076bc7b403e1687b93)

Provide conditional variants of 'ModifiedLayout' and 'Choose', so that modifications (specific layouts) are only applied when a particular condition is met.

Co-authored-by: Tomas Janousek <tomi@nomi.cz>
Co-authored-by: Ivan Malison <IvanMalison@gmail.com>

-------

### Description

behind them.

### Checklist

        manually, ...) and concluded: cursory testing on my computer yielded the desired results, though someone else should look at this.

### DIFF
--- a/XMonad/Layout/ConditionalLayout.hs
+++ b/XMonad/Layout/ConditionalLayout.hs
@@ -1,0 +1,191 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StrictData #-}
+
+{- |
+   Module       : XMonad.Layout.ConditionalLayout
+   Description  : Conditionally apply layout and layout modifiers.
+   Copyright    : (c) Ivan Malison   <IvanMalison@gmail.com>,
+                      Tomas Janousek <tomi@nomi.cz>
+   License      : BSD
+   Maintainer   : Tony Zorman <soliditsallgood@mailbox.org>
+
+   This module provides conditional variants of 'ModifiedLayout' and 'Choose',
+   so that modifications (specific layouts) are only applied when a particular
+   condition is met.
+-}
+module XMonad.Layout.ConditionalLayout (
+  -- * Usage
+  -- $usage
+
+  -- * Combinators
+  condChoose,
+  conditional,
+
+  -- * Building Conditions
+  ModifierCondition (shouldApply),
+) where
+
+import XMonad hiding (hide)
+import XMonad.Layout.LayoutModifier
+import XMonad.Prelude
+import XMonad.StackSet (Workspace)
+import qualified XMonad.StackSet as W
+
+{- $usage
+
+You can use this module by importing it in your @xmonad.hs@
+
+> import XMonad.Layout.ConditionalLayout
+
+and writing an appropriate condition. Then, when 'conditional' is used,
+a layout modifier will be applied whenever the condition is true.
+Alternatively, 'condChoose' can decide which of two given layouts is to
+be applied.
+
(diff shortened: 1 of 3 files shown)


XMonad/Layout/ConditionalLayout.hs
```diff
@@ -0,0 +1,191 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StrictData #-}
```
geekosaur, line 7: Is this really necessary? I generally regard `StrictData` as cargo-culting instead of understanding laziness.

XMonad/Layout/ConditionalLayout.hs
```diff
@@ -0,0 +1,191 @@
...
+>
+> main :: IO ()
+> main = xmonad $ def
+>   { terminal   = "urxvt"
+>   , layoutHook = conditional IfMax2 (addTabsAlways shrinkText def) $ layoutHook def
+>   }
+
+alternatively, to conditionally switch between two layouts:
+
+>                  -- Full for workspaces with more than 4 windows, Tall otherwise.
```
geekosaur, line 82: Isn't this "more than two"? Or is the condition applied twice somehow?

XMonad/Layout/ConditionalLayout.hs
slotThe: I mean, I don't see any reason *not* to have either of

``` haskell
data CondModifiedLayout c m l a = CondModifiedLayout Bool c (ModifiedLayout m l a)
data CondChoose         c l r a = CondChoose         Bool c (Choose l r a)
```

with strict fields. If you want, I can also add explicit annotations instead

XMonad/Layout/ConditionalLayout.hs
slotThe: Ah, good catch, this must still be a remnant from an earlier version

XMonad/Layout/ConditionalLayout.hs
geekosaur: I'm just deeply suspicious of `StrictData`. You can talk to `c_wraith` in `#haskell` for a good rundown of why it's generally not a good idea.

geekosaur: There is also that one now has to check three different places to determine strictness: strictness annotations, `StrictData` at the top of the file, `StrictData` in `default-extensions` of the Cabal file. I strongly prefer they be where they're used instead of acting at a distance.

slotThe:
There's a but that I can't figure out right now: Assuming something like

``` haskell
myLayout = onWorkspace "4" (tall ||| Full) $ Full ||| tall ||| Mirror tall
 where
  tall = Tall 1 (3/100) (1/2)
```

one needs to issue five `M-SPC` calls to cycle through all layouts, regardless which branch one is on right now (some of those `M-SPC` invocations will just not do anything—at least not visibly).

Haven't tried debugging this yet, as its getting late, but perhaps someone else immediately knows what's up.

geekosaur:
Is using `Choose` internally wise? That's going to add a couple extra mod-space-s and I'm wondering how that interacts with the explicit condition (wouldn't it effectively invert it?).

geekosaur:
Also it may be better to leave `PerScreen` alone until we decide what's happening with #813.

Issue #831, opened 2023-09-19
`XMonad.Prompt` switch to previous completion


l29ah:
Would be cool if it allowed things like hitting Shift-Tab for a previous completion, like it now accepts Tab to go to the next one. Otherwise it becomes tedious when you miss the completion you need by tabbing excessively.

PR #834, opened 2023-10-03
X.Prompt: Allow for backwards cycling of completions

slotThe:
Fixes: https://github.com/xmonad/xmonad-contrib/issues/831

+ Add a new `prevCompletionKey` to `XPConfig`, in order to cycle backwards.
  Bound to `S-<Tab>` by default.

+ Already handle null keystrings (times when only a modifier was
  pressed) in `handleMain`, such that completions aren't cleared
  prematurely.

+ Augment `nextComplIndex` (now `computeComplIndex`) with the ability to go
  in an arbitrary 1-dimensional direction. As a result, that function,
  as well as `handleCompletion` and `handleCompletionMain` now take an
  additional `Direction1D` argument.

Based on: https://github.com/solomon-b/xmonad-contrib/tree/feature/scrolling-prompt-completions

Co-authored-by: Solomon Bothwell <ssbothwell@gmail.com>

### Checklist

        manually, ...) and concluded: manually confirmed for this to work, but @l29ah should double check

### DIFF
--- a/XMonad/Prompt.hs
+++ b/XMonad/Prompt.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE MultiWayIf #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Prompt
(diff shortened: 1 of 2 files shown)


XMonad/Prompt.hs
```diff
@@ -1,11 +1,12 @@
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TupleSections #-}
-{-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE ExistentialQuantification #-}
```
geekosaur, line 2: I'd be careful about alpha-sorting these given that ghc can add or remove pragmas based on other pragmas. In other words, order can matter.

geekosaur (approved):

XMonad/Prompt.hs
slotThe: Can it really? I don't think adding pragmas would cause any problems, but removing sounds problematic indeed. I've never heard that language extensions have these capabilities, though, do you have an example?

geekosaur: The GHC manual describes this as "implies". See for example [Type families](https://downloads.haskell.org/ghc/9.4.7/docs/users_guide/exts/type_families.html#type-families), which says:

Implies: MonoLocalBinds, KindSignatures, ExplicitNamespaces

Off the top of my head I can't think of any that get turned off by other extensions, but it's certainly possible so some care needs to be given to ordering as an extension may thereby get ignored or conversely might lead to odd results if one normally turned off gets turned on later.

slotThe: I knew about the "implies" stuff, but that should be irrelevant of the order (and we're not turning anything off here). Anyways, I will leave the old sorting in place for now, only adding the new extension to the bottom

geekosaur: Sadly, it's not independent of the order and things like turning `MonoLocalBinds` back off after enabling `TypeFamilies` are known idioms.

PR #837, opened 2023-10-22
introduce single active scratchpad hook

philib:
### Description

I added a logHook that ensures that only the most recent scratchpad is active on the workspace at any given time. When opening a new scratchpad, the previous scratchpad is moved to the corresponding scratchpad workspace.

Motivation: I use floating scratchpad to open terminals, notes, etc. quickly. Since `nsHideOnFocusLoss` hides scratchpad more often, then i intended to, the scratchpads can pile up very fast, which leads to the fact that I have to "dismantle" the stack of scratchpads manually. To prevent this, I now use this logHook. I thought I'd share it in case anyone has a similar workflow and is also interested.

I am a Haskell beginner, so please bear with me. I would be glad about feedback.

### Checklist

        manually, ...) and concluded:  manually tested

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -83,6 +83,11 @@
       ordered lexicographically, as before.  Currently focused window will
       always be the topmost, meaning the last in the list.
 
+  * `XMonad.Util.NamedScratchpad`
+
+    - Added `nsSingleScratchpadPerWorkspace`—a logHook to allow only one
+      active scratchpad per workspace.
+
 ### New Modules
 
   * `XMonad.Layout.CenterMainFluid`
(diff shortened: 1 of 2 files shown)


slotThe (commented):
Thank you for the contribution! May I ask how this is different from `nsHideOnFocusLoss`? To me, it seems that

> only the most recent scratchpad is active on the workspace at any given time. When opening a new scratchpad, the previous scratchpad is moved to the corresponding scratchpad workspace.

is describing exactly the same functionality

slotThe:
Thank you for the contribution :tada:

Issue #830, opened 2023-09-17
New `-Wx-partial` warnings with ghc 9.8


geekosaur:
### Problem Description

ghc 9.8 produces a great many warnings about partial functions, mostly `head` and `tail`.

### Steps to Reproduce

Build `xmonad-contrib` with ghc 9.8 alpha3. (I will repeat this with alpha4 when it is available.)
This currently requires a patched `setlocale` package and an `allow-newer` on `hsc2hs`, which I am taking up with the respective upstreams.

A possibly useful `cabal.project` is:
```
packages: */*.cabal
-- setlocale and xmonad-contrib packages
with-compiler: ghc-9.8
allow-newer: hsc2hs:base
```

### Configuration File

Please include the smallest _full_ configuration file that reproduces
the problem you are experiencing:

```haskell
module Main (main) where

import XMonad

main :: IO ()
main = xmonad def
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17.2
    - [x] With `xmonad-contrib` version 0.17.1.9 (commit 51ba23b if using git)

geekosaur:
```
XMonad/Layout/Grid.hs:64:69: warning: [GHC-63394] [-Wx-partial]
    In the use of ‘tail’
    (imported from Prelude, but defined in GHC.List):
    "This is a partial function, it throws an error on empty lists. Replace it with drop 1, or use pattern matching or Data.List.uncons instead. Consider refactoring to use Data.List.NonEmpty."
   |
64 |     chop n m = ((0, m - k * fromIntegral (pred n)) :) . map (, k) . tail . reverse . take n . tail . iterate (subtract k') $ m'
   |                                                                     ^^^^

XMonad/Layout/Grid.hs:64:95: warning: [GHC-63394] [-Wx-partial]
    In the use of ‘tail’
    (imported from Prelude, but defined in GHC.List):
    "This is a partial function, it throws an error on empty lists. Replace it with drop 1, or use pattern matching or Data.List.uncons instead. Consider refactoring to use Data.List.NonEmpty."
   |
64 |     chop n m = ((0, m - k * fromIntegral (pred n)) :) . map (, k) . tail . reverse . take n . tail . iterate (subtract k') $ m'
   |                                                                                               ^^^^

[...]

PR #838, opened 2023-10-24
ci: Regenerate haskell-ci

liskin:
(automatic PR by https://github.com/liskin/xmonad-maintenance)


slotThe:
We probably only want this when all warnings from #830 are fixed

liskin:
Ok, keep it open then. The scheduled workflow will probably forcepush into it every week, let's see then how much noise it creates.

slotThe:
Hooray :tada:

PR #843, opened 2023-11-16
deprecate `XMonad.Util.Ungrab`

iogrt:
### Description

Deprecate `XMonad.Util.Ungrab` as it is going to be moved to `XMonad.Operations` in https://github.com/xmonad/xmonad/pull/479

### Checklist

        manually, ...) and concluded: Tested manually with a config that uses unGrab and removed the XMonad.Util.Ungrab import.

### DIFF
--- a/XMonad/Config/Mate.hs
+++ b/XMonad/Config/Mate.hs
@@ -1,5 +1,7 @@
 {-# OPTIONS_GHC -fno-warn-missing-signatures #-}
-
+-- TODO: Remove when we depend on a version of xmonad that has unGrab.
+{-# OPTIONS_GHC -Wno-deprecations  #-}
+{-# OPTIONS_GHC -Wno-dodgy-imports #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module       : XMonad.Config.Mate
@@ -28,15 +30,14 @@ module XMonad.Config.Mate (
     desktopLayoutModifiers
     ) where
 
-import XMonad
-import XMonad.Config.Desktop
-import XMonad.Util.Run (safeSpawn)
-import XMonad.Util.Ungrab
-import XMonad.Prelude (toUpper)
-
+import System.Environment (getEnvironment)
 import qualified Data.Map as M
 
-import System.Environment (getEnvironment)
+import XMonad hiding (unGrab)
+import XMonad.Config.Desktop
+import XMonad.Prelude (toUpper)
+import XMonad.Util.Run (safeSpawn)
+import XMonad.Util.Ungrab (unGrab)
 
 -- $usage
 -- To use this module, start with the following @~\/.xmonad\/xmonad.hs@:
--- a/XMonad/Util/Ungrab.hs
+++ b/XMonad/Util/Ungrab.hs
@@ -13,7 +13,7 @@
 --
 -----------------------------------------------------------------------------
 
-module XMonad.Util.Ungrab
+module XMonad.Util.Ungrab {-# DEPRECATED "Use XMonad.Operations.unGrab instead" #-}
     ( -- * Usage:
       -- $usage
       unGrab
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -93,6 +93,9 @@
     - The function `readKeySequence` now returns a non-empty list if it
       succeeded.
 
+  * Deprecate `XMonad.Util.Ungrab`; it was moved to `XMonad.Operations`
+    in core.
+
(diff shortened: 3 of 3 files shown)


geekosaur (approved):

XMonad/Util/Ungrab.hs
```diff
@@ -42,4 +40,4 @@ import XMonad.Core
 
 -- | Release xmonad's keyboard grab, so other grabbers can do their thing.
 unGrab :: X ()
-unGrab = withDisplay $ \d -> io (ungrabKeyboard d currentTime >> ungrabPointer d currentTime >> sync d False)
+unGrab = Operations.unGrab
```
slotThe, line 43: I think we should keep the old definition for now. Otherwise, we'd have to bump `xmonad`s version, and specify that as a minimum version in contrib's cabal file. To me, that seems like quite a bit of effort for a change like this

XMonad/Config/Mate.hs
```diff
@@ -31,7 +31,6 @@ module XMonad.Config.Mate (
 import XMonad
 import XMonad.Config.Desktop
 import XMonad.Util.Run (safeSpawn)
-import XMonad.Util.Ungrab
```
slotThe, line 34: This will result in a build failure related to [this](https://github.com/xmonad/xmonad-contrib/pull/843#discussion_r1396790444). I'm not sure it's worth bumping our bounds over, so we should probably be a bit more tame here as well

iogrt: What I found is that if I don't remove this line, the compiler will complain about an ambiguous reference in the new version. Because `unGrab` will be defined both in `XMonad.Operations` (which is imported into this file from `import XMonad`) and `XMonad.Util.Ungrab`.

But thinking about it again. I could probably just do `import XMonad hiding (unGrab)`. Would that be preferrable?

slotThe: Yes, I was thinking about hiding the function from the `XMonad` import and perhaps adding a `{-# OPTIONS_GHC -Wno-dodgy-imports #-}` at the top of the file, to pacify the compiler when this is compiled against xmonad HEAD

slotThe:
Thanks!

XMonad/Config/Mate.hs
liskin: I'd be in favor of bumping xmonad and having `#if MIN_VERSION_xmonad(0, 17, 3)` in X.U.Ungrab so that users following the stable releases don't get the `Ambiguous occurrence ‘unGrab’` error when compiling their configs, and just get a deprecaton warning.

liskin: (Note that for that to work, we can't have `unGrab = Operations.unGrab` but must reexport the import.)

liskin: Hm, thinking a bit more about this, might as well bump the dependency instead of introducing an `#if`… :-)

slotThe: > I'd be in favor of bumping xmonad and having `#if MIN_VERSION_xmonad(0, 17, 9)` in X.U.Ungrab so that users following the stable releases don't get the `Ambiguous occurrence ‘unGrab’` error when compiling their configs, and just get a deprecaton warning.

Right now users compiling against `xmonad-0.17.2` should not get an error (as evidenced by the CI), since we're hiding (the possible non-existent) `unGrab` from the `XMonad` import. Still, bumping might not be such a bad idea

liskin: I wasn't talking about compiling xmonad-contrib itself, but about compiling the xmonad.hs config. Anyone who imports both `XMonad` and `XMonad.Util.Ungrab` will start getting errors. I mean, yeah, it is mentioned in "Breaking Changes", but I don't think it should actually be a _Breaking Change_ when we can turn it into just a deprecation warning.

liskin: (https://github.com/xmonad/xmonad/commit/5c6c3d4d71dc28502ffbcf3a7fddad64c6b3d8b8 & https://github.com/liskin/xmonad-contrib/commit/8526d935066e7de5dd5cadd7afca208d7e11b754 are one way to solve this, if we want an ifdef)

slotThe: Oh, right, that might indeed be an issue. The ifdef sounds fine in that case

liskin: Okay, I pushed the two commits then.

PR #848, opened 2023-12-11
Add new module: X.L.CircleEx: advanced version of Circle layout

portnov:
### Description

X.L.Circle layout is somewhat nice, but it does not provide much possibilities for customization. This adds a new module, X.L.CircleEx, which is similar to X.L.Circle, but has some new features:
* Number of "master" windows can be adjusted. For example, with 0 master windows, all windows go in a circle, nothing in the center.
* It's possible to configure sizes of master window and slave window (as a portion of screen size).
* Size of each next slave window is slightly less than size of previous (configurable)
* It's possible to rotate whole setup clockwise / counterclockwise. For example, by default first slave window opens near right edge of the screen, but I like it to open near left top corner.

### Checklist

  - [x] I've done extensive manual testing by using it for a month :)

### DIFF
--- a/XMonad/Layout/Circle.hs
+++ b/XMonad/Layout/Circle.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleInstances, MultiParamTypeClasses #-}
+{-# LANGUAGE PatternSynonyms #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -15,15 +16,14 @@
 --
 -----------------------------------------------------------------------------
 
-module XMonad.Layout.Circle (
-                             -- * Usage
-                             -- $usage
-                             Circle (..)
-                            ) where -- actually it's an ellipse
+module XMonad.Layout.Circle {-# DEPRECATED "Use XMonad.Layout.CircleEx instead" #-}
+  ( -- * Usage
+    -- $usage
+    pattern Circle
+  ) where -- actually it's an ellipse
 
-import XMonad.Prelude
-import XMonad
-import XMonad.StackSet (integrate, peek)
+import GHC.Real (Ratio(..))
+import XMonad.Layout.CircleEx
 
 -- $usage
 -- You can use this module with the following in your @~\/.xmonad\/xmonad.hs@:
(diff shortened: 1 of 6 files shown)


XMonad/Layout/CircleEx.hs
```diff
@@ -0,0 +1,161 @@
...
+
+import Data.Ratio
+
+import XMonad.Prelude
+import XMonad
+import qualified XMonad.StackSet as W
+
+-- $usage
+-- The layout puts first N windows (called master) to the center of screen. 
+-- Others (called slave) are organized in a circle (well, ellipse). Size of
```
slotThe, line 36: I'd `s/slave/stack` throughout the whole module. (I've been wanting to do this across contrib for a while, but haven't gotten around to it)

slotThe (approved):
Based on a cursory read, this look fine, thanks! :+1: 

Since you say that this essentially supersedes X.L.Circle, would it be possible to (i) implement that in terms of this new module, and (ii) perhaps deprecate X.L.Circle (if that's desired)?

XMonad/Layout/CircleEx.hs
portnov: Hm, I'm okay, but the word "stack" does not seem to describe "other-than-master" windows to me. As far as I understand, in xmonad stack are all windows except floating ones.
Maybe "secondary" instead?

portnov:
> (i) implement that in terms of this new module, and (ii) perhaps deprecate X.L.Circle (if that's desired)?

Ok. For (i), I added a variant of default settings which mimics X.L.Circle's hardcoded settings.

XMonad/Layout/CircleEx.hs
slotThe: > As far as I understand, in xmonad stack are all windows except floating ones.

This is technically true, but at this point "stack windows" seems to be synonymous with "non-master windows" in contrib. I don't have any strong opiniont on this, though, "secondary" sounds fine as well

XMonad/Layout/Circle.hs
```diff
@@ -15,11 +15,11 @@
 --
 -----------------------------------------------------------------------------
 
-module XMonad.Layout.Circle (
-                             -- * Usage
-                             -- $usage
-                             Circle (..)
-                            ) where -- actually it's an ellipse
+module XMonad.Layout.Circle {-# DEPRECATED "Use XMonad.Layout.CircleEx instead" #-}
```
slotThe, line 18: If you export `CircleEx`, then (I think) this whole module could be replaced by something like

``` haskell
pattern Circle :: CircleEx a
pattern Circle = CircleEx 1 (70 % 99) (2 % 5) 1 0
```

In this way we wouldn't duplicate the implementation details

TheMC47 (commented):
Thanks for the contribution, just came by to say that a screenshot would be great :smile:

portnov:
@TheMC47 
Screenshot_20231218_152452

XMonad/Layout/Circle.hs
portnov: That gives me

```
/home/portnov/src/xmonad-contrib-portnov/XMonad/Layout/Circle.hs:43:40: error: Parse error in pattern: 2 % 5
   |
43 | pattern Circle = CircleEx 1 (70 % 99) (2%5) 1 0
   |                                        ^^^
```

I suspect `pattern` thing is only applicable to literals, not expressions.

slotThe: Ah, right. In that case I guess `GHC.Real.(:%)` should work?

geekosaur: Try it with the extra spaces; ghc since 9.2 requires spaces around infix operators, as part of rationalizing parsing of `OverloadedRecordDot`, `BangPatterns`, `NumericLiterals`, and various other special forms.

portnov:
@slotThe 
ok, importing (:%) from GHC.Real did the trick. I'm not very happy with exporting data constructor, because it will complicate things if / when it will be required to add fields to data type. But probably you're right and not having a duplicate of logic overweights that potential problem.

Now, when I deprecated X.L.Circle module, I see some number of deprecation warnings, since it was used in several places. I think I should also replace Circle with CircleEx in these places.

slotThe:
> Now, when I deprecated X.L.Circle module, I see some number of deprecation warnings, since it was used in several places. I think I should also replace Circle with CircleEx in these places.

Sounds good :+1:

portnov:
Done. Please review.

slotThe:
Thanks! (I just pushed some doc fixes)

Issue #845, opened 2023-11-26
Part of keymap is ignored in XPrompt


dunric:
### Problem Description

When invoking [shellPrompt](https://xmonad.github.io/xmonad-docs/xmonad-contrib/src/XMonad.Prompt.Shell.html#shellPrompt) dialog with `promptKeymap` config field set f.E. to [emacsKeymap](https://xmonad.github.io/xmonad-docs/xmonad-contrib/src/XMonad.Prompt.html#emacsLikeXPKeymap%27), all keys *without a modifier*(equal to 0) are ignored:

``` haskell
  ++
  map (first $ (,) 0) -- <key>
  [ (xK_Return, acceptSelection)
  , (xK_KP_Enter, acceptSelection)
  , (xK_BackSpace, deleteString Prev)
  , (xK_Delete, deleteString Next)
  , (xK_Left, moveCursor Prev)
  , (xK_Right, moveCursor Next)
  , (xK_Home, startOfLine)
  , (xK_End, endOfLine)
  , (xK_Down, moveHistory W.focusUp')
  , (xK_Up, moveHistory W.focusDown')
  , (xK_Escape, quit)
  ]
```

### Steps to Reproduce

 1. modify XConfig for XPrompt to use emacs keymap
     `myPromptTheme = def { promptKeymap = emacsLikeXPKeymap }`
 2. assign XPrompt action to some keybinding
     ` ("M-<Escape>",      shellPrompt myPromptTheme)`
 3. invoke shellPrompt with above keybinding
 4. attempt to use any of keys without a modifier (ie. without controlMask, Mod1Mask) like Up, Down, Home, Right do not work
   
### Configuration File

```haskell
module Main (main) where

import XMonad
import XMonad.Prompt
import XMonad.Prompt.Shell
import XMonad.Util.EZConfig

main :: IO ()
main = xmonad $ def
  { modMask = mod4Mask
  }
  `additionalKeysP`
  myKeymap

myKeymap =
  [  ("M-<Escape>",      shellPrompt myPromptTheme)
  ]

myPromptTheme = def { promptKeymap = emacsLikeXPKeymap }
```
  - I tested my configuration
    - [ X] With `xmonad` version 0.17.2.9 (commit XXX if using git)
    - [ X] With `xmonad-contrib` version git-20231011 (commit XXX if using git)

EDIT: config file

geekosaur:
[...]

PR #850, opened 2023-12-16
X.Prompt: Execute keypress when it has an action associated to it

slotThe:
### Description

It might be that the keypress does not have a valid stroke associated to it, but is still bound to an action (e.g., xK_Left an friends). In this case, we still want to execute it.

Closes: https://github.com/xmonad/xmonad-contrib/issues/845

### Checklist

        manually, ...) and concluded: quick manual test


dunric:
Seems ok to me. Fix does also work in my complex setup.

portnov:
This fixed the problem for me too :)

slotThe:
Great, thanks to you two!

PR #856, opened 2023-12-19
X.Prompt: handle Return key in vim normal mode

l29ah:
### Description

Avoiding RAGE when one can't just use Return key in the normal mode like in vim.

### Checklist


slotThe:
Thanks!

PR #861, opened 2024-01-09
XMonad.Prompt: export getCurrentCompletions (and setCurrentCompletions)

dcousens:
### Description

I am playing with using the key bindings to modify the input, and when trying to change the input to the next [or previous] completion in the currently available completions, I needed `getCurrentCompletions` and `selectedCompletion` to be available.  I have additionally added `setCurrentCompletions`, but if we don't want to export that, I'm happy to remove that from the PR.

I am new to Haskell, maybe I'm missing another way to do this.

### Checklist

        manually, ...) and concluded: N/A


slotThe:
Thanks!

PR #864, opened 2024-01-18
X.P: Add escape hatch for preventing X.P IO

dcousens:
### Rationale

When using `XMonad.Prompt` for multiple (8+) different use cases in my configuration, I never once needed a command history.

The default of reading and writing a file for the command history opaque to the user is strange to me, and each prompt is sharing and replacing the same `prompt-history` file... but I digress.

My feature request is to stop that behavior, where `historySize = 0` didn't cut the mustard since `prompt-history` is still reading and writing on every prompt.

## ~Proposal~

~This pull request adds `historyRead` and `historyWrite` configuration options to `XPConfig`, allowing users to write something like:~

```haskell
  ...
  , historyRead = \cacheDir -> return empty
  , historyWrite = \history cacheDir -> return ()
  ...
``` 

~This could open up functionality for using other state for a command history, but maybe that isn't desired.
My main priority with this pull request is to prevent the reading and writing of the `prompt-history` file, and as such I am marking this as a draft until discussion resolves the best path forward.~

### Proposal (current)
`historySize = 0` should branch the IO behavior, and prevent any related IO (reading or writing).

### Checklist

        manually, ...) and concluded: unknown, I am using these changes locally

### DIFF
--- a/XMonad/Prompt.hs
+++ b/XMonad/Prompt.hs
@@ -562,7 +562,7 @@ mkXPromptImplementation historyKey conf om = do
   s <- gets $ screenRect . W.screenDetail . W.current . windowset
   cleanMask <- cleanKeyMask
   cachedir <- asks (cacheDir . directories)
-  hist <- io $ readHistory cachedir
+  hist <- io $ readHistory conf cachedir
   fs <- initXMF (font conf)
   let width = getWinWidth s (position conf)
   st' <- io $
@@ -582,7 +582,7 @@ mkXPromptImplementation historyKey conf om = do
   releaseXMF fs
   when (successful st') $ do
     let prune = take (historySize conf)
-    io $ writeHistory cachedir $
+    io $ writeHistory conf cachedir $
       M.insertWith
       (\xs ys -> prune . historyFilter conf $ xs ++ ys)
       historyKey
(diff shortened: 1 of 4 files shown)


XMonad/Prompt.hs
```diff
@@ -1803,7 +1810,8 @@ historyCompletionP p = do
     cd <- asks (cacheDir . directories)
     pure $ \x ->
         let toComplList = deleteConsecutive . filter (isInfixOf x) . M.foldr (++) []
-         in toComplList . M.filterWithKey (const . p) <$> readHistory cd
+--          in toComplList . M.filterWithKey (const . p) <$> (historyFetch conf) cd
+         in toComplList . M.filterWithKey (const . p) <$> (readHistory cd)
```
dcousens, line 1814: I don't know if `historyCompletionP` can be updated without a breaking change, which I am indicating here by still using `readHistory`

dcousens:
Maybe a better path forward is to remove the `prompt-history` writing behaviour by default in a breaking change, and then supporting users to opt-in by adding this kind of configuration?

XMonad/Prompt.hs
dcousens: The problem, fwiw, is that `conf` isn't provided to `historyCompletionP`

slotThe (commented):
Wouldn't the easiest thing be to just not write the history file if the history size is 0?

dcousens:
@slotThe `historyCompletionP` needs `conf` either way, but yes branching as I mentioned is an alternative approach
> historySize = 0 could branch the behavior, and prevent any related IO (reading or writing).

dcousens:
@slotThe I have changed the pull request to pattern match if `historySize = 0` for `readHistory` and `writeHistory`

slotThe (approved):
This looks fine to me, and is definitely the least intrusive change. Neither `readHistory`, nor `writeHistory` is exported, so this does not result in a breaking change for anyone.

dcousens:
@slotThe the only downside to this approach is that `historyCompletionP`, if used, will not respect the new behaviour (for better or worse)

slotThe:
Both `historyCompletion` and `historyCompletionP` aren't very widely used within contrib, so I would even be open for a breaking change there, perhaps making them take an `XPConfig` (or even just the `historySize`) as an additional argument. I think the number of (actually useful) custom prompts using these in the wild will be relatively low

slotThe:
@dcousens ping for ^

slotThe:
@dcousens sorry for only getting back to you now: I'm saying that I feel ambivalent about introducing a breaking change for `historyCompletion[P]`; these functions probably aren't very widely used, and the uses they have inside of contrib are all in the vicinity of an `XPConfig` argument. However, if you feel otherwise them I'm also fine with merging this as-is

dcousens:
OK, I'll have a shot at that, I'd prefer that personally

slotThe:
@dcousens friendly ping :)

dcousens:
Rebased and added a breaking commit, which ensures that `historyCompletion*` now accepts an `XPConfig`

slotThe:
Thank you!

Issue #820, opened 2023-07-22
Dialog window (zenity) not floated when ewmhFullscreen present


mrozigor:
### Problem Description

When `ewmhFullscreen` is present, then fullscreen apps (i.e. Firefox in fullscreen mode) works OK - by that I mean properly overlapping xmobar/temporarily removing bar from screen. But with these settings `zenity` dialogs don't work properly - they are not floated at all and are normally tiled. When I remove `ewmhFullscreen` from config, then dialogs works as expected - properly floated - but fullscreen apps don't (xmobar is always visible).

I'm not sure if this is my config mistake (probably) or something is wrong with xmonad. I'll be thankful for any help or documentation page that will help resolving the issue ;) Thanks.

### Steps to Reproduce

Run zenity dialog (e.g. `zenity --info --title "TITLE"`) with given config.

### Configuration File

```haskell
import XMonad
import XMonad.Hooks.DynamicLog
import XMonad.Hooks.EwmhDesktops
import XMonad.Hooks.ManageDocks
import XMonad.Hooks.ManageHelpers
import XMonad.Util.Run(spawnPipe)
import XMonad.Util.EZConfig(additionalKeys)
import XMonad.Hooks.SetWMName
import XMonad.Layout.NoBorders
import XMonad.Layout.IndependentScreens
import XMonad.Layout.Fullscreen
import qualified XMonad.StackSet as W
import System.IO
import Control.Monad
import Data.Maybe
import Data.List
import XMonad.Util.WorkspaceCompare (getSortByXineramaRule) 
import Data.Ratio
import XMonad.StackSet

-- Check if window is modal
checkModal :: Query Bool
checkModal = isInProperty "_NET_WM_STATE" "_NET_WM_STATE_MODAL"

main = do
  xmproc <- spawnPipe "xmobar"

  xmonad $ ewmhFullscreen . ewmh $ docks $ def
    { manageHook = manageDocks <+> composeAll
      [ isDialog --> doCenterFloat,
        checkModal --> doCenterFloat,
         --isFullscreen --> doFullFloat,
         title =? "Ediff" --> doCenterFloat,
         className =? "zenity" --> doFloat,
         transience' ]
    , layoutHook = smartBorders . avoidStruts $ Full ||| Tall 1 0.02 0.5 ||| Mirror (Tall 1 0.02 0.5) -- layoutHook def
    , handleEventHook = handleEventHook def
    , terminal = "urxvt"
[...]

Issue #865, opened 2024-01-18
`ewmhFullscreen` doesn't restore floating windows


geekosaur:
If a floating window is made fullscreen and then toggled back, it is sunk instead of floated. The code at https://github.com/xmonad/xmonad-contrib/blob/afd6824ce00063bb8e9b7a1c5daf0737c2f61616/XMonad/Hooks/EwmhDesktops.hs#L442C5-L444 unconditionally sinks the window when `_NET_WM_STATE_FULLSCREEN` is removed.

This will require saving the old window position for floats; the current code assumes sinking the window will handle positioning.

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17.2
    - [x] With `xmonad-contrib` version 0.17.1

geekosaur:
This could reasonably be parked until it matters to someone. Also, it's better not to try to address it until #866 goes in, since otherwise there will be a merge conflict.

liskin:
I believe https://github.com/xmonad/xmonad-contrib/pull/768 solved this already.

geekosaur:
"This commit only adds the hooks. Neither of the motivations are dealt with yet, so the docs are a bit terse still."

liskin:
Well that's the first commit. The following one adds the thing.

PR #866, opened 2024-01-18
only sink window if actually fullscreened

geekosaur:
### Description

`EwmhFullscreen` unconditionally sinks windows in response to a `_NET_WM_STATE` root window message removing `_NET_WM_STATE_FULLSCREEN`. Unfortunately, at least some versions of either the Gtk or GNOME libraries send this on startup while restoring a window's last known state, which means a `manageHook` `doFloat` will be undone.

This change ignores the remove if the window is not fullscreen.

A followup issue, #865, has been added for the follow-on problem that the floating state of a window is not restored on removal of the fullscreen state.

Closes: #820

### Checklist

        manually, ...) and concluded: a user is already using this change successfully

  - [x] I updated the `CHANGES.md` file (unnecessary)

### DIFF
--- a/XMonad/Hooks/EwmhDesktops.hs
+++ b/XMonad/Hooks/EwmhDesktops.hs
@@ -545,10 +545,10 @@ fullscreenEventHook'
       chWstate f = io $ changeProperty32 dpy win wmstate aTOM propModeReplace (f wstate)
 
   when (managed && typ == wmstate && fi fullsc `elem` dats) $ do
-    when (action == add || (action == toggle && not isFull)) $ do
+    when (not isFull && (action == add || action == toggle)) $ do
       chWstate (fi fullsc:)
       windows . appEndo =<< runQuery fullscreenHook win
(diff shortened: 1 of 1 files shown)


slotThe (approved):

XMonad/Hooks/EwmhDesktops.hs
```diff
@@ -548,7 +548,7 @@ fullscreenEventHook'
     when (action == add || (action == toggle && not isFull)) $ do
```
liskin, line 548: Hm, do we perhaps want to make the same change here to prevent windows from being fullscreened multiple times if a rogue app makes multiple add requests?

geekosaur: I left that one alone because it seems to be idempotent.

liskin: But we have no idea what `fullscreenHook` does so it's only idempotent with the default one.

PR #881, opened 2024-03-17
update time to <1.15

geekosaur:
### Description

Update bounds of `time` to allow the new 1.14 release.

### Checklist

        manually, ...) and concluded: build with `--constraint='time == 1.14'` passed locally

  - [ ] I updated the `CHANGES.md` file (n/a)

### DIFF
--- a/xmonad-contrib.cabal
+++ b/xmonad-contrib.cabal
@@ -59,7 +59,7 @@ library
                    containers >= 0.5 && < 0.8,
                    directory,
                    filepath,
-                   time >= 1.8 && < 1.13,
+                   time >= 1.8 && < 1.15,
                    process,
                    random,
                    mtl >= 1 && < 3,


geekosaur:
Actually, since `time` is a bootlib CI is still using 1.12.2. Have to see how to test it locally.

liskin:
I believe @slotThe did the testing a couple times in the past when I pinged him on IRC after the Packdeps CI failed.
Perhaps we should add a note about this to MAINTAINERS.md so we don't need to remember (and also to decrease the bus factor).

slotThe:
I saw the 

> build with --constraint='time == 1.14' passed locally

edit and figured that the local tests ran through.

> We could even make the Packdeps job point us to the relevant piece of documentation to minimise the brain function required to deal with this :-D

We could also look into using https://github.com/nomeata/cabal-plan-bounds to generate version constraints automatically, or https://github.com/nomeata/cabal-force-upper-bound to force CI to use the upper bound of a package

geekosaur:
> edit and figured that the local tests ran through.

They did, and I updated to show it. Should I have deleted that comment afterward? (or edited it, I guess)

PR #887, opened 2024-05-04
Add a new layout Columns

jecaro:
### Description

This layout organizes windows in columns and allows to move/resize them in every direction. This layout is showcased here: https://jeancharles.quillet.org/posts/2024-05-02-Columns-a-new-XMonad-layout.html

I was reluctant to send this layout in a PR because I'm not sure I'd have time to maintain it outside my main configuration. But I've been told that some people might be volunteers to maintain it :slightly_smiling_face:  So here it is.

Something that might be annoying is that the parameter to toggle between wide screen and small screen is currently hardcoded. As the layout lives in my config, this is not a problem for me but it might be for others. I can make it configurable if needed tho.

### Checklist

  - [X] I've considered how to best test these changes (property, unit, manually, ...) and concluded: I've been using this layout for more than one year now. It works fine for my usage. It has a few annoyances sometimes when using a tabbed sublayout but despite that it works great.

### DIFF
--- a/XMonad/Layout/Columns.hs
+++ b/XMonad/Layout/Columns.hs
@@ -1,0 +1,478 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
+
+-- |
+-- Module: XMonad.Layout.Columns
+-- Description: A layout which tiles the windows in columns.
+-- Copyright: Jean-Charles Quillet
+-- License: BSD-style (see LICENSE)
+--
+-- Maintainer: none
+-- Stability: unstable
+-- Portability: unportable
+--
+-- A layout which tiles the windows in columns. The windows can be moved and
+-- resized in every directions.
+--
+-- The first window appears:
+--
+-- * in the center on wide screens
+-- * fullscreen otherwise
+--
+-- The second window appears on a second column.
+--
+-- Subsequent windows appear on the bottom of the last columns.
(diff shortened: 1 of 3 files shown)


XMonad/Layout/Columns.hs
```diff
@@ -0,0 +1,478 @@
...
+  doLayout (Columns columns) rectangle stack =
+    pure (rectangles, Just (Columns columns'))
+    where
+      hackedColumns = hackForTabs columns stack
+      columns' = updateWindowList hackedColumns stack
+      rectangles = toRectangles rectangle' columns'
+      -- If there is only one window and the screen is big, we reduce the
+      -- destination rectangle to put the window on the center of the screen.
+      rectangle'
+        | rect_width rectangle > 2000 && (length . toList $ stack) == 1 =
```
geekosaur, line 176: I would agree that this should be configurable. You may also want to look into [XMonad.Layout.PerScreen](https://hackage.haskell.org/package/xmonad-contrib-0.18.0/docs/XMonad-Layout-PerScreen.html) (but see #813); perhaps there's a way to farm this decision out to users' configs, allowing them to decide if they even want the behavior.

geekosaur (changes requested):
I'm sorry for the delay; my network's been out for two weeks.

XMonad/Layout/Columns.hs
jecaro: Wow. Two weeks with a broken network. Sounds terrible :fearful: 

I wasn't aware of that `PerScreen` module. Thx. Indeed, with it, that feature becomes irrelevant. 

I've made the width of the first column configurable. 

I can now reproduce the old behavior using `PerScreen` with two instances of `Columns` and a different value for that first column width. That's fine for my use case.

geekosaur (approved):

PR #892, opened 2024-05-26
X.A.Repeatable: Auto-detect modifier keys from currentEvent

liskin:
### Description

This makes all the cycleSomething key bindings easier to use: users don't need to manually list the KeySyms of modifier keys in addition to specifying the mask for the keybinding itself. We can easily get those KeySyms from the KeyEvent currently being handled with the help of XGetModifierMapping.

**This only implements the logic in `X.A.Repeatable`.** We need to decide
whether to change the API of all the user-facing functions (drop the `mods`
argument), just document that empty list means "auto-detect", or some
combination thereof.

### Checklist

        manually, ...) and concluded:  
        seems to work fine for me…  
        automated testing not feasible

### DIFF
--- a/XMonad/Actions/Repeatable.hs
+++ b/XMonad/Actions/Repeatable.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Actions.Repeatable
@@ -23,6 +24,8 @@ module XMonad.Actions.Repeatable
   , repeatableM
   ) where
 
+import Data.Bits
+
 -- mtl
 import Control.Monad.State (StateT(..))
 
@@ -31,6 +34,7 @@ import Graphics.X11.Xlib.Extras
 
 -- xmonad
 import XMonad
+import XMonad.Prelude
 
 
 -- | An action that temporarily usurps and responds to key press/release events,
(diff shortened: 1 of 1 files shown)


liskin:
Anyone happens to have an opinion about the API?

slotThe:
This sounds much better than the current situation, but I guess breaking user configs by dropping an argument is always a bit "eh" :/

geekosaur:
New API and mark the old one as deprecated? Or just go with "empty list means get it from the event"?
Also, do we want to think about this for https://hackage.haskell.org/package/xmonad-contrib-0.18.0/docs/XMonad-Actions-CycleWindows.html#v:cycleRecentWindows?

liskin:
> Also, do we want to think about this for https://hackage.haskell.org/package/xmonad-contrib-0.18.0/docs/XMonad-Actions-CycleWindows.html#v:cycleRecentWindows?

X.A.Repeatable is used in X.A.CycleWindows, and also in X.A.{CycleWorkspaceByScreen,CycleRecentWS,MostRecentlyUsed} so changing the API would result in quite a lot of breaking changes, or at least quite a lot of new function names and lot of deprecations.

I guess going with the "empty list …" option is better then. Or something.

LSLeary:
Re API, the functions in `X.A.Repeatable` are mostly not used directly in config, but indirectly via client contrib modules.
As such, the `mods` argument here should just be removed entirely; let the clients handle the change.

PR #894, opened 2024-06-09
XMonad.Actions.GridSelect: added gs_cancelOnEmptyClick field

sylecn:
### Description

In the original code, when a GridSelect is shown, user has to use a keyboard to cancel it (ESC key by default). With this field added, when it is set to True, mouse click on empty space can cancel the GridSelect.

When user trigger a GridSelect action using mouse, this change makes it easier to cancel the Grid, without reaching to the keyboard.

### Checklist

        manually, ...) and concluded:

    I have tested the code manually. The default setting preserves the original behavior. User has to opt-in to get the new feature, to reduce surprises.

    To test the new behavior,
```haskell
import XMonad.Actions.GridSelect

myGoToSelected = goToSelected def { gs_cancelOnEmptyClick = True }

-- try it with a key binding
, ("M-<Space>", myGoToSelected)
```

### DIFF
--- a/XMonad/Actions/GridSelect.hs
+++ b/XMonad/Actions/GridSelect.hs
@@ -203,10 +203,13 @@ data GSConfig a = GSConfig {
       gs_colorizer :: a -> Bool -> X (String, String),
       gs_font :: String,
       gs_navigate :: TwoD a (Maybe a),
+      -- ^ Customize key bindings for a GridSelect
       gs_rearranger :: Rearranger a,
       gs_originFractX :: Double,
       gs_originFractY :: Double,
-      gs_bordercolor :: String
+      gs_bordercolor :: String,
+      gs_cancelOnEmptyClick :: Bool
+      -- ^ When True, click on empty space will cancel GridSelect
 }
 
 -- | That is 'fromClassName' if you are selecting a 'Window', or
@@ -386,13 +389,20 @@ updateElementsWithColorizer colorizer elementmap = do
 stdHandle :: Event -> TwoD a (Maybe a) -> TwoD a (Maybe a)
 stdHandle ButtonEvent{ ev_event_type = t, ev_x = x, ev_y = y } contEventloop
     | t == buttonRelease = do
-        s@TwoDState { td_paneX = px, td_paneY = py,
-                         td_gsconfig = (GSConfig ch cw _ _ _ _ _ _ _ _) } <- get
+        s@TwoDState{ td_paneX = px
+                   , td_paneY = py
+                   , td_gsconfig = GSConfig{ gs_cellheight = ch
+                                           , gs_cellwidth = cw
+                                           , gs_cancelOnEmptyClick = cancelOnEmptyClick
+                                           }
+                   } <- get
         let gridX = (fi x - (px - cw) `div` 2) `div` cw
             gridY = (fi y - (py - ch) `div` 2) `div` ch
         case lookup (gridX,gridY) (td_elementmap s) of
              Just (_,el) -> return (Just el)
-             Nothing -> contEventloop
+             Nothing     -> if cancelOnEmptyClick
+                            then return Nothing
+                            else contEventloop
     | otherwise = contEventloop
 
 stdHandle ExposeEvent{} contEventloop = updateAllElements >> contEventloop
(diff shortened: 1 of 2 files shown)


XMonad/Actions/GridSelect.hs
```diff
@@ -387,12 +392,16 @@ stdHandle :: Event -> TwoD a (Maybe a) -> TwoD a (Maybe a)
...
         let gridX = (fi x - (px - cw) `div` 2) `div` cw
             gridY = (fi y - (py - ch) `div` 2) `div` ch
         case lookup (gridX,gridY) (td_elementmap s) of
              Just (_,el) -> return (Just el)
-             Nothing -> contEventloop
+             Nothing -> if cancelOnEmptyClick
+               then
+                 return Nothing
+               else
+                 contEventloop
```
slotThe, line 404: I guess this could be `whenMaybe`

XMonad/Actions/GridSelect.hs
```diff
@@ -387,12 +392,16 @@ stdHandle :: Event -> TwoD a (Maybe a) -> TwoD a (Maybe a)
 stdHandle ButtonEvent{ ev_event_type = t, ev_x = x, ev_y = y } contEventloop
     | t == buttonRelease = do
         s@TwoDState { td_paneX = px, td_paneY = py,
-                         td_gsconfig = (GSConfig ch cw _ _ _ _ _ _ _ _) } <- get
+                         td_gsconfig = (GSConfig ch cw _ _ _ _ _ _ _ _ cancelOnEmptyClick) } <- get
```
slotThe, line 395: Oh boy :) I would use field punning here, to make this more readable (but this is personal style, I suppose)

XMonad/Actions/GridSelect.hs
```diff
@@ -203,10 +204,14 @@ data GSConfig a = GSConfig {
       gs_colorizer :: a -> Bool -> X (String, String),
       gs_font :: String,
       gs_navigate :: TwoD a (Maybe a),
+                                -- ^ Customize key bindings for a GridSelect
       gs_rearranger :: Rearranger a,
       gs_originFractX :: Double,
       gs_originFractY :: Double,
-      gs_bordercolor :: String
+      gs_bordercolor :: String,
+      gs_cancelOnEmptyClick :: Bool
```
slotThe, line 212: Since `GSConfig` exports its fields, this should be mentioned as a breaking change

slotThe (commented):
Looks fine to me, thanks :)

sylecn:
Hi slotThe, thanks for the review.

I fixed some issues you mentioned and force pushed.
Remaining two issues:

1. about whenMaybe. whenMaybe is in extra package, which is not currently in dependency list. Do you want to add it as a dependency, or just add a local definition?
2. how to use whenMaybe here? A quick look on the doc gives the following code which doesn't type check.
```haskell
Nothing -> whenMaybe (not $ gs_cancelOnEmptyClick gsconfig) contEventloop
```
```
xmonad-contrib>     • Couldn't match type ‘a’ with ‘Maybe a’
xmonad-contrib>       Expected: TwoD a a
xmonad-contrib>         Actual: TwoD a (Maybe a)
xmonad-contrib>       ‘a’ is a rigid type variable bound by
xmonad-contrib>         the type signature for:
xmonad-contrib>           stdHandle :: forall a.
xmonad-contrib>                        Event -> TwoD a (Maybe a) -> TwoD a (Maybe a)
```
3. since the change is moved to "Breaking Changes" section, would you say it's okay to set gs_cancelOnEmptyClick to True by default? Which value is a better default?

slotThe:
>     1. about whenMaybe. whenMaybe is in extra package, which is not currently in dependency list. Do you want to add it as a dependency, or just add a local definition?

Oops, I thought we have this in ManageHook. Nevermind then, you don't have to create an extra definition for this.

>     3. since the change is moved to "Breaking Changes" section, would you say it's okay to set gs_cancelOnEmptyClick to True by default? Which value is a better default?

You don't have to necessarily move the whole thing to breaking changes, just the part that's actually a breaking change. I'd say this is can be a reasonable default

sylecn:
I set the default value to True and updated the doc accordingly.

XMonad/Actions/GridSelect.hs
```diff
@@ -648,7 +657,7 @@ gridselect gsconfig elements =
     liftIO $ mapWindow dpy win
     liftIO $ selectInput dpy win (exposureMask .|. keyPressMask .|. buttonReleaseMask)
     status <- io $ grabKeyboard dpy win True grabModeAsync grabModeAsync currentTime
-    io $ grabPointer dpy win True buttonReleaseMask grabModeAsync grabModeAsync none none currentTime
+    _ <- io $ grabPointer dpy win True buttonReleaseMask grabModeAsync grabModeAsync none none currentTime
```
geekosaur, line 660: `XMonad.Prelude` exports `void` for this, which is more idiomatic.
```suggestion
    void $ io $ grabPointer dpy win True buttonReleaseMask grabModeAsync grabModeAsync none none currentTime
```

geekosaur (approved):

slotThe:
Thanks!

PR #895, opened 2024-06-16
X.A.WindowBringer: Add copyMenu

thalting:
### XMonad.Prompt.WindowBringer (copyMenu)

A convenient way to copy a window to the current workspace.

### Checklist

### DIFF
--- a/XMonad/Actions/WindowBringer.hs
+++ b/XMonad/Actions/WindowBringer.hs
@@ -22,6 +22,7 @@ module XMonad.Actions.WindowBringer (
                     WindowBringerConfig(..),
                     gotoMenu, gotoMenuConfig, gotoMenu', gotoMenuArgs, gotoMenuArgs',
                     bringMenu, bringMenuConfig, bringMenu', bringMenuArgs, bringMenuArgs',
+                    copyMenu, copyMenuConfig, copyMenu', copyMenuArgs, copyMenuArgs',
                     windowMap, windowAppMap, windowMap', bringWindow, actionMenu
                    ) where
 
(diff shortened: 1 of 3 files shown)


XMonad/Actions/WindowBringer.hs
```diff
@@ -90,6 +93,37 @@ gotoMenu' cmd = gotoMenuConfig def { menuArgs = [], menuCommand = cmd }
 gotoMenuArgs' :: String -> [String] -> X ()
 gotoMenuArgs' cmd args = gotoMenuConfig def { menuCommand = cmd, menuArgs = args }
 
+-- | Pops open a dmenu with window titles. Choose one, and it will be copied into your current workspace.
+copyMenu :: X ()
```
geekosaur, line 97: I can't help but wonder if this indicates we need some refactoring so we don't have to define five versions of everything. (No, I'm not suggesting you do it!)

geekosaur (approved):

geekosaur:
Looks like you forgot to include your changes to `xmonad-contrib.cabal`?

thalting:
> Looks like you forgot to include your changes to `xmonad-contrib.cabal`?

What changes? Wouldn't that be for new modules?

geekosaur:
Oh, I see, you don't run the test suite. The problem is that the test suite tests `XMonad.Actions.WindowBringer`, so its new dependencies need to be added to its `other-modules`.

geekosaur:
Okay, that looks better. Although now that I look, I don't see the test suite actually using WindowBringer at all, so possibly it needs to be removed (not your problem, it's future work and I'll have to check for any other dependencies the test suite isn't using).

geekosaur:
Thanks!

geekosaur:
Looks like something is using it after all…

PR #896, opened 2024-06-29
X.A.WindowNavigation: better handling of floating windows and Full layout

Rogach:
### Description

Previous version of WindowNavigation just got "stuck" on floating windows, switching back and forth between floating window and tiled window underneath. This was because "magic point" was left in the same position and thus next navigation commands selected the same windows over and over again. Now the "magic point" is moved around such that it doesn't overlap with the previously selected window, and there are more complicated rules to make navigation between floating and tiled windows more natural.

The original behavior of navigating between tiled windows is preserved almost precisely (the tree-like layouts - which should be overwhelmingly the most popular ones - will have exactly the same behavior, the only differences might be in some convoluted custom layouts).

Previous version of WindowNavigation also prevented switching windows using focusUp/focusDown when in Full layout. Now there's a special case that handles such situations.

### Checklist

  - [x] I've tested the changes with unit tests. Just look how pretty the tests are - there are diagrams that explain what each case is actually testing!

### DIFF
--- a/XMonad/Actions/WindowNavigation.hs
+++ b/XMonad/Actions/WindowNavigation.hs
@@ -1,10 +1,12 @@
+{-# LANGUAGE TupleSections #-} -- I didn't want this, it's hlint's "suggestion" and it's apparently non-negotiable
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Actions.WindowNavigation
 -- Description :  Experimental rewrite of "XMonad.Layout.WindowNavigation".
 -- Copyright   :  (c) 2007  David Roundy <droundy@darcs.net>,
 --                          Devin Mullins <me@twifkak.com>
--- Maintainer  :  Devin Mullins <me@twifkak.com>
+-- Maintainer  :  Devin Mullins <me@twifkak.com>,
+--                Platon Pronko <platon7pronko@gmail.com>
 -- License     :  BSD3-style (see LICENSE)
 -- Stability   :  unstable
 -- Portability :  unportable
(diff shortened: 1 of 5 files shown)


XMonad/Actions/WindowNavigation.hs
```diff
@@ -135,87 +172,249 @@ swap = withTargetWindow swapWithFocused
...
+-- | Inserts new position into the state.
+modifyState :: WindowSet -> Point -> WNState -> WNState
+modifyState oldWindowSet =
+  M.insert (W.currentTag oldWindowSet)
+
+-- | "Jumps" the current position into the middle of target rectangle.
+-- (keeps the position as-is if it is already inside the target rectangle)
+centerPosition :: Rectangle -> Point -> Point
+centerPosition (Rectangle rx ry rw rh) pos@(Point x y) = do
+  if x >= rx && x < rx + fromIntegral rw && y >= ry && y < ry + fromIntegral rh
```
geekosaur, line 227: We have https://hackage.haskell.org/package/xmonad-0.18.0/docs/XMonad-Operations.html#v:pointWithin, would it be useful here?

XMonad/Actions/WindowNavigation.hs
```diff
@@ -135,87 +172,249 @@ swap = withTargetWindow swapWithFocused
...
+navigableTargets input@(_, oldWindowSet, _, _) dir currentRect currentPos = do
+  allScreensWindowsAndRectangles <- mapSnd (rectTransform dir) <$> windowRects input
+  let
+    screenWindows = S.fromList $ W.integrate' $ W.stack $ W.workspace $ W.current oldWindowSet
+    (thisScreenWindowsAndRectangles, otherScreensWindowsAndRectangles) = partition (\(w, _) -> S.member w screenWindows) allScreensWindowsAndRectangles
+
+    pos = pointTransform dir currentPos
+    wr = rectTransform dir currentRect
+
+    rectInside r = (rect_p1 r >= rect_p1 wr && rect_p1 r < rect_p2 wr && rect_p2 r > rect_p1 wr && rect_p2 r <= rect_p2 wr) &&
```
geekosaur, line 246: We also have https://hackage.haskell.org/package/xmonad-0.18.0/docs/XMonad-Operations.html#v:containedIn

geekosaur (approved):

XMonad/Actions/WindowNavigation.hs
Rogach: Yes, that's useful, I'll update the code.

XMonad/Actions/WindowNavigation.hs
Rogach: This one is less useful, because it doesn't handle the special case of a window that fully overlaps the current window in the orthogonal direction.

geekosaur:
I'm going to wait a few days for one of the other maintainers to look at it, but as they're often unavailable I won't wait too long.

slotThe:
> I'm going to wait a few days for one of the other maintainers to look at it, but as they're often unavailable I won't wait too long.

FWIW I definitely plan to look at this in the coming days

slotThe:
Thanks!

PR #900, opened 2024-08-22
X.A.UpKeys: Init

slotThe:
### Description

I've known about [this stackoverflow](https://stackoverflow.com/a/11308086) post for a while, but seemingly the code has never made it to contrib. I think that it could come in handy once in a while, so why not add it?

### Checklist

        manually, ...) and concluded: I tested the extended example and things seem to work nicely; no actual tests (yet‽)

    TODO

### DIFF
--- a/XMonad/Actions/UpKeys.hs
+++ b/XMonad/Actions/UpKeys.hs
@@ -1,0 +1,169 @@
+{-# LANGUAGE BlockArguments #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE TypeApplications #-}
+{- |
+   Module      : XMonad.Actions.UpKeys
+   Description : Bind an action to the release of a key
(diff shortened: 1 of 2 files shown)


XMonad/Actions/UpKeys.hs
```diff
@@ -0,0 +1,169 @@
...
+
+{- $usage
+You can use this module with the following in your @xmonad.hs@:
+
+> import XMonad.Actions.UpKeys
+
+Next, define the keys and actions you want to have happen on the release
+of a key:
+
+> myUpKeys = ezUpKeys $
```
slotThe, line 40: There are, I guess, a million ways one could go about implementing this. Do we ask users to specify whether they want the key grabbed for every key (possibly with a global default fallback)? Do we also accept bindings for key presses here, so that one doesn't need to "split" the functionality across different parts of the config? I haven't thought too much about this, and just went with the most obvious one. Open for opinions

geekosaur (approved):

Issue #899, opened 2024-08-06
Add more portable way of getting hostname in Layout.OnHost


cacharle:
It seems like it currently only checks for the `HOST` environment variable. It would be nice if it could check `/etc/hostname` or other environment variables.

Maybe use `Network.HostName`? https://hackage.haskell.org/package/hostname-1.0/docs/src/Network-HostName.html#getHostName

https://github.com/xmonad/xmonad-contrib/blob/eb7268451c017eeb56258b97effacc3050c8461d/XMonad/Layout/OnHost.hs#L119

geekosaur:
The commit log may be of interest (it looks like I added this before we moved to GitHub, so there's no PR and the commit description is a little wonky):
```
commit c691988bbfe4c267a2e562689cc9640e53a1ec88
Author: Brandon S Allbery KF8NH <allbery.b@gmail.com>
Date:   Tue Mar 20 03:09:12 2012 +0000

    XMonad.Layout.OnHost allows host-specific modifications to a layout, which
    
    is otherwise very difficult to do.  Similarly to X.L.PerWorkspace, it provides
    onHost, onHosts, modHost, and modHosts layout modifiers.  It attempts to do
    smart hostname comparison, such that short names will be matched with short
    names and FQDNs with FQDNs.
    
    This module currently requires that $HOST be set in the environment.
    You can use System.Posix.Env.setEnv to do so in xmonad.hs if need be.
    (Properly, this should be done via the network library, but I'm trying to
    avoid adding that dependency.)  An alternative would be to shell out to
    get the name, but that has considerable portability hurdles.
```
The `hostname` package doesn't look terrible, but any additional dependency requires some discussion. (Probably not much in this case.)

slotThe:
Would shelling out really be that much of a pain in terms of portability? I'd guess (well, hope :)) that all systems we still support have a `hostname` in path. 

In either case, the relevant code in that package seems to small (most of it is MS Windows stuff) that vendoring sounds like a better option to me, should we choose to go down that route: 

``` haskell
type HostName = String

foreign import ccall unsafe "gethostname" gethostname :: CString -> CSize -> IO CInt

getHostName :: IO HostName
getHostName = allocaArray0 size $ \cstr -> do
        throwErrnoIfMinus1_ "getHostName" $ gethostname cstr (fromIntegral size)
        peekCString cstr
    where size = 256
```

geekosaur:
[...]

PR #901, opened 2024-08-27
X.L.OnHost: Query gethostname if $HOST lookup fails

slotThe:
### Description

I hope I did this right.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/899

### Checklist

        manually, ...) and concluded:

    **I have not tested this**, so I'd be very glad if other people could do that for me! :)
    @cacharle

    TBD/maybe not even necessary.

### DIFF
--- a/XMonad/Layout/OnHost.hs
+++ b/XMonad/Layout/OnHost.hs
@@ -1,5 +1,6 @@
-{-# LANGUAGE FlexibleInstances, MultiParamTypeClasses #-}
-
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Layout.OnHost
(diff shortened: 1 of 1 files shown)


XMonad/Layout/OnHost.hs
```diff
@@ -154,3 +159,17 @@ eqFQDN a b
...
   | otherwise                    =                    a ==                    b
+
+-----------------------------------------------------------------------
+-- cbits
+
+foreign import ccall "gethostname" gethostname :: CString -> CSize -> IO CInt
+
+getHostName :: IO (Maybe String)
+getHostName = allocaArray0 size $ \cstr -> do
+  throwErrnoIfMinus1_ "getHostName" $ gethostname cstr (fromIntegral size)
```
slotThe, line 170: I was briefly worried here, but since `runLayout` is in a `catchX` statement in the core this should never blow up XMonad

geekosaur: No, but we do occasionally have people who get confused because their layout suddenly became `Full`.

geekosaur (approved):

XMonad/Layout/OnHost.hs
slotThe: I guess this shouldn't worsen the problem, since `$HOST` is still checked first, and only if that fails do we call this function

liskin: After reading the manpage a couple times, I convinced myself this will never fail on Linux. And seems quite unlikely to fail elsewhere too.

liskin (approved):

PR #904, opened 2024-09-15
fix: deprecated url literals

cafkafk:
### Description

Nix url litterals are being depricated in flake inputs and fails to build on
never versions. I've just quoted them so it keeps building.

Signed-off-by: Christina Sørensen <christina@cafkafk.com>

### Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,9 +2,9 @@
 # See NIX.md for an overview of module usage.
 {
   inputs = {
-    flake-utils.url = github:numtide/flake-utils;
-    git-ignore-nix.url = github:hercules-ci/gitignore.nix/master;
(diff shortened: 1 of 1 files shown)


cafkafk:
I'm making the assumption that no one cares to include this in `CHANGES.md`, but if people want nix flake fixes in there as well LMK and I'll add a note

liskin:
> I'm making the assumption that no one cares to include this in `CHANGES.md`, but if people want nix flake fixes in there as well LMK and I'll add a note

I don't think we distribute the nix stuff in the release tarball so yeah I think it's okay not to mention it in CHANGES

geekosaur:
We don't.

LSLeary (approved):

PR #915, opened 2024-12-18
docs: Don't link to re-exports from XMonad.Config.Prime

sol:


geekosaur (approved):
👍🏽

liskin:
I vaguely remember trying this a couple years ago but whatever version of Haddock we were using back then didn't seem to change its behaviour. Does this actually work now? That would be awesome! (But we might need to bump xmonad-docs to a newer version of ghc/haddock)

sol:
> Does this actually work now? That would be awesome! (But we might need to bump xmonad-docs to a newer version of ghc/haddock)

I didn't actually try. I just did the edit quickly on GitHub from mobile device.

slotThe:
From a cursory local test, seems like this doesn't work :(

sol:
`{-# OPTIONS_HADDOCK hide #-}`

slotThe (approved):
> `{-# OPTIONS_HADDOCK hide #-}` does the trick.

Seems to work :+1:

> This has the disadvantage that we lose the documentation of `XMonad.Config.Prime`, but given that it is deprecated I hope this is an easy compromise.

Indeed, a small price to pay

sol:
OK, great, from my side this is ready to merge.

sol:
Is there anything else to discuss here? If not, can somebody click the merge button?

PR #918, opened 2025-01-02
Auto-format `OnScreen` and `ScreenCorners` to HLS

nilscc:
### Description

Auto-format both my modules `OnScreen` and `ScreenCorners` with HLS. No other (functional) changes were done.

### Checklist

### DIFF
--- a/XMonad/Actions/OnScreen.hs
+++ b/XMonad/Actions/OnScreen.hs
@@ -1,4 +1,3 @@
------------------------------------------------------------------------------
 -- |
 -- Module      :  XMonad.Actions.OnScreen
 -- Description :  Control workspaces on different screens (in xinerama mode).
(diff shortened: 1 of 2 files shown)


slotThe (commented):
Are you planning on working on these modules in the future? If not, I don't personally think that putting another indirection into the git history is worth it, especially since the other contrib modules don't follow a certain style, but are sort of all over the place

nilscc:
@slotThe Yes, there might be an issue with screen corners on multi monitor setups. Working on it will be easier if HLS does not change the entire file once I save it...

geekosaur:
There are definitely issues with screen corners on multi-monitor setups:

1. The bounding box for screen corners is defined by the largest monitors, meaning that the lower left/right corners are often unreachable.
2. If the screen layout changes (hotplugging monitors, for example) you need to restart xmonad with mod-q to fix the screen corners.

Note that the first one actually requires some significant design work to fix. I made a start on it, but had to put it on the back burner.

nilscc:
@geekosaur Yes, exactly. Do you have an issue for it or should we create one? Can you also link your previous work?

geekosaur:
I chose not to file an issue yet after seeing how much work the redesign would be; and unfortunately I'm not finding the initial (very preliminary) design work, although I remember the fundamentals. The gist is that, especially with window sides support, you need to track sides per monitor and place the corners according to those.

```
+------------------------+------------------------+----------------------+
|                        |                        |                      |
|                        |                        |                      |
|                        |                        |                      |
|                        |                        |                      |
|                        |                        |                      |
|                        |                        |                      |
|                        |                        |                      |
|                        |                        |                      |
+------------------------+                        |                      |
                         |                        |                      |
                         |                        |                      |
                         |                        +----------------------+
                         |                        |
                         |                        |
*                        +------------------------+                      *
```

Both the lower left and right borders and corners are unreachable currently, because the root window's bounding box is dominated by the center monitor. (See the asterisks at the bottom; those are the current bottom corners.)

Some additional questions need to be answered, which is another reason I haven't put anything up:

1. Should the lower left/lower right corners be defined by the width or the height? That is, should the corners be those of the left and right monitors (defined by width), or those of the higher center monitor (defined by height)?
2. Should the edges similarly be defined, or (not applicable to corners) should both be active? Should there be some way to specify different behavior for the center monitor, if so?
3. This isn't the only possible layout; here I set all the tops at `y == 0`, but with xrandr or arandr you can align bottoms instead or even center them. This leads to even more complex issues (and complicates the above two questions even more).

geekosaur:
BTW, there is a pretty strong argument against trying to use the edges of the center window: X11 doesn't quite handle the difference between size and resolution well. The above is not quite how my current screen layout works (the left and right monitors are actually the same size and resolution in my setup), but the mouse often gets "stuck" moving into the center monitor (my laptop's display) from either, possibly because it's physically smaller while having a larger resolution so something thinks it's moving into an unreachable region.

nilscc:
Created an issue: https://github.com/xmonad/xmonad-contrib/issues/919

Please merge the PR so we can have clean diffs in the future.

PR #921, opened 2025-01-12
add XF86WLAN special key

### LINKED ISSUE #27: [XMonad.Util.EZConfig] Function key XF86WLAN missing
The `XF86WLAN` function key missing in the supported function keys of EZConfig:
http://xmonad.org/xmonad-docs/xmonad-contrib/XMonad-Util-EZConfig.html

On my ThinkPad T450s I've got this one to enable/disable wifi.
https://wiki.archlinux.org/index.php/Lenovo_ThinkPad_T450s#Function_Keys
Edit: There are actually one more unsupported function key:

```
$ xev | awk -F'[ )]+' '/^KeyPress/ { a[NR+2] } NR in a { printf "%-3s %s\n", $5, $8 }'
246 XF86WLAN
198 XF86AudioMicMute
```

geekosaur:
### Description

Add XF86WLAN to EZConfig.
Closes #27.

### Checklist

  - [x] I updated the `CHANGES.md` file (next push)

### DIFF
--- a/.github/workflows/stack.yml
+++ b/.github/workflows/stack.yml
@@ -12,10 +12,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - resolver: lts-14  # GHC 8.6
-            yaml: stack.yaml
-          - resolver: lts-14  # GHC 8.6
-            yaml: stack-master.yaml
           - resolver: lts-16  # GHC 8.8
             yaml: stack-master.yaml
           - resolver: lts-18  # GHC 8.10
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -8,9 +8,9 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.19.20241114
+# version: 0.19.20250115
 #
-# REGENDATA ("0.19.20241114",["github","cabal.project"])
+# REGENDATA ("0.19.20250115",["github","cabal.project"])
 #
 name: Haskell-CI
 on:
@@ -74,11 +74,6 @@ jobs:
           - compiler: ghc-8.8.4
             compilerKind: ghc
             compilerVersion: 8.8.4
-            setup-method: ghcup
-            allow-failure: false
-          - compiler: ghc-8.6.5
-            compilerKind: ghc
-            compilerVersion: 8.6.5
             setup-method: ghcup
             allow-failure: false
       fail-fast: false
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@
 
 ### Bug Fixes and Minor Changes
 
+  * `XMonad.Util.EZConfig`
+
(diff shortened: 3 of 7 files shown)


geekosaur:
@slotThe, OrgMode test failure only on 8.6.5, looks like a QuickCheck version problem possibly?

slotThe:
> @slotThe, OrgMode test failure only on 8.6.5, looks like a QuickCheck version problem possibly?

Yes, it seems to pick 2.7.x, while `arbitraryPrintableChar` is only available from `2.10.x` onwards. I'm curious, though, why this has never happened before?

geekosaur:
Betting some dependency changed its bounds, possibly by removing its own 8.6 support.

slotThe:
Debian stable already seems to be on 9.0… Should we just drop support for 8.6? (not because `arbitraryPrintableChar` is so difficult to define ourselves, but just as a matter of principle)

@liskin any opinions?

liskin:
> Debian stable already seems to be on 9.0… Should we just drop support for 8.6? (not because `arbitraryPrintableChar` is so difficult to define ourselves, but just as a matter of principle)
> 
> @liskin any opinions?

Yeah that's reasonable. Ubuntu 22.04, which seems to still be used quite a lot, has ghc 8.8.4, so we may want to keep that one.

Or not, I dunno. We're currently supporting (and testing against in CI!) 9 different major GHC versions, which seems quite excessive. It'd be 10 once the PRs adding 9.12 are merged (only looked at those briefly, some work needs to be done, and I'm kind of in the middle of doing that work for X11 and X11-xft). Just removing 8.6 gets us down to 9. Probably all right, we're far from running out of GitHub Actions budget :-)

slotThe (approved):
I think I did this right. Should be good to go now

geekosaur:
LGTM

.github/workflows/stack.yml
```diff
@@ -12,10 +12,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - resolver: lts-14  # GHC 8.6
-            yaml: stack.yaml
```
liskin, line 16: This was here to make sure we test compatibility with the latest hackage release of xmonad and X11 on both oldest and newest supported GHC. So only one of these should be dropped, and the stack.yaml one should be moved to a newer resolver.

slotThe: Ah, so that was what that was :)

geekosaur: I'm on it.

liskin: I should've added a comment really. It's a small miracle I still remember what the point was. Perhaps I don't and what I said isn't even true 😂

geekosaur: Looks correct to me from inspecting both the `stack.yaml` files. (I'm reminded of the stuff I added to `cabal`'s CI to do much the same thing, but it's a little more involved since of course they don't build with `stack`.)

Issue #776, opened 2022-11-03
Customize _NET_CURRENT_DESKTOP to w.greedyView desktop


jceb:
### Problem Description

I'd like xmonad to change the current workspace via `W.greedyView` when clicking on a workspace in the desktop bar. Currently, `ewmh` uses `W.view` in `ewmhDesktopsEventHook'`  which doesn't bring the selected workspace into sight when the workspace is displayed on a another monitor. Before version 0.17 I was able to patch `ewmhDesktopsEventHook'` manually and feed it back into the configuration via the hook system. With version 0.17 this workaround doesn't seem to be possible anymore.

This function is the one that causes the issue for me: https://hackage.haskell.org/package/xmonad-contrib-0.17.1/docs/src/XMonad.Hooks.EwmhDesktops.html#ewmhDesktopsEventHook%27

### Steps to Reproduce

1. Install a desktop bar like tint2
2. Start xmonad in a multi monitor xinerama setup
3. Show workspace 1 on monitor 1
4. Show workspace 2 on monitor 2
5. On monitor 2 click on workspace 1 in the desktop bar to switch to workspace 1
6. Observe that the workspaces didn't change monitors. The only difference is that now workspace 1 on monitor 1 is active

### Configuration File

Please include the smallest _full_ configuration file that reproduces
the problem you are experiencing:

```haskell
module Main (main) where

import XMonad
import XMonad.Hooks.EwmhDesktops
import XMonad.Hooks.ManageDocks

main :: IO ()
main = xmonad $ docks . ewmh $ def
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17.1
    - [x] With `xmonad-contrib` version 0.17.1

liskin:
We can add a configurable hook to https://hackage.haskell.org/package/xmonad-contrib-0.17.1/docs/src/XMonad.Hooks.EwmhDesktops.html#EwmhDesktopsConfig

(pull request welcome but one of us would get to it eventually too :-))

jceb:
Okay, I managed to get it to work .. it felt a bit more involved than before. Since it's not easily possible to feed my own custom `handleEventHook` in - `ewmh` places its own event hook in front of any custom hook that I pass in, I resolved to patching the config that `ewmh` returns.
[...]

PR #923, opened 2025-01-24
X.H.EwmhDesktops: Add customization for handling the _NET_CURRENT_DESKTOP requests

m1mir:
### Description

This adds a `setEwmhSwitchDesktopAction` function to customize how switching to the desired workspace is done when handling a _NET_CURRENT_DESKTOP request. The motivation for this was that commands using the EWMH fuctionality such as `wmctrl -s` pulled workspaces to the wrong screen in certain cases when using the `X.L.IndependentScreens` module, instead of making them active on the screen that they belong on, but the behavior seemed to be a sensible default since use of that module is not mandatory.

With the new function I changed the action to the following:
`action workspaceId = viewOnScreen (unmarshallS workspaceId) workspaceId`
This no longer pulls the workspace to the wrong screen when it is hidden, and it is on a screen that does not contain the current workspace, this is also used as the example in the documentation.

### Checklist

  - [x] I've considered how to best test these changes (property, unit,manually, ...) and concluded:
         that manual testing is adequate since change just added a customization option without changing the default.
         I did not find a difference in the default behavior, using this version and the example is currently working as expected
         on my system.
 
### DIFF
--- a/XMonad/Hooks/EwmhDesktops.hs
+++ b/XMonad/Hooks/EwmhDesktops.hs
@@ -42,6 +42,10 @@ module XMonad.Hooks.EwmhDesktops (
     -- $customActivate
     setEwmhActivateHook,
 
+    -- ** Workspace switching
+    -- $customWorkspaceSwitch
+    setEwmhSwitchDesktopHook,
+
     -- ** Fullscreen
     -- $customFullscreen
     setEwmhFullscreenHooks,
@@ -114,6 +118,8 @@ data EwmhDesktopsConfig =
             -- ^ configurable handling of window activation requests
         , fullscreenHooks :: (ManageHook, ManageHook)
             -- ^ configurable handling of fullscreen state requests
+        , switchDesktopHook :: WorkspaceId -> WindowSet -> WindowSet
+            -- ^ configurable action for handling _NET_CURRENT_DESKTOP
         , manageDesktopViewport :: Bool
             -- ^ manage @_NET_DESKTOP_VIEWPORT@?
         }
@@ -124,6 +130,7 @@ instance Default EwmhDesktopsConfig where
         , workspaceRename = pure pure
         , activateHook = doFocus
         , fullscreenHooks = (doFullFloat, doSink)
+        , switchDesktopHook = W.view
         , manageDesktopViewport = True
         }
 
@@ -243,6 +250,31 @@ setEwmhWorkspaceRename f = XC.modifyDef $ \c -> c{ workspaceRename = f }
 -- "XMonad.ManageHook", "XMonad.Hooks.ManageHelpers" and "XMonad.Hooks.Focus".
 setEwmhActivateHook :: ManageHook -> XConfig l -> XConfig l
 setEwmhActivateHook h = XC.modifyDef $ \c -> c{ activateHook = h }
+
+
+-- $customWorkspaceSwitch
+-- When a client sends a @_NET_CURRENT_DESKTOP@ request to switch to a workspace,
+-- the default action used to do that is the 'W.view' function.
+-- This may not be the desired behaviour in all configurations.
+--
+-- For example if using the "XMonad.Layout.IndependentScreens" the default action
+-- might move a workspace to a screen that it isn't supposed to be on.
+-- This behaviour can be fixed using the following:
+--
+-- > import XMonad.Actions.OnScreen
+-- > import XMonad.Layout.IndependentScreens
+-- >
+-- > main = xmonad $ ... . setEwmhSwitchDesktopHook focusWorkspace . ewmh . ... $
+-- >  def{
+-- >    ...
+-- >      workspaces = withScreens 2 (workspaces def)
+-- >    ...
+-- >  }
+
+-- | Set (replace) the action which is invoked when a client sends a
+-- @_NET_CURRENT_DESKTOP@ request to switch workspace.
+setEwmhSwitchDesktopHook :: (WorkspaceId -> WindowSet -> WindowSet) -> XConfig l -> XConfig l
+setEwmhSwitchDesktopHook action = XC.modifyDef $ \c -> c{ switchDesktopHook = action }
 
 
 -- $customFullscreen
@@ -449,7 +481,7 @@ mkViewPorts winset = setDesktopViewport . concat . mapMaybe (viewPorts M.!?)
 ewmhDesktopsEventHook' :: Event -> EwmhDesktopsConfig -> X All
 ewmhDesktopsEventHook'
         ClientMessageEvent{ev_window = w, ev_message_type = mt, ev_data = d}
-        EwmhDesktopsConfig{workspaceSort, activateHook} =
+        EwmhDesktopsConfig{workspaceSort, activateHook, switchDesktopHook} =
     withWindowSet $ \s -> do
         sort' <- workspaceSort
         let ws = sort' $ W.workspaces s
@@ -462,7 +494,7 @@ ewmhDesktopsEventHook'
         if  | mt == a_cw ->
                 killWindow w
             | mt == a_cd, n : _ <- d, Just ww <- ws !? fi n ->
-                if W.currentTag s == W.tag ww then mempty else windows $ W.view (W.tag ww)
+                if W.currentTag s == W.tag ww then mempty else windows $ switchDesktopHook (W.tag ww)
             | mt == a_cd ->
                 trace $ "Bad _NET_CURRENT_DESKTOP with data=" ++ show d
             | not (w `W.member` s) ->
--- a/XMonad/Layout/IndependentScreens.hs
+++ b/XMonad/Layout/IndependentScreens.hs
@@ -27,7 +27,7 @@ module XMonad.Layout.IndependentScreens (
     whenCurrentOn,
     countScreens,
     workspacesOn,
-    workspaceOnScreen, focusWindow', focusScreen, nthWorkspace, withWspOnScreen,
+    workspaceOnScreen, focusWindow', focusScreen, focusWorkspace, nthWorkspace, withWspOnScreen,
     -- * Converting between virtual and physical workspaces
     -- $converting
     marshall, unmarshall, unmarshallS, unmarshallW,
@@ -40,6 +40,7 @@ import XMonad
 import XMonad.Hooks.StatusBar.PP
 import XMonad.Prelude
 import qualified XMonad.StackSet as W
+import XMonad.Actions.OnScreen (viewOnScreen)
 
 -- $usage
 -- You can use this module with the following in your @xmonad.hs@:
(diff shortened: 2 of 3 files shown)


XMonad/Hooks/EwmhDesktops.hs
```diff
@@ -245,6 +252,33 @@ setEwmhActivateHook :: ManageHook -> XConfig l -> XConfig l
...
+-- > import XMonad.Layout.IndependentScreens
+-- >
+-- > main = xmonad $ ... . setEwmhSwitchDesktopAction switchDesktopAction . ewmh . ... $
+-- >  def{
+-- >    ...
+-- >      workspaces = withScreens 2 (workspaces def)
+-- >    ...
+-- >  }
+-- >
+-- > switchDesktopAction workspaceId = viewOnScreen (unmarshallS workspaceId) workspaceId
```
geekosaur, line 274: You might consider defining and exporting this from X.L.IndependentScreens.

geekosaur (approved):
Otherwise, looks good.

XMonad/Hooks/EwmhDesktops.hs
m1mir: Added the function to X.L.IndependentScreens, and documented it.

m1mir:
I've tested the example usage of the new focusWorkspace manually, as far as I can tell it works as expected.

geekosaur:
I'm interested in seeing what @liskin  or @slotThe says about it.

geekosaur (approved):

geekosaur:
Sorry, I'd already re-reviewed but didn't push the "approve" button; wanted to wait on CI.

XMonad/Hooks/EwmhDesktops.hs
```diff
@@ -114,6 +118,8 @@ data EwmhDesktopsConfig =
             -- ^ configurable handling of window activation requests
         , fullscreenHooks :: (ManageHook, ManageHook)
             -- ^ configurable handling of fullscreen state requests
+        , switchDesktopAction :: WorkspaceId -> WindowSet -> WindowSet
```
liskin, line 121: I'd maybe consider this being named a `…Hook` as well. It's not a ManageHook, but it is conceptually a hook — just like `handleEventHook` and similar.

XMonad/Layout/IndependentScreens.hs
```diff
@@ -163,6 +164,12 @@ focusWindow' window ws
 focusScreen :: ScreenId -> WindowSet -> WindowSet
 focusScreen screenId = withWspOnScreen screenId W.view
 
+
```
liskin, line 167: extra line slipped here somehow? :-)

liskin (commented):
It is beatiful. :-)

Commented about a couple of nits anyway, as I usually do.

Plus worth noting that a complete integration with IndependentScreens would also involve using `focusWindow'` for `activateHook` (converted into a ManageHook), but it would still potentially misbehave because _NET_ACTIVE_WINDOW requests from pagers (and tools like wmctrl) aren't overridable right now. :-(

m1mir:
I tested it again manually after the rename. It works as expected.

XMonad/Layout/IndependentScreens.hs
m1mir: Yes, I accidentally copied one more for the documentation and didn't read the diff properly when committing.

XMonad/Hooks/EwmhDesktops.hs
m1mir: I've renamed based on your suggestion. It really makes more sense to call it that, I was just hesitant to do so because it wasn't its type.

m1mir:
> Plus worth noting that a complete integration with IndependentScreens would also involve using `focusWindow'` for `activateHook` (converted into a ManageHook), but it would still potentially misbehave because _NET_ACTIVE_WINDOW requests from pagers (and tools like wmctrl) aren't overridable right now. :-(

Could you clarify what you meant by this paragraph?
I've did a little experimenting and if I just change the `switchDesktopHook` to `focusWorkspace` and I just leave the activate hook as the default `doFocus`, then `wmctrl -a` works as expected.
I think that's because [this function in wmctrl](https://github.com/Conservatory/wmctrl/blob/7c1dcbaa221a16e8a9765bf2b14d61d640a52304/main.c#L703) as far as I can tell always does a `_NET_CURRENT_DESKTOP` request first when using XMonad and then it does the  `_NET_ACTIVE_WINDOW` request, and the first request was probably the source of the weird behavior.

liskin:
Oh okay I didn't realise wmctrl switches the desktop first - that's awesome. There may be something else that doesn't, but maybe we don't need to care - not having to complicate the code any more is good.

geekosaur:
We are permitted to convert it to an urgency indicator if it's not on the active workspace.

m1mir:
> Oh okay I didn't realise wmctrl switches the desktop first - that's awesome. There may be something else that doesn't, but maybe we don't need to care - not having to complicate the code any more is good.

The things that I intended to convey were the following:

- Just customizing the switch desktop action is enough for `wmctrl -a` to work "properly" with IndependentScreens.
- With the current API, by which I mean the package's released version's, the case when just a `_NET_ACTIVE_WINDOW` is sent is already easily handled by changing the activate hook.

Therefore I consider the contents of this PR complete.

m1mir:
Also I haven't noticed until now but this probably closes this #776 issue.

geekosaur (approved):

slotThe:
Thanks!

liskin:
> > Oh okay I didn't realise wmctrl switches the desktop first - that's awesome. There may be something else that doesn't, but maybe we don't need to care - not having to complicate the code any more is good.
> 
> The things that I intended to convey were the following:
> 
> * Just customizing the switch desktop action is enough for `wmctrl -a` to work "properly" with IndependentScreens.
> * With the current API, by which I mean the package's released version's, the case when just a `_NET_ACTIVE_WINDOW` is sent is already easily handled by changing the activate hook.
> 
> Therefore I consider the contents of this PR complete, unless modifications are requested.

Yeah, I agree that the PR was good and no modifications were requested from me. Sorry for not being clear earlier, must've been in a hurry, and then went AFK.

The only little issue that may still be unclear is that there's that special case for `_NET_ACTIVE_WINDOW` requests from pagers (an external tool for switching windows/desktops) that is unconditionally handled using `focusWindow` as requests from pagers aren't meant to be ignored. The code for customization of `_NET_ACTIVE_WINDOW` was motivated by the need to ignore some requests, but it seems ill-prepared for use-cases such as integration with IndependentScreens — so the unconditional special case isn't overridable. Fortunately, it's not needed with `wmctrl`, but in theory there might be a pager that just sends a `_NET_ACTIVE_WINDOW` request without doing `_NET_CURRENT_DESKTOP`. But I think it's okay not to complicate the code until someone actually needs it.

Hope this makes it clearer :-)

m1mir:
Sorry, I know that this has been merged. But I 

> > > Oh okay I didn't realise wmctrl switches the desktop first - that's awesome. There may be something else that doesn't, but maybe we don't need to care - not having to complicate the code any more is good.
> > 
> > 
> > The things that I intended to convey were the following:
> > 
> > * Just customizing the switch desktop action is enough for `wmctrl -a` to work "properly" with IndependentScreens.
> > * With the current API, by which I mean the package's released version's, the case when just a `_NET_ACTIVE_WINDOW` is sent is already easily handled by changing the activate hook.
> > 
> > Therefore I consider the contents of this PR complete, unless modifications are requested.
> 
> Yeah, I agree that the PR was good and no modifications were requested from me. Sorry for not being clear earlier, must've been in a hurry, and then went AFK.
> 
> The only little issue that may still be unclear is that there's that special case for `_NET_ACTIVE_WINDOW` requests from pagers (an external tool for switching windows/desktops) that is unconditionally handled using `focusWindow` as requests from pagers aren't meant to be ignored. The code for customization of `_NET_ACTIVE_WINDOW` was motivated by the need to ignore some requests, but it seems ill-prepared for use-cases such as integration with IndependentScreens — so the unconditional special case isn't overridable. Fortunately, it's not needed with `wmctrl`, but in theory there might be a pager that just sends a `_NET_ACTIVE_WINDOW` request without doing `_NET_CURRENT_DESKTOP` first. But I think it's okay not to complicate the code until someone actually needs it.
> 
> Hope this makes it clearer :-)

I know that this has been merged, but this seems to be the appropriate space to ask. What do you mean by ill suited?

In my config I use `doF . focusWindow' =<< ask` for the activate hook as I mentioned previously. It is almost the same as the original `doFocus` but it uses the IndependentScreens module's `focusWindow'` instead of the the StackSet one and it does work perfectly as far as I can tell. For example it focuses Firefox and Discord without pulling them and their workspace to the wrong physical screen.

If that is a hook that solves the issues that you are thinking about, I could add it to X.L.IndependentScreens in a PR tomorrow.

liskin:
> I know that this has been merged, but this seems to be the appropriate space to ask. What do you mean by ill suited?

I'm specifically referring to this bit: https://github.com/xmonad/xmonad-contrib/blob/4fc3642fa28e1cb405afdf12f922576a7e0705a9/XMonad/Hooks/EwmhDesktops.hs#L509-L512
Since I didn't consider the use-case of integrating IndependentScreens when implementing this, the `W.focusWindow` there isn't overridable.

> In my config I use `doF . focusWindow' =<< ask` for the activate hook as I mentioned previously. It is almost the same as the original `doFocus` but it uses the IndependentScreens module's `focusWindow'` instead of the the StackSet one and it does work perfectly as far as I can tell. For example it focuses Firefox and Discord without pulling them and their workspace to the wrong physical screen.
> 
> If that is a hook that solves the issues that you are thinking about, I could add it to X.L.IndependentScreens in a PR tomorrow.

It doesn't solve the issue I'm thinking about, but then we don't know if that issue isn't purely hypothetical. 🙂

It definitely **does solve** the actual issue you encountered, so either documenting it (it's simple enough really) or adding it as a function to IndependentScreens seems valuable.

---

tl;dr - only read this bit if you enjoy discussing complex hypothetical technical shit

To expand a bit more on my thinking back when the hooks were implemented — I was _aware_ of the theoretical possibility of someone needing to override all the ways `_NET_ACTIVE_WINDOW` is handled, but I couldn't come up with a code that would look good. It was either:

* a `Bool → ManageHook`, which is ugly (boolean blindness) and prevents just using an existing function of type `ManageHook` and thus requires Haskell knowledge to be useful
* a `SomeNewSumType → ManageHook`, which is nicer but also too much new lines of code
* two `ManageHook`s, which is almost okay and I probably should've done this but then it was sorta okay to wait until it's actually needed (and we **still don't know** if it actually is or if it's just hypothetical)

m1mir:
I understand now. Thank you for the explanation. I found the second part especially fascinating because I typically design poor APIs, and this gave me examples of what to consider.

PR #962, opened 2025-12-15
X.H.SB.PP: Added `ppCurrentNoWindows`

fpeterek:
### Description

We already have `ppVisibleNoWindows` and `ppHiddenNoWindows`, this change adds the missing `ppCurrentNoWindows` printer, as well as matching predicates (`isCurrentNoWindows`, `isCurrentHasWindows`, `hasWindows`). `isCurrent` remains as was and doesn't care about windows. However, the new predicates allow you to check whether a layout is the `current` one but also contains/doesn't contain windows. `ppCurrentNoWindows` falls back to `ppCurrent` if set to `Nothing` (also the default value), so existing configurations should not break in any way, but can be extended easily.

This PR shouldn't break anything as the existing public API remains unchanged, with only slight internal implementation changes and new additions to the API.

It's also not strictly necessary, you could achieve similar results by using `ppPrinters` and creating a custom predicate, but that just seems like an inconsistent and complicated way to do it, when we already have the other two printers. The change in code is rather small and could simplify configuration for some users.

In the first picture, the workspace has at least one window open.

current

And here's an example of the current workspace not having any windows open.

current-no-windows

Sample configuration, which should allow you to reproduce the example with Xmobar.

```haskell
xmobarPP' :: PP
xmobarPP' = def
    { ppCurrent          = pad . xmobarBorder "Bottom" "#6dcbfa" 3 . white
    , ppCurrentNoWindows = Just $ pad . xmobarBorder "Bottom" "#6dcbfa" 3 . gray
    , ppVisible          = pad . xmobarBorder "Bottom" "#215975" 3 . white
    , ppVisibleNoWindows = Just $ pad . xmobarBorder "Bottom" "#215975" 3 . gray
    , ppHidden           = white . pad
    , ppHiddenNoWindows  = gray . pad
    }
  where
      white        = xmobarColor "#e0e0e0" ""
      gray         = xmobarColor "#89878e" ""
```

### Checklist

  - [*] I've considered how to best test these changes (property, unit,
        manually, ...) and concluded: Test manually, swap to a workspace with no windows

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,15 +6,6 @@
 
   * Drop support for GHC 8.6
 
-  * `XMonad.Hooks.StatusBar.PP`
-
-    - Added `ppCurrentNoWindows` to `PP`, to match `ppHiddenNoWindows`
-      and `ppVisibleNoWindows`.
-    - Added `isCurrentNoWindows` and `hasWindows` predicates.
-    - `isCurrent` now checks if the workspace contains any windows,
-      akin to `isVisible`. This may break certain configurations that
-      rely on `isCurrent`.
-
 ### Bug Fixes and Minor Changes
 
   * `XMonad.Actions.WindowBringer`
@@ -99,6 +90,18 @@
   * `XMonad.Hooks.ManageDocks`
 
     - Added `onAllDocks` to run an action on all recognised docks.
+
+  * `XMonad.Hooks.StatusBar.PP`
+
+    - Added `ppCurrentNoWindows` to `PP`, to match `ppHiddenNoWindows`
+      and `ppVisibleNoWindows`. `ppCurrentNoWindows` falls back
+      to `ppCurrent` if unset/set to `Nothing`.
+    - Added `isCurrentNoWindows`, `isCurrentHasWindows`, and `hasWindows`
+      predicates.
+    - `isCurrent` remains as was and doesn't check for presence
+      of windows. The changes only added a simple way to check whether
+      the workspace is a current workspace that does/doesn't contain
+      windows.
 
 ## 0.18.1 (August 20, 2024)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,15 @@
 
   * Drop support for GHC 8.6
 
+  * `XMonad.Hooks.StatusBar.PP`
+
+    - Added `ppCurrentNoWindows` to `PP`, to match `ppHiddenNoWindows`
+      and `ppVisibleNoWindows`.
+    - Added `isCurrentNoWindows` and `hasWindows` predicates.
+    - `isCurrent` now checks if the workspace contains any windows,
+      akin to `isVisible`. This may break certain configurations that
+      rely on `isCurrent`.
+
 ### Bug Fixes and Minor Changes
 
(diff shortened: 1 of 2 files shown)


geekosaur:
I will wait to see what Solid has to say, but my initial reaction is that breaking an API going back to 2007 is probably not the greatest of ideas, especially when it'd be silent breakage.

fpeterek:
Fair point, we can also change the predicates to make the change non-breaking, at the cost of inconsistency (probably worth it). Adding just the printer should not break anything.

If you want to merge it in some form or another, lmk, I can change the PR. Or you can just close it, I don't care either way, I'm just recently switching to XMonad and found this surprisingly difficult to configure for what it is, but I'm fine just changing my own config.

slotThe:
> I will wait to see what Solid has to say, but my initial reaction is that breaking an API going back to 2007 is probably not the greatest of ideas, especially when it'd be silent breakage.

Yeah, I think I would agree. Perhaps a separate printer would be better

TheMC47:
> > I will wait to see what Solid has to say, but my initial reaction is that breaking an API going back to 2007 is probably not the greatest of ideas, especially when it'd be silent breakage.
> 
> Yeah, I think I would agree. Perhaps a separate printer would be better

or just a logger?

slotThe:
> or just a logger? 

Yes! This is what I was thinking of, since we were talking about
ppExtras anyways. In my head it was clear, I swear! :)

geekosaur (approved):

TheMC47 (commented):
I was checking the code again, and I'm not sure of extending the PP type for every predicate (and I'm not sure where we should draw the line).

Did you check the `ppPrinters` API?You could add your own predicate with your own format, and it will take precedence

fpeterek:
@TheMC47 I had to go through the API to implement this and figured as much, yes. I get your point and I know I can implement it in my config. As I mentioned, my point was not really to add "just any predicate." Personally, I just found it needlessly confusing that there are `ppVisibleNoWindows` and `ppHiddenNoWindows`, but there's no `ppCurrentNoWindows`.

I do not mean to argue or provoke, and I am aware XMonad is purposefully minimalistic. But not having something akin to `ppCurrentNoWindows` seemed more like an oversight, bug, or just an arbitrary choice, especially when it's a tiny change, and other window managers implement it out-of-the-box.

I submitted a PR because it was an easy thing to implement, and thus I thought it would be easier to discuss over an existing implementation, but I'm of course not adamant on having this merged, especially as a first time contributor. We can also merge just the predicates without the pretty printer change, or we can throw this PR away completely. Feel free to request changes, or close the PR without merging.

TheMC47 (approved):
Thanks for your contribution! I think it's fine to go forward with this as it's backwards compatible and maybe a more accessible API. Minimalism in contrib is not _that_ important.

I'll wait to see if someone has a different opinion, otherwise I'll merge this.

geekosaur:
I don't think minimalism has ever been a criterion for contrib, except with respect to dependencies (that's where extras comes in).

TheMC47:
Thank you!

TheMC47:
Crap, I messed up the change log. I'll push a fix

PR #974, opened 2026-03-10
Add support for `_NET_WM_USER_TIME`

basil-conto:
### Description

This PR implements:
- #973

Namely, it:
- advertises `_NET_WM_USER_TIME` and `_NET_WM_USER_TIME_WINDOW` support in `X.H.EwmhDesktops.setSupported`; and
- adds a predicate `X.H.ManageHelpers.hasNoFocusOnMap` which checks these properties.

Things I'm unsure of:

- The name and sense of the new predicate. Negatives can be confusing, but at the same time it checks for a non-default situation, thus does not require logical negation in a custom `ManageHook`, as seen in the example I included in its Haddock comment.

- ~Why I see this error when running `stack test` locally:~
  ```
  xmonad-contrib> Preprocessing test suite 'tests' for xmonad-contrib-0.18.2.9...
  xmonad-contrib> Building test suite 'tests' for xmonad-contrib-0.18.2.9...
  xmonad-contrib> [45 of 65] Compiling XMonad.Hooks.ManageHelpers
  xmonad-contrib> /home/blc/.config/xmonad/xmonad-contrib/XMonad/Hooks/ManageHelpers.hs:86:1: error: [GHC-87110]
  xmonad-contrib>     Could not load module ‘Control.Monad.Trans.Maybe’.
  xmonad-contrib>     It is a member of the hidden package ‘transformers-0.6.1.1’.
  xmonad-contrib>     Perhaps you need to add ‘transformers’ to the build-depends in your .cabal file.
  xmonad-contrib>     Use -v to see a list of the files searched for.
  xmonad-contrib>    |
  xmonad-contrib> 86 | import Control.Monad.Trans.Maybe (MaybeT (..))
  xmonad-contrib>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  xmonad-contrib>
  ```
  ~seeing as `C.M.T.Maybe` is already used in `xmonad-contrib`, and `xmonad-contrib.cabal` already lists `transformers` under `build-depends`.~

- ~Whether to split the first commit (which updates stale freedesktop.org URLs, like xmonad/xmonad#559) into a separate PR.~

- Related to the previous two points: this is my first public contribution in Haskell, let alone XMonad, so please point out any mistakes or changes needed.

Thanks!

### Checklist

        manually, ...) and concluded: manually, with Emacs as an X client

### DIFF
--- a/XMonad/Hooks/ManageHelpers.hs
+++ b/XMonad/Hooks/ManageHelpers.hs
@@ -52,6 +52,7 @@ module XMonad.Hooks.ManageHelpers (
     isMinimized,
     isDialog,
     isNotification,
+    hasNoFocusOnMap,
     pid,
     desktop,
     transientTo,
@@ -82,6 +83,7 @@ import XMonad.Prelude
 import qualified XMonad.StackSet as W
 import XMonad.Util.WindowProperties (getProp32s)
 
+import Control.Monad.Trans.Maybe (MaybeT (..))
 import System.Posix (ProcessID)
 
 -- | Denotes a side of a screen. @S@ stands for South, @NE@ for Northeast
@@ -203,6 +205,32 @@ isDialog = isInProperty "_NET_WM_WINDOW_TYPE" "_NET_WM_WINDOW_TYPE_DIALOG"
 isNotification :: Query Bool
 isNotification =
   isInProperty "_NET_WM_WINDOW_TYPE" "_NET_WM_WINDOW_TYPE_NOTIFICATION"
+
+-- | A predicate to check whether a window should avoid stealing focus
+-- when it is mapped. This is the case when the @_NET_WM_USER_TIME@
+-- property for a particular window is zero. Useful with
+-- 'XMonad.Hooks.Focus.keepFocus', for example:
+--
+-- > import XMonad.Hooks.Focus
+-- > -- Honour requests from Emacs to not switch focus to new windows.
+-- > myFocus :: FocusHook
+-- > myFocus = new (className =? "Emacs" <&&> hasNoFocusOnMap) --> keepFocus
+--
+-- See <https://specifications.freedesktop.org/wm/1.5/ar01s05.html#id-1.6.16>.
+hasNoFocusOnMap :: Query Bool
+hasNoFocusOnMap = do
+    w <- ask
+    -- Per the spec, _NET_WM_USER_TIME_WINDOW should take precedence
+    -- over _NET_WM_USER_TIME. However, some X clients set a zero
+    -- value only on the base window, so check both.
+    liftX . fmap isJust . runMaybeT $ maybeZero w <|> maybeZeroW w
+  where
+    maybeZero w = do
+        [t] <- MaybeT $ getProp32s "_NET_WM_USER_TIME" w
+        guard $ t == 0
+    maybeZeroW w = do
+        [w'] <- MaybeT $ getProp32s "_NET_WM_USER_TIME_WINDOW" w
+        maybeZero $ fi w'
 
 -- | This function returns 'Just' the @_NET_WM_PID@ property for a
 -- particular window if set, 'Nothing' otherwise.
--- a/XMonad/Hooks/ManageHelpers.hs
+++ b/XMonad/Hooks/ManageHelpers.hs
@@ -83,7 +83,6 @@ import XMonad.Prelude
 import qualified XMonad.StackSet as W
 import XMonad.Util.WindowProperties (getProp32s)
 
-import Control.Monad.Trans.Maybe (MaybeT (..))
 import System.Posix (ProcessID)
 
 -- | Denotes a side of a screen. @S@ stands for South, @NE@ for Northeast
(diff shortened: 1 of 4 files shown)


XMonad/Hooks/FadeInactive.hs
```diff
@@ -44,7 +44,7 @@ import qualified XMonad.StackSet as W
 -- > main = xmonad def { logHook = myLogHook }
 --
 -- fadeAmount can be any rational between 0 and 1.
--- you will need to have xcompmgr <http://freedesktop.org/wiki/Software/xapps>
+-- you will need to have xcompmgr <https://freedesktop.org/wiki/Software/xapps/>
```
geekosaur: It might be worth updating that to reference picom; it's showing how long ago I wrote it. 😁

geekosaur (approved):

XMonad/Hooks/FadeInactive.hs
basil-conto: The reason I hesitated to do that is because Picom can be configured to fade inactive windows directly (and I didn't know if Xcompmgr already supported this).

geekosaur: It can, but it's less flexible than xmonad FadeHooks and has no access to xmonad internal state.

geekosaur (approved):

slotThe (commented):
I wonder about advertising support for something without actually supporting it. Is there anything that stops us from just honouring the no-focus request when it comes up? Then essentially the whole change could live in `EwmhDesktops`

basil-conto:
> > > Hooking X.H.Focus into EwmhDesktops seems like something that may have unintended consequences. X.H.Focus uses quite a bit of magic to work around limitations of xmonad's design, and I'm not sure we want to enable that broadly.
> > 
> > 
> > FWIW, I've been using this in my personal config and haven't noticed any issues so far (I frequently want new temporary windows that don't receive focus).
> > Is there a better way for me/others to achieve the no-focus-on-map case?
> 
> I'm not saying it's bad, I'm just saying it's intricate enough to be wary of enabling it for everyone. As an example, if someone already uses X.H.InsertPosition in their config, then that might interfere weirdly with X.H.Focus.

To be clear, the current patch isn't enabling this for everyone, nor does it introduce `X.H.Focus` into `X.H.EwmhDesktops`. It currently only adds a `Query` predicate which users are free to incorporate in their config as needed.

However, in order for the predicate to work reliably, XMonad needs to advertise support for at least `_NET_WM_USER_TIME`. This is the part that is enabled for 'everyone' (namely, those that use `X.H.EwmhDesktops.ewmh`). However, this happens once-off in `startupHook`.

> Also, X.H.InsertPosition mentions that the order of managehooks matters — doShift actions will be undone by InsertPosition, as it's implemented by removing the window (added by X.Operations.manage) from the StackSet and inserting it back into the right position and adjusting focus. If EwmhDesktops itself added something to the manageHook, it'd be difficult to get the ordering right.

The patch does not currently change the handling of `manageHook` in `EwmhDesktops`. What am I missing?

> Speaking of that, InsertPosition is considerably simpler (and faster) than X.H.Focus, and this specific use case of not focusing the new window might be implemented similarly without the heavy machinery of manageFocus. Doesn't negate anything I said about those complex interactions and ordering, but might be "better" than X.H.Focus, if no-focus-on-map is all you need.

Thanks, I'll try that out soon.

liskin:
> > The patch does not currently change the handling of `manageHook` in `EwmhDesktops`. What am I missing?
> 
> Oh, you must have been commenting on the idea to do something like `handleAboveBelowState`. Sorry about the noise/confusion.

Yes, I was commenting on the first of the "two alternative approaches" you outlined earlier. And also on Tony's general idea of "advertise support _and_ also actually support by honoring the request". I know the current PR only does the advertising, but I agree with Tony that it's a bit weird. And as I tried to explain, "just support it" doesn't seem feasible, so my preferred outcome is that the manageHook is accompanied by a simple way to enable that advertising.

But... I'm actually starting to recall why addSupported isn't exported. It's probably because I wanted to redesign the whole EWMH support back in 2021, and the new interface was meant to be more modular - so instead of having a boolean flag `handleAboveBelowState`, and the whole implementation living here in EwmhDesktops, you'd have a separate module that defined the EWMH atoms it supports (and these would be added together with the rest and advertised in _NET_WM_SUPPORTED) and also the hooks for it. And then you'd have a simple EWMH module that would tie it all together. I think there's a draft PR somewhere. Or at least an old branch in my fork.

That never made it over the finish line, though (and quite possibly it even predates ExtensibleConf... it was a dead end kinda?). It's just a bit of context for why things are left in whatever state they are left in. Sorry for all this rambling BTW, it's not necessarily meant for you to action on, it's just context for whoever is curious enough.

I'll take another look when I'm sober. Perhaps addSupported needs a bit of love before it can be added to the public API safely.

XMonad/Hooks/ManageHelpers.hs
```diff
@@ -204,6 +206,32 @@ isNotification :: Query Bool
...
+    -- over _NET_WM_USER_TIME. However, some X clients set a zero
+    -- value only on the base window, so check both.
+    liftX . fmap isJust . runMaybeT $ maybeZero w <|> maybeZeroW w
+  where
+    maybeZero w = do
+        [t] <- MaybeT $ getProp32s "_NET_WM_USER_TIME" w
+        guard $ t == 0
+    maybeZeroW w = do
+        [w'] <- MaybeT $ getProp32s "_NET_WM_USER_TIME_WINDOW" w
+        maybeZero $ fi w'
```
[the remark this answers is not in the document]
slotThe: You can push it if you want :)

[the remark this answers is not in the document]
basil-conto: Now done :).

geekosaur (approved):

PR #986, opened 2026-07-24
`DebugWindow`: refactor and implement proper EWMH atom handling

geekosaur:
### Description

Fixes a number of bugs, simplifies the code, and implements the documented behavior of switching to raw output if any non-EWMH atoms are present.

The refactor will help support KDE extension atoms if I decide to implement them at some point, such that they will behave like EWMH atoms.

### Checklist

        manually, ...) and concluded: testing locally

  - [ ] ~~I updated the `CHANGES.md` file~~ no new features, just bug fixes

### DIFF
--- a/XMonad/Util/DebugWindow.hs
+++ b/XMonad/Util/DebugWindow.hs
@@ -130,6 +130,7 @@ tryCompound t@(TextProperty _ enc _ _) =  do
   when (enc /= cOMPOUND_TEXT) $ error "String is not COMPOUND_TEXT"
   withDisplay $ \d -> io $ wcTextPropertyToTextList d t
 
+-- @@@ can this be reused above?
 splitNul    :: String -> [String]
 splitNul "" =  []
 splitNul s  =  let (s',ss') = break (== '\NUL') s in s' : splitNul ss'
@@ -156,7 +157,6 @@ wrap s =  ' ' : '"' : wrap' s ++ "\""
                   | otherwise  =        s' : wrap' ss
     wrap' ""                   =             ""
 
--- and so is getCommand
 safeGetCommand     :: Display -> Window -> X [String]
 safeGetCommand d w =  do
   wC <- getAtom "WM_COMMAND"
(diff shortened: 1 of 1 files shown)


XMonad/Util/DebugWindow.hs
```diff
@@ -177,45 +177,59 @@ getMachine w =  catchX' (getAtom "WM_CLIENT_MACHINE" >>= getDecodedStringProp w)
...
+                                  sep | i         = " "
+                                      | otherwise = ""
+                              go as (t ++ sep ++ s) True
+
+    -- someday refactor so a list of simplifications can be passed, per KDE comment above
+    simplify       :: String -> String -> String
+    simplify pfx s =  if pfx `isPrefixOf` s then
+                        map toLower (drop (length pfx) s)
+                      else
+                        fail "incorrect prefix"
```
slotThe, line 221: this just returns `""`, right? `fail` for `[Char]` is just `[]`

XMonad/Util/DebugWindow.hs
```diff
@@ -177,45 +177,59 @@ getMachine w =  catchX' (getAtom "WM_CLIENT_MACHINE" >>= getDecodedStringProp w)
...
+
+    rawDecode             :: [Atom] -> String -> String -> X String
+    rawDecode as' pfx sfx =  go as' "{" False
       where
-        go []     t = return t
-        go (a:as) t = simplify "_NET_WM_STATE_" a >>= \t' -> go as (t ++ ' ':t')
+        go            :: [Atom] -> String -> Bool -> X String
+        go []     s False =  return $ pfx ++ s ++ "}" ++ sfx
+        go []     _ True  =  return ""
+        go (a:as) s i     =  do
```
slotThe: Do I read this right that if `as` is non-empty, it just always returns `""`?

XMonad/Util/DebugWindow.hs
geekosaur: Mmm, I switched that around and apparently didn't fix it. The idea was for it to be in `X`, so `fail` would cause an exception caught by `catchX'`, forcing the whole list to be reformatted in "raw" mode. But I later combined two almost identical implementations of "simplify" (one monadic, the other not) and forgot that I needed to keep it in `X` despite a now-removed comment saying so.

XMonad/Util/DebugWindow.hs
geekosaur: The `True` case means we're at the start, so if the list is empty there's nothing to do and the empty string is correct.

XMonad/Util/DebugWindow.hs
geekosaur: And line 206 actually shows the same problem.

XMonad/Util/DebugWindow.hs
slotThe: But the recursive call to `go` below calls it with `True` unconditionally, so don't we always throw away the string we've just accumulated?

geekosaur: Oh, I see, I got `rawDecode` backwards. I'll patch that up and also fix `unAtoms` so the currently three use sites don't need to know they need to pass `False`.

geekosaur: I blame not having gotten enough sleep the past several days (when I started out I seemed to have caught up, but since then it's tanked again).

geekosaur:
One more quick patch-up and I think this is ready: I'd forgotten that I didn't need an extra space before the window type and state, because it's wrapped in parentheses.

